PR #2, opened 2020-06-29
Allow building in a container

ashcrow:
Instead of modifying my Silverblue host to get the deps to build `bootupd` I did a quick container image. Since others may benefit from this as well I decided to do some quick `Makefile` updates and send it up to see if it would be helpful.

## New files
- Dockerfile.build: The container recipe used to make the build image

## New make targets
- `create-build-container`: Uses the container runtime to make a build container
- `build-in-container`: Executes make inside the build container

## New make variables
- `CONTAINER_RUNTIME`: Allows overriding what container runtime will be used

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,9 @@
+FROM registry.fedoraproject.org/fedora:latest
+
+VOLUME /srv/bootupd
+
+WORKDIR /srv/bootupd
+
+RUN dnf update -y && \
+    dnf install -y make cargo rust glib2-devel openssl-devel ostree-devel && \
+    dnf clean all
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 DESTDIR ?=
 PREFIX ?= /usr
 RELEASE ?= 1
+CONTAINER_RUNTIME ?= podman
+IMAGE_PREFIX ?=
+IMAGE_NAME ?= bootupd-build
 
 ifeq ($(RELEASE),1)
         PROFILE ?= release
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -16,6 +17,14 @@ units = $(addprefix systemd/, bootupd.service bootupd.socket)
 all: $(units)
 	cargo build ${CARGO_ARGS}
 
+.PHONY: create-build-container
+create-build-container:
+	${CONTAINER_RUNTIME} build -t bootupd-build -f Dockerfile.build
+
+.PHONY: build-in-container
+build-in-container: create-build-container
+	${CONTAINER_RUNTIME} run -ti --rm -v .:/srv/bootupd:z localhost/bootupd-build make
```
miabbott, line 26: If a user is using `docker` instead of `podman`, I don't think the `localhost/bootupd-build` image name will work...though, I haven't used `docker` in ages, so maybe I am wrong

Dockerfile.build
```diff
@@ -0,0 +1,8 @@
+FROM registry.fedoraproject.org/fedora:latest
+
+VOLUME /srv/bootupd
+
+WORKDIR /srv/bootupd
+
+RUN dnf update -y && \
+    dnf install -y make cargo rust glib2-devel openssl-devel ostree-devel
```
miabbott, line 8: Maybe a `dnf clean all` at the end here

cgwalters:
You don't use a `toolbox` by default?  I live inside an instance of https://github.com/cgwalters/coretoolbox 

Though there is also the "cosa buildroot" which we should document too that's used for CI.

I'm OK with this, though I'm unlikely to use it myself because I work on enough projects that I find it much easier to live inside *one* toolbox container rather than working in a shell on the host and bouncing in and out of different containers.  This approach is definitely used by other projects though, so that's fine by me!

Dockerfile.build
ashcrow: Good call!

Makefile
ashcrow: It won't -- I'll put in an override for that too.

ashcrow:
Will update from feedback. I'll also put a note in the devel instructions noting that toolbox usage is as good if one is already following that pattern :+1:

PR #5, opened 2020-09-03
Fix update, clippy warnings, and start passing test suite

cgwalters:
The test suite now passes again and has somewhat basic reasonable
coverage.

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -140,34 +140,49 @@ fn acquire_write_lock<P: AsRef<Path>>(sysroot: P) -> Result<std::fs::File> {
 }
 
 fn update(_opts: &UpdateOptions) -> Result<String> {
+    let sysroot = openat::Dir::open("/")?;
     let _lock = acquire_write_lock("/")?;
     let mut r = String::new();
-    let state = get_saved_state("/")?.unwrap_or_else(|| SavedState {
+    let mut state = get_saved_state("/")?.unwrap_or_else(|| SavedState {
         ..Default::default()
     });
     for component in get_components() {
         let installed = if let Some(i) = state.installed.get(component.name()) {
             i
         } else {
-            writeln!(r, "Component {} is not installed", component.name()).unwrap();
+            writeln!(r, "Component {} is not installed", component.name())?;
             continue;
         };
-        if let Some(update) = component.query_update()? {
-            component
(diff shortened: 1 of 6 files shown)


src/bootupd.rs
```diff
@@ -152,11 +153,24 @@ fn update(_opts: &UpdateOptions) -> Result<String> {
...
                 .run_update(&installed)
                 .with_context(|| format!("Failed to update {}", component.name()))?;
-            writeln!(r, "Updated {}: {:?}", component.name(), update).unwrap();
+            writeln!(
+                r,
+                "Updated {}: {}",
+                component.name(),
+                format_version(&update)
+            )
+            .unwrap();
```
lucab, line 171: Spare `unwrap` (there are a few pre-existing ones in the surroundings too).

I have some bad feelings about this whole function, as it can fail mid-way through the iterations with some components silently updated, but an `Err` returned.
I'd suggest to return some kind of `Vec<UpdateOutcome>` instead so you can later track which (and how many) components succeeded or failed.

Additionally, do you really want an halt-on-first-error behavior here? I was naively expecting a keep-going iterating approach here (assuming components are independent) and a non-fallible function, but perhaps my understanding is a bit off.

cgwalters: We're calling `write!()` to a `String` - it can't fail.  But, I guess it's just as easy to `?` so I'll do that.

cgwalters: As far as halt-on-first-error yeah...but we do save our state after calling it successfully.

We only have one component right now so it's a moot point.  I added a FIXME comment.

cgwalters: Cleaned up all the `unwrap()` since I agree it is something to avoid in general.

cgwalters: > I'd suggest to return some kind of Vec<UpdateOutcome> instead so you can later track which (and how many) components succeeded or failed.

To elaborate on this, the "source of truth" is the JSON file we save into `/boot`, so a later `bootupd status` will still show the right thing.

lucab: Ack, thanks for the explanation. I have a bit of an instinctive reaction to `unwrap`s as they are usually encountered as quick-prototyping leftovers, while `expect`s are a bit more explicit intentions.

lucab (approved):

PR #9, opened 2020-09-09
Clean up used crates

cgwalters:
First, sort alphabetically.

Then drop some unused crates, some needed looser versions.  For now drop bs58
since it's a minor optimization for UX for the (currently not really
used sha512string).

This gets us to just needing openat-ext packaged for Fedora.


Cargo.toml
```diff
@@ -14,28 +14,22 @@ path = "src/main.rs"
...
-hex = "0.4.2"
-bincode = "1.3.1"
-bs58 = "0.3.1"
-byteorder = "1"
-c_utf8 = "0.1.0"
-lazy_static = "1.1.0"
-serde_plain = "0.3.0"
-chrono = { version = "0.4.11", features = ["serde"] }
+serde = { version = "^1.0", features = ["derive"] }
+serde_derive = "^1.0"
```
lucab, line 29: This is probably redundant given that the `serde` entry above enables the `derive` feature already, I guess?

lucab:
You likely want to run a `cargo update` on this change too.

cgwalters:
> You likely want to run a cargo update on this change too.

We don't have `Cargo.lock` in git right now.  Probably later.

Cargo.toml
cgwalters: Trying to drop that gives me failures, I simplified it the other way.

lucab (approved):

Cargo.toml
lucab: I was curious/unsure about what you were seeing here, so I tried to quickly reproduce it but I didn't manage to. https://github.com/coreos/bootupd/pull/10 is an happy by-product of that.

PR #10, opened 2020-09-10
cargo: directly use serde derive feature

lucab:
Minor followup to https://github.com/coreos/bootupd/pull/9#discussion_r486300204.
This tweaks serde usage in order to directly consume its derive feature.


cgwalters:
Oh duh of course, thanks.

PR #11, opened 2020-09-11
Add functional RPM packaging

cgwalters:
Roughly derived from `rust-afterburn.spec`.


jlebon (approved):
Seems reasonable to me!

PR #15, opened 2020-09-15
Expose binary in /usr/bin/bootupctl

cgwalters:
A good way to look at what we need is something much like
systemd's `bootctl` except for shim/grub2.

It's not as nice for admins to explicitly type
`/usr/libexec/bootupd`, plus it intermixes options
with things they shouldn't use.

Split options into frontend (`/usr/bin/bootupctl`) and backend
(things invoked on the build side).  Install via hardlinks
and dispatch by default on `argv[0]`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ name = "bootupd"
 path = "src/bootupd.rs"
 
 [[bin]]
-name = "bootupd"
+name = "bootupctl"
 path = "src/main.rs"
 
 [dependencies]
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -58,10 +58,11 @@ struct StatusOptions {
     json: bool,
 }
 
+// Options exposed by `bootupctl backend`
 #[derive(Debug, Serialize, Deserialize, StructOpt)]
 #[structopt(name = "boot-update")]
 #[structopt(rename_all = "kebab-case")]
-enum Opt {
+enum BackendOpt {
     /// Install data from available components into a disk image
     Install {
         /// Source root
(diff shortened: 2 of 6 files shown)


cgwalters:
Eh let's just move the binary, it's simpler.

src/bootupd.rs
```diff
@@ -58,10 +58,11 @@ struct StatusOptions {
     json: bool,
 }
 
+// Options exposed by `bootupctl backend`
```
lucab, line 61: Nit: proper docstrings need three slashes, `///`.

src/bootupd.rs
```diff
@@ -75,6 +76,13 @@ enum Opt {
         /// Physical root mountpoint
         sysroot: String,
     },
+}
+
+// "end user" options, i.e. what people should run on client systems
```
lucab, line 81: Same nit as above.

lucab (approved):

Cargo.toml
```diff
@@ -11,7 +11,7 @@ name = "bootupd"
 path = "src/bootupd.rs"
 
 [[bin]]
-name = "bootupd"
+name = "bootupctl"
```
lucab, line 14: If you ever get back to this, I'd much prefer keeping the previous binary name and hardlink `bootupctl` later to it.

lucab:
Minor nit, but feel free to merge. I'd too prefer the hardlinking path, but I do understand complexity/time tradeoffs. Depending on the final outcome, a ticket to track that as a to-do item would be good.

src/bootupd.rs
cgwalters: I think that would result in structopt showing them in `--help` which I don't actually want here.

Cargo.toml
cgwalters: Yeah I did it that way in a previous version of this PR, it worked fine but then I realized...what's the point?  The fact that it's a daemon is ideally mostly an implementation detail.

This PR also changes the CLI to split the "end user interface" off from the "OS builder" side so it's an incompatible change anyways.  I could keep the previous version but eh.

(i.e. we should think of it more like "bootupd-ater" than "bootupd-aemon")

cgwalters:
Merging for now but happy to do followup changes from this!

PR #17, opened 2020-09-18
ci: Fix test script

cgwalters:
The test naming convention changed in cosa, let's grab all
the bootupd tests anyways in case we add more.


cgwalters:
I'm not sure Prow is fully functional here yet...not sure why, need to look at https://github.com/openshift/release/pull/11774 again

PR #18, opened 2020-09-18
Track when an update is interrupted and print it

cgwalters:
We could handle this better, but let's at least start
tracking when an update was interrupted so we can print it.


jlebon (approved):
Not familiar with the codebase, but looks sane to me.

cgwalters:
@jlebon can you retry lgtm?

PR #19, opened 2020-09-21
Client refactor

cgwalters:
Move more update logic to client side
    
    Similar to the previous move of `status`, return a structure
    from the daemon and render it on the client.  This actually
    moves more logic for updating to the client too; the IPC request
    updates just one component, and the client iterates over all
    installed components to update.
    
    This way we also better return "partial" status.

---

    Move formatting of status to client
    
    The way we were just serializing a `String` from daemon → client
    was rather...crude.  Instead have the daemon return a serialized
    status object which we print on the client end.

---

    Rework IPC to use generics
    
    Prep for moving things like printing output to the client.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -46,6 +46,29 @@ pub(crate) struct SavedState {
     pub(crate) pending: Option<BTreeMap<String, ContentMetadata>>,
 }
 
+/// The status of an individual component.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct ComponentStatus {
+    /// Currently installed version
+    pub(crate) installed: ContentMetadata,
+    /// In progress update that was interrupted
+    pub(crate) interrupted: Option<ContentMetadata>,
+    /// Update in the deployed filesystem tree
+    pub(crate) update: Option<ContentMetadata>,
+}
+
+/// Representation of bootupd's worldview at a point in time.
+/// This is intended to be a stable format that is output by `bootupctl status --json`
+/// and parsed by higher level management tools.  Transitively then
+/// everything referenced from here should also be stable.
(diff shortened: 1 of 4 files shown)


src/model.rs
```diff
@@ -46,6 +46,32 @@ pub(crate) struct SavedState {
     pub(crate) pending: Option<BTreeMap<String, ContentMetadata>>,
 }
 
+/// Representation of bootupd's worldview at a point in time.  This is intended
```
lucab, line 49: Nit: it is customary to put a newline between the starting line and the rest of the text. That way `rustdocs` knows how to split the short summary and the longer-form description.

src/model.rs
```diff
@@ -46,6 +46,32 @@ pub(crate) struct SavedState {
...
+
+/// Representation of bootupd's worldview at a point in time.  This is intended
+/// to be a stable format that is output by `bootupctl status --json`
+/// and parsed by higher level management tools.  Transitively then
+/// everything referenced from here should also be stable.
+#[derive(Serialize, Deserialize, Default, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct Status {
+    /// Maps a component name to status
+    pub(crate) components: BTreeMap<String, ComponentStatus>,
```
lucab, line 74: Just checking: did you perhaps intend to care about the *insertion* order (versus current *lexicographic* order)? If so, see https://docs.rs/indexmap.

lucab (approved):

src/model.rs
cgwalters: We want predictable serialization order, so lexicographic is right.

ashcrow (approved):
Test failure seen in other PR and doesn't not related specifically to this change:

```
 /bin/bash: line 2: ./target/release/bootupd: No such file or directory 
```

/lgtm

PR #20, opened 2020-09-21
main: setup logging

lucab:
This introduces and sets up a logging environment for all CLI logic.
It also re-arranges docstrings so that they properly appear in rustdoc
output.


src/main.rs
```diff
@@ -1,26 +1,37 @@
...
+/*!
+Bootupd command-line application.
 
-//! See: https://github.com/coreos/fedora-coreos-tracker/issues/510
-//! This is an early prototype hidden/not-yet-standardized mechanism
-//! which just updates EFI for now (x86_64/aarch64 only).
-//!
-//! But in the future will hopefully gain some independence from
-//! ostree and also support e.g. updating the MBR etc.
+To run the unit tests for this code, use `make check TESTS=tests/check/test-ex-boot-update.sh`
```
cgwalters, line 4: Not new but this has been wrong for a while, mind removing it too?

cgwalters (approved):

src/bootupd.rs
```diff
@@ -1,8 +1,15 @@
-/*
- * Copyright (C) 2020 Red Hat, Inc.
- *
- * SPDX-License-Identifier: Apache-2.0
- */
+/*!
+**Boot**loader **upd**ater.
```
ashcrow, line 2: :smile: :+1:

ashcrow (approved):
Test failure does not look related to this change
```
/bin/bash: line 2: ./target/release/bootupd: No such file or directory
```

Will await response to [this comment](https://github.com/coreos/bootupd/pull/20/files#r492022366) before giving it the go to merge.

cgwalters:
> /bin/bash: line 2: ./target/release/bootupd: No such file or directory

https://github.com/coreos/bootupd/issues/21

COPYRIGHT
```diff
@@ -0,0 +1,7 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
```
cgwalters, line 1: I'm "meh" on this...it doesn't seem to add much information that isn't already in the git repo.  But, not a problem either.

cgwalters (approved):
/lgtm

COPYRIGHT
lucab: I can leave this one out, but it replaces those ubiquitous boilerplate copyright lines (that I dropped here).

PR #22, opened 2020-09-21
tests: Validate that we can't run bootupd as non-root

cgwalters:
Perhaps in the future we'll allow non-root status but
that's a lot more code complexity, no immediate need
for it.


ashcrow (approved):

lucab:
@cgwalters is this something that the client should detect and warn on its own (until expanded to non-root), rather than depending on a EPERM at some point in the chain?

PR #24, opened 2020-09-22
cli: split handlers and runners for sub-commands

lucab:
This moves CLI parsing logic and sub-commands handling to its own
module, making full use of structopt for matching verbs to enum
variant and for dispatching to the relevant entrypoint.

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -24,6 +24,8 @@ use structopt::StructOpt;
 
 // #[cfg(any(target_arch = "x86_64"))]
 // mod bios;
+mod cli;
+pub use cli::CliOptions;
 mod component;
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 mod efi;
@@ -82,7 +84,7 @@ enum Opt {
 }
 
 /// A message sent from client to server
-#[derive(Debug, Serialize, Deserialize, StructOpt)]
+#[derive(Debug, Serialize, Deserialize)]
 enum ClientRequest {
     /// Update a component
     Update { component: String },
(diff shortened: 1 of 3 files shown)


src/bootupd.rs
```diff
@@ -82,7 +84,7 @@ enum Opt {
 }
 
 /// A message sent from client to server
-#[derive(Debug, Serialize, Deserialize, StructOpt)]
```
lucab, line 85: @cgwalters I was surprised by this one, and I still can't figure what is the purpose. Am I missing anything?

cgwalters: I think having `StructOpt` was just a copy-paste mistake - is that the question?

src/cli/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+/// CLI sub-commands.
+#[derive(Debug, StructOpt)]
+pub(crate) enum CliCommand {
+    #[structopt(name = "backend")]
+    Backend(crate::BackendOpt),
+    #[structopt(name = "daemon")]
+    Daemon,
```
cgwalters, line 33: Part of the intent before was that these options were hidden.  That said perhaps what we really want to do is go the hardlink path again with `/usr/libexec/bootupd` and dispatch again based on `argv[0]`.

cgwalters (approved):
/lgtm

src/bootupd.rs
lucab: Yes, thanks. I just wasn't sure if there was something else behind that derive being there.

src/cli/mod.rs
lucab: The multiexec approach looks cleaner. How do you want to bucket the sub-commands?
 1. `bootupd daemon` ?
 1. `bootupctl status|update` ?
 1. where does `backend` belong?

PR #25, opened 2020-09-22
generate-update-metadata: Print the result

cgwalters:
Might as well output information at build time.


lucab (approved):
/lgtm

cgwalters:
Hmm lucab is in `OWNERS` as an approver.  :confused:

PR #27, opened 2020-09-22
update: Require BOOTUPD_ACCEPT_PREVIEW in environment

cgwalters:
Further ensure admins don't accidentally try this until we've
tested it very heavily.


ashcrow (approved):
Looks good -- it may be helpful to add this to the README.md as well.

PR #28, opened 2020-09-22
Misc

cgwalters:
model: Remove some dead code
    
    We aren't implementing transitions right now, updates are manually
    initiated.

---

    Add some docstrings
    
    Just trying to increase doc coverage.

---

    De-indent status & update code
    
    Use the "return early" pattern to avoid excess indentation.


lucab (approved):
/lgtm

PR #29, opened 2020-09-22
Use UTC for serialized timestamp

cgwalters:
Using local timestamps is obviously broken; while our
build process will probably use UTC, relying on that is
wrong, and clients may not use UTC.

Noticed this while going to do some other work around
comparing `ContentMetadata`.

I did verify that RPM's `%{buildtime}` is UTC - the source
invokes `time(NULL)` and the value isn't interpreted
when printed.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -14,7 +14,7 @@ pub(crate) const BOOTUPD_UPDATES_DIR: &str = "usr/lib/bootupd/updates";
 #[derive(Serialize, Deserialize, Clone, Debug, Hash, PartialEq, Eq)]
 pub(crate) struct ContentMetadata {
     /// The timestamp, which is used to determine update availability
-    pub(crate) timestamp: NaiveDateTime,
+    pub(crate) timestamp: DateTime<Utc>,
     /// Human readable version number, like ostree it is not ever parsed, just displayed
     pub(crate) version: Option<String>,
 }
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -9,10 +9,10 @@ use std::io::prelude::*;
 use std::path::Path;
 use std::process::Command;
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use serde::{Deserialize, Serialize};
 
-use chrono::NaiveDateTime;
+use chrono::prelude::*;
 
 use crate::component::*;
 use crate::filetree;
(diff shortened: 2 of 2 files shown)


src/model.rs
```diff
@@ -14,7 +14,7 @@ pub(crate) const BOOTUPD_UPDATES_DIR: &str = "usr/lib/bootupd/updates";
 #[derive(Serialize, Deserialize, Clone, Debug, Hash, PartialEq, Eq)]
 pub(crate) struct ContentMetadata {
     /// The timestamp, which is used to determine update availability
-    pub(crate) timestamp: NaiveDateTime,
+    pub(crate) timestamp: DateTime<Utc>,
```
lucab, line 17: As this holds content which should stay stable across time, it would be better to use a plain-old type here (i64) and just note that it is a Unix-epoch timestamp.

lucab, line 17: As this holds content which should stay stable across time, it would be better to use a plain-old type here (i64) and just note that it is a Unix-epoch timestamp.

cgwalters: Right, that's what ostree and RPM do, as well as many other binary formats.  But JSON is intended to be human readable and serializing to a string (ISO 8601) is a popular choice: https://stackoverflow.com/questions/10286204/what-is-the-right-json-date-format

Related to this of course is JSON/JavaScript's issue with [representing numbers as double](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) which certainly doesn't matter for seconds-since-epoch but I would currently always hesitate to put 64 bit values into JSON on general principle.

So I think this is the right approach.

lucab: I think I was more concerned with stability of `DateTime<Utc>` serialization. Though rustdoc seems to explicitly mention being a RFC 3339 string, so I guess that's a moot point.

cgwalters: Yep https://github.com/chronotope/chrono/blob/8d6d83f7b8a33625ce0263b99b344b12307995da/src/datetime.rs#L848

cgwalters: Now we could definitely choose now to use something other than JSON.... https://github.com/ron-rs/ron is much more human readable/writable even and doesn't have the 64 bit issue but...we ship `jq` as part of the OS for a reason - JSON is a popular lowest common denominiator.  (We could serialize as ron or something else and keep `bootupctl status --json` - they are independent)

lucab (approved):
/lgtm

PR #30, opened 2020-09-22
filetree: Remove timestamp

cgwalters:
UEFI doesn't care about timestamps on binaries, there's no reason
for us to do so either.  Nothing was using this value.


lucab (approved):
/lgtm

PR #31, opened 2020-09-23
Make version number not optional

cgwalters:
There's no good reason to have version numbers be optional.
They're essential for human understanding of things.

We still use a timestamp around for total ordering.

Rework the model to more precisely keep track of "upgradable".

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -154,7 +154,7 @@ impl Component for EFI {
 
         let meta = ContentMetadata {
             timestamp: **largest_timestamp,
-            version: Some(version),
+            version: version,
         };
         write_update_metadata(sysroot_path, self, &meta)?;
         Ok(meta)
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -9,10 +9,10 @@ use std::io::prelude::*;
 use std::path::Path;
 use std::process::Command;
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use serde::{Deserialize, Serialize};
 
-use chrono::NaiveDateTime;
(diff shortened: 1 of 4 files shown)


cgwalters:
OK rebased :surfer: - hadn't realized the degree of interlocked with two other PRs.

src/efi.rs
```diff
@@ -9,10 +9,10 @@ use std::io::prelude::*;
 use std::path::Path;
 use std::process::Command;
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use serde::{Deserialize, Serialize};
 
-use chrono::NaiveDateTime;
+use chrono::prelude::*;
```
ashcrow, line 15: nit: This module is meant to be used as a `*` import. However, it would be nice to denote this and maybe link to [the doc](https://docs.rs/chrono/0.4.15/chrono/prelude/index.html) in a comment.

src/model.rs
```diff
@@ -70,3 +98,24 @@ pub(crate) struct Status {
...
     pub(crate) components: BTreeMap<String, ComponentStatus>,
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use chrono::Duration;
+
+    #[test]
+    fn test_meta_compare() {
```
ashcrow, line 108: :heart: unittests

ashcrow (approved):

src/efi.rs
cgwalters: The `prelude` naming is a pretty strong Rust convention for "use this for `*` imports" I think.

lucab (commented):
/lgtm

PR #32, opened 2020-09-23
cli: rework into a multicall binary

lucab:
This reworks the CLI options parser to branch the inner logic
to accommodate two entrypoints, `bootupd` and `bootupctl`, into
a single multicall binary.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -1,97 +1,39 @@
 //! Command-line interface (CLI) logic.
 
-use crate::ipc::ClientToDaemonConnection;
-use anyhow::{Context, Result};
+use anyhow::Result;
 use structopt::StructOpt;
 
-/// Top-level CLI options.
+mod bootupctl;
+mod bootupd;
+
+/// Top-level multicall CLI.
 #[derive(Debug, StructOpt)]
-pub struct CliOptions {
-    /// CLI sub-commands.
-    #[structopt(subcommand)]
-    pub(crate) cmd: CliCommand,
+pub enum MultiCall {
+    Ctl(bootupctl::CtlCommand),
+    D(bootupd::DCommand),
 }
 
-impl CliOptions {
-    /// Run CLI application.
+impl MultiCall {
+    pub fn from_args() -> Self {
+        use std::os::unix::ffi::OsStrExt;
+
+        // This is a multicall binary, dispatched based on the introspected
+        // filename found in argv[0].
+        let exe_name = {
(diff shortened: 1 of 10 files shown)


lucab:
CI complains about:
```
Executing `postprocess` inline script '0'
+ /usr/bin/bootupctl backend generate-update-metadata /
error: Found argument 'backend' which wasn't expected, or isn't valid in this context

USAGE:
    bootupctl <SUBCOMMAND>
```

Am I right assuming that we'd like this to be `bootupd backend` instead? Perhaps I should leave an hidden one on `bootupctl` for the time being in order to ratchet changes?

cgwalters (changes requested):
> Am I right assuming that we'd like this to be bootupd backend instead? 

I think we want it to be `/usr/libexec/bootupd generate-update-metadata`.  Though this is debatable for sure.  I'd most like for the main binary in `/usr/bin` to be the "end user facing" thing and keep the daemon and backend bits separate.

> Perhaps I should leave an hidden one on bootupctl for the time being in order to ratchet changes?

Yeah I think we should ratchet.  Though I wouldn't have a problem force-pushing an update to https://github.com/coreos/fedora-coreos-config/pull/595 after force-landing a PR.

src/cli/mod.rs
```diff
@@ -4,39 +4,81 @@ use crate::ipc::ClientToDaemonConnection;
...
     }
 }
 
-/// CLI sub-commands.
+/// `bootupd` sub-commands.
 #[derive(Debug, StructOpt)]
-pub(crate) enum CliCommand {
+#[structopt(
+    name = "bootupd",
+    about = "Bootload updater, backend and daemon logic."
```
cgwalters, line 43:
```suggestion
    about = "Backend commands"
``` ?

File: tests/kola/test-bootupd [hunk 2]
```diff
@@ -63,7 +63,7 @@ prepare_efi_update
 # echo somenewfile > ${ostefi}/somenew.efi
 rm -v ${ostefi}/shim.efi
 echo bootupd-test-changes >> ${ostefi}/grubx64.efi
-bootupctl backend generate-update-metadata /
+bootupd backend generate-update-metadata /
```
cgwalters, line 66: This will also need to be `/usr/libexec/bootupd` since `/usr/libexec` is intentionally not in `$PATH`.

cgwalters:
Looks like this needs a rebase.

lucab:
Rebased, CI is green.

cgwalters (approved):
/lgtm

PR #34, opened 2020-09-23
Add a `validate` command

cgwalters:
Let's add support for checking the on-disk state.  This helps
cross-check our implementation and is also useful for administrators
(ref `ostree fsck`, `rpm -V` etc).

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -89,6 +89,8 @@ enum Opt {
 enum ClientRequest {
     /// Update a component
     Update { component: String },
+    /// Validate a component
+    Validate { component: String },
     /// Print the current state
     Status,
 }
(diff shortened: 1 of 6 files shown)


src/bootupd.rs
```diff
@@ -208,6 +210,21 @@ fn update(name: &str) -> Result<ComponentUpdateResult> {
     })
 }
 
+/// daemon implementation of component validate
+fn validate(name: &str) -> Result<ValidationResult> {
+    let _lock = acquire_write_lock("/")?;
```
lucab, line 215: This pattern seems problematic in many regards:
 * to my understanding there is no writing involved here. It seems that you are aiming for an RW Lock, and at this specific site the R half would be enough.
 * the lock is decoupled from what it is protecting. I *think* it is only guarding `get_saved_state`, but I'm not sure if it is meant to protect other part of component walking/validation too.
 * related to the above, it would be good to drop the lock early as soon as it not needed anymore, and let the compiler aware of this via proper typing.
 * there are some callsites where `get_saved_state` is called without holding this lock.

I'd rather see a dedicated R/W lock guard object (`BootStateGuard`?), with methods attached to it to retrieve lock-protected content. In this scenario, `SavedState` may also need to borrow the guard if you care about its content not becoming stale mid-flight.

cgwalters: Yeah, we can probably just drop this lock.

cgwalters: >  I think it is only guarding get_saved_state

We replace that file atomically, so it's not guarding that.  The intention is to be a "double check" against two instances of `bootupd update` somehow running.  In practice, running as a systemd unit *should* enforce this - the old process should be killed by the time a new one starts.  But I like having the redundancy.

lucab (approved):
/lgtm

PR #35, opened 2020-09-23
efi: Some minor cleanup

cgwalters:
- Don't need to serialize anymore
- Use `Default` since it's nicer

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -125,7 +125,7 @@ pub(crate) fn get_components() -> Vec<Box<dyn Component>> {
     let mut components: Vec<Box<dyn Component>> = Vec::new();
 
     #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-    components.push(Box::new(efi::EFI::new()));
+    components.push(Box::new(efi::EFI::default()));
 
     // #[cfg(target_arch = "x86_64")]
     // components.push(Box::new(bios::BIOS::new()));
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -10,7 +10,6 @@ use std::path::Path;
 use std::process::Command;
 
 use anyhow::{bail, Context, Result};
-use serde::{Deserialize, Serialize};
 
 use chrono::prelude::*;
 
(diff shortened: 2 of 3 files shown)


src/efi.rs
```diff
@@ -24,15 +23,9 @@ use crate::util::CommandRunExt;
 /// The path to the ESP mount
 pub(crate) const MOUNT_PATH: &str = "boot/efi";
 
-#[derive(Serialize, Deserialize)]
+#[derive(Default)]
 pub(crate) struct EFI {}
 
-impl EFI {
-    pub(crate) fn new() -> Self {
```
lucab, line 31: What about repurposing this as a `pub(crate) fn new_component() -> Box<dyn Component>`?

src/bootupd.rs
```diff
@@ -124,8 +124,9 @@ pub(crate) fn install(source_root: &str, dest_root: &str) -> Result<()> {
 pub(crate) fn get_components() -> Vec<Box<dyn Component>> {
     let mut components: Vec<Box<dyn Component>> = Vec::new();
 
+    // Unwrap safety - this component is built in for these targets
     #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-    components.push(Box::new(efi::EFI::new()));
+    components.push(component::new_from_name("EFI").expect("efi"));
```
lucab, line 129: I did like the previous state of this. Why did you change the approach? Would the `EFI::new_component` suggestion below help?

cgwalters: Hm...is your objection the new unwrap?  I reworked it to call directly again.

src/efi.rs
cgwalters: If I understand correctly, this doesn't seem much clearer to me than inlining the `Box::new(..:default())` - I reworked to do that.  Does that work?

src/bootupd.rs
lucab: Not directly, I think it was overall on the "lookup via hardcoded string label". It's somehow subjective and minor, thus sorry I couldn't articulate it better. Current form looks good :+1:.

lucab (approved):
/lgtm

PR #36, opened 2020-09-23
status: Also print the boot method

cgwalters:
I forsee people pasting the output of `bootupctl status` into
issues and bugs, so let's gather this one by default.


lucab (commented):
/lgtm

PR #37, opened 2020-09-23
efi: Do check ESP is FAT

cgwalters:
After refactoring this method was lying around as `#[allow(dead_code)]`.
Previously we were trying to use Rust unit tests more but
now we favor the real kola tests.

Let's use it again to double-check we're operating on the
right thing.


lucab (approved):
/lgtm

PR #41, opened 2020-09-28
cli: add flag to control logging level

lucab:
This introduces a new `-v(vv)` flag separately on both `bootupd` and `bootupctl`
to control log verbosity. The default level only shows warnings and
higher priorities.
The service unit sets up the daemon to log info messages too, and
can be overridden via a drop-in.
This also introduces some basic unit-tests on CLI parsing.

/cc @cgwalters

### DIFF
--- a/systemd/bootupd.service
+++ b/systemd/bootupd.service
@@ -4,7 +4,8 @@ Documentation=https://github.com/coreos/bootupd
 
 [Service]
 Type=notify
-ExecStart=/usr/libexec/bootupd daemon
+Environment=BOOTUPD_VERBOSITY="-v"
+ExecStart=/usr/libexec/bootupd daemon $BOOTUPD_VERBOSITY
 # On general principle
 ProtectHome=yes
 # So we can remount /boot writable
--- a/src/cli/bootupctl.rs
+++ b/src/cli/bootupctl.rs
@@ -1,12 +1,37 @@
 use crate::ipc::ClientToDaemonConnection;
 use anyhow::Result;
+use log::LevelFilter;
 use structopt::clap::AppSettings;
 use structopt::StructOpt;
 
 /// `bootupctl` sub-commands.
 #[derive(Debug, StructOpt)]
 #[structopt(name = "bootupctl", about = "Bootupd client application")]
-pub enum CtlCommand {
+pub struct CtlCommand {
+    /// Verbosity level (higher is more verbose).
+    #[structopt(short = "v", parse(from_occurrences), global = true)]
+    verbosity: u8,
+
+    /// CLI sub-command.
+    #[structopt(subcommand)]
+    pub cmd: CtlVerb,
+}
+
+impl CtlCommand {
+    /// Return the log-level set via command-line flags.
+    pub(crate) fn loglevel(&self) -> LevelFilter {
+        match self.verbosity {
+            0 => LevelFilter::Warn,
+            1 => LevelFilter::Info,
+            2 => LevelFilter::Debug,
+            _ => LevelFilter::Trace,
+        }
+    }
+}
+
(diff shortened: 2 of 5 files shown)


systemd/bootupd.service
```diff
@@ -4,7 +4,8 @@ Documentation=https://github.com/coreos/bootupd
 
 [Service]
 Type=notify
-ExecStart=/usr/libexec/bootupd daemon
+Environment=BOOTUPD_VERBOSITY="-v"
+ExecStart=/usr/libexec/bootupd daemon $BOOTUPD_VERBOSITY
```
cgwalters, line 8: Feels like systemd really needs to grow some sort of
`ExecStartArgs=` that could be added with a drop-in or something; this pattern of bridging environment to args seems somewhat fixed and repetitive.

Wouldn't it be better to support configuration via environment directly?  Can't they already do that via `RUST_LOG`?

cgwalters (commented):
Seems sane to me overall, but one question/concern.

systemd/bootupd.service
lucab: > Feels like systemd really needs to grow some sort of ExecStartArgs=

We can probably cover this RFE somewhere before going to the systemd folks, but I'm not so sure and I think that the final interface would be quite similar (a map of string -> string) with the only difference being leaving `envp` pristine.

> Wouldn't it be better to support configuration via environment directly? Can't they already do that via RUST_LOG?

Yes `RUST_LOG` would still work, especially for more advanced things (e.g. selectively enabling debug messages from a specific library).
It isn't a very discoverable/friendly UX though, and for `bootupctl` specifically there are few bleeding edges as the crate name (used to control per-module logging) does not match the binary name.
This is also narrower in scope, as it only bridges the top-level crate level (versus the global logging level) to a CLI flag, allowing to directly get debug/trace messages from the application without all the transitive dependencies too (i.e. `RUST_LOG=bootupd=debug`).
To me it seems worth at least for `bootupctl`, and possibly on `bootupd` to avoid having to each every single user about the module-syntax of `RUST_LOG`.

lucab:
Rebased and followed up on comments.

systemd/bootupd.service
travier: I feels to me that this would be rarely used and maybe best left to a doc page with general debugging instructions.

travier: But this is similar to what you did in [incati](https://github.com/coreos/zincati/blob/master/dist/systemd/system/zincati.service) and there is value in being consistent so I don't know.

cgwalters (approved):
/lgtm

lucab:
Sorry, GH somehow screwed up the notifications for the additional comments. Yes this is how agent logging can be [configured](https://coreos.github.io/zincati/usage/logging/) in Zincati, but we don't have to stick to that.

A few interesting points are:
 * the [default filter](https://github.com/coreos/bootupd/blob/master/src/main.rs#L21) enables warnings (and higher) for **this application crate only**. I find this default useful to avoid spamming user with info messages if everything is going fine.
 * most times I've found it useful asking users to look at higher verbosity logs, for the application only. reqwest/tokio/mio/$DEPS logs are very noisy and unlikely useful for users, and `RUST_LOG=<level>` turns on all of it.
 * bootupd also has a client-side `bootupctl` (which Zincati doesn't have right now), which also needs to have some kind of log-level knob, and the crate name there does not match the binary name.

We can approach this some other way, I'm open to suggestion. My sweet spot is a simple way to control **application** logs, warning always on for both client and daemon, and service configured to emit info messages too (for historical analysis).
For the client additionally, I find that `RUST_LOG=bootupd=info bootupctl <verb>` isn't a great UX and hardly discoverable by casual users.

PR #42, opened 2020-09-28
daemon: log and keep going on client errors

lucab:
This tweaks daemon core-logic to keep going on client-induced errors,
logging them at "error" level.
It also introduces docstrings and minor reflows to ease casual code navigation.

/cc @cgwalters

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -15,7 +15,6 @@ Refs:
 
 use anyhow::{bail, Context, Result};
 use fs2::FileExt;
-use nix::sys::socket as nixsocket;
 use openat_ext::OpenatDirExt;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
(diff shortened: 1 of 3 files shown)


src/bootupd.rs
```diff
@@ -27,6 +26,7 @@ use std::path::Path;
 mod cli;
 pub use cli::MultiCall;
 mod component;
+mod daemon;
```
cgwalters, line 29: This is great, I had started on more clearly splitting the daemon/client but shelved it.

cgwalters (approved):

src/bootupd.rs
lucab: Thanks for the feedback, I wasn't sure about the code layout so I did it here opportunistically in a separate commit (for easy dropping). I'm taking a mental note, if I happen to touch more of client logic I'll go for a similar setup.

lucab: Mostly same setup, a bit of spacemacs too. But to avoid brain meltdown I try not to have too many tabs in a row, and prefer split windows. Although there are other breadcrumb/hints, I can see the `mod.rs` overcrowding you mention, indeed that PR would help.

PR #44, opened 2020-09-28
WIP: e2e test that upgrades grub

cgwalters:
This isn't working yet, something going wrong in 9p


tests/e2e/e2e-in-vm.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
lucab, line 1: I think this would need to source `libtest.sh` in VM, no?

cgwalters:
OK I got a pass on this locally.  Requires https://github.com/coreos/bootupd/pull/48

cgwalters:
Hmm we seem to have lost CoreOS CI...

cgwalters:
/hold
until CoreosCI is back up and runs this

tests/e2e/e2e.sh
```diff
@@ -0,0 +1,99 @@
...
+        RequiresMountsFor=/run/testtmp
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        Environment=TARGET_COMMIT=${target_commit}
+        Environment=TARGET_GRUB_NAME=${target_grub_name}
+        Environment=TARGET_GRUB_PKG=${target_grub_pkg}
+        Environment=SRCDIR=/run/bootupd-source
+        # Run via shell because selinux denies systemd writing to 9p apparently
+        ExecStart=/bin/sh -c '/run/bootupd-source/tests/e2e/e2e-in-vm.sh &>>/run/testtmp/out.txt; test -f /run/rebooting || poweroff -ff'
```
lucab, line 78: Just noting here, possibly for future consideration, from an out-of-band chat: in order to avoid spiraling into a scripting trap and VM-to-kola dependencies, we could explore having an hidden `bootupd exttests` runner and move the testing logic into the codebase.

cgwalters: Another variant of this is going to a workspace and having a separate test binary - this is like https://github.com/ostreedev/ostree/tree/master/tests/inst (the main ostree is written in C so it's not a workspace)

cgwalters:
It looks like the previous coreosci run didn't execute the new CI logic  for some reason.  Retrying.

cgwalters:
Hooray, got a successful run of both the new and old CI.  I'm debating a bit whether to keep the kola test...it's much cheaper to run but also much less useful.

tests/e2e/testrpmbuild.sh
```diff
@@ -0,0 +1,142 @@
+# Copied from rpm-ostree
```
cgwalters, line 1: Note this isn't used *yet* but I plan to do so.

lucab:
LGTM. I have a feeling this may end up a flaking a lot (the koji/grub thing likely), but let's see how it goes.

PR #47, opened 2020-09-30
daemon: avoid direct printing, log instead

lucab:
This converts all daemon logic to stop directly printing to stdout/stderr,
and instead perform logging at the appropriate level.

/cc @cgwalters


cgwalters (approved):
/lgtm

PR #48, opened 2020-09-30
filetree: Fix creation of parent directories

cgwalters:
An embarrassing bug where I accidentally intermixed
absolute path handling with fd-relative, and wasn't caught
by either the unit tests or integration tests because the
current directory was writable.

But it was caught by a new test suite which doesn't use
`rpm-ostree usroverlay`.

### DIFF
--- a/src/filetree.rs
+++ b/src/filetree.rs
@@ -303,7 +303,7 @@ pub(crate) fn apply_diff(
         let path = Path::new(pathstr);
         if let Some(parent) = path.parent() {
             // TODO: care about directory modes?  We don't for FAT.
-            std::fs::create_dir_all(parent)?;
+            destdir.ensure_dir_all(parent, 0o755)?;
         }
         let destp = tmpname_for_path(path);
         copy_file_at(srcdir, destdir, path, destp.as_path())
@@ -339,6 +339,7 @@ pub(crate) fn apply_diff(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::fs;
     use std::io::Write;
 
     fn run_diff(a: &openat::Dir, b: &openat::Dir) -> Result<FileTreeDiff> {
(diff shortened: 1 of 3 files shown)


cgwalters:
Uses https://github.com/cgwalters/openat-ext/pull/4

src/filetree.rs
```diff
@@ -303,7 +303,7 @@ pub(crate) fn apply_diff(
         let path = Path::new(pathstr);
         if let Some(parent) = path.parent() {
             // TODO: care about directory modes?  We don't for FAT.
-            std::fs::create_dir_all(parent)?;
+            destdir.ensure_dir_all(parent, 0o755)?;
```
lucab, line 306: Nit: I think you also took care of the `TODO` above with this. If so, it can be dropped.
Also, there is a very confusing `#[allow(dead_code)] // Used for testing` at the top of this function, which judging from this patch I think it isn't really true.

lucab:
One small nit/confusion from my side, but otherwise big thanks for also covering this with a regression test!

src/filetree.rs
cgwalters: Mmm...I think the `TODO` isn't quite fixed, actually it looks like:
```
rpm -qlv efi-filesystem
drwx------    2 root     root                        0 Jan 28  2020 /boot/efi
drwxr-xr-x    2 root     root                        0 Jan 28  2020 /boot/efi/EFI
drwx------    2 root     root                        0 Jan 28  2020 /boot/efi/EFI/BOOT
drwx------    2 root     root                        0 Jan 28  2020 /boot/efi/EFI/fedora
```

So perhaps we should be 0700 in all of these directories.  (But eh, if we do that why not just make the toplevel `/boot/efi` 0700?)


As far as `dead_code` - yeah definitely but let's fix that as a followup?

lucab: Fine for doing both later, to be honest I thought the new `0o755` was exactly a reaction to the TODO.

lucab:
@cgwalters I think I've seen coreos-ci coming back on other repositories, but you may need to force-push this one to re-trigger it.

cgwalters:
CoreOS CI passing here!

lucab (approved):
/lgtm

PR #49, opened 2020-09-30
filetree: Remove lots of #[allow(dead_code)]

cgwalters:
For a while the filetree code wasn't used by EFI, now it is
again, so remove most of the `allow(dead_code)` and only use
`cfg(test)` for a few of them that are only used by tests.

Came up in PR review.


lucab (approved):
/lgtm

Issue #43, opened 2020-09-28
implement auto-exit on idle


cgwalters:
I am pretty sure at one point we did the "auto-exit on idle" pattern but it doesn't look like we are today.  We should just to conserve resources.

Let's depend on tokio and add an idle timeout in the daemon.

lucab:
I think the ~daemon~ devil is in the details here, it looks like "depend on tokio" is a much larger task which first requires making sure that nothing in the daemon logic blocks the executor and things can be gracefully canceled, no? In particular, the `accept` [step](https://github.com/coreos/bootupd/blob/bfd47c93dbff810722ef333151ed17a4fb267cdd/src/daemon/mod.rs#L64) indefinitely blocks right now and can't be retired (and every IPC calls from there down too).

Perhaps a shorter-term goal with an higher immediate benefit is to process a single client and then exit. That would induce the same start-latency upfront, and make sure that we don't mistakenly start cobbling things that assume an always running service (e.g pull-based metrics).

cgwalters:
> Perhaps a shorter-term goal with an higher immediate benefit is to process a single client and then exit.

Yep, that seems simpler indeed.

cgwalters:
Well I tried
```
diff --git a/src/cli/bootupd.rs b/src/cli/bootupd.rs
index 377bf65..4a3834d 100644
--- a/src/cli/bootupd.rs
+++ b/src/cli/bootupd.rs
@@ -57,7 +57,7 @@ impl DCommand {
     /// Run CLI application.
     pub fn run(self) -> Result<()> {
         match self.cmd {
-            DVerb::Daemon => crate::daemon::run_coreloop(),
+            DVerb::Daemon => crate::daemon::run(),
             DVerb::Install(opts) => Self::run_install(opts),
             DVerb::GenerateUpdateMetadata(opts) => Self::run_generate_meta(opts),
         }
diff --git a/src/daemon/mod.rs b/src/daemon/mod.rs
index f1ed85d..f5209b4 100644
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -5,26 +5,29 @@ use anyhow::{bail, Context, Result};
 use nix::sys::socket as nixsocket;
 use std::os::unix::io::RawFd;
 
-/// Run daemon core-logic loop, endlessly.
-pub fn run_coreloop() -> Result<()> {
+/// Accept a single client and then exit; we don't want to
+/// persistently run as a daemon.  The systemd unit is mostly
[...]

PR #51, opened 2020-10-01
daemon: Exit when done processing requests

cgwalters:
There's no reason to sit there taking up RAM just because
someone did `bootupctl status`.  Properly implementing
exit-on-idle really requires an event loop (e.g. `tokio`)
but that'd be a large change.

For now let's just exit when a client disconnects.  However,
this quickly broke the test suite which does a lot of
commands in a row.

Tweak the systemd unit to allow more restarts, and also
add a few strategic sleeps.

Closes: https://github.com/coreos/bootupd/issues/43


lucab (approved):
/lgtm

PR #52, opened 2020-10-01
filetree: Remove TODO and link to fcos umask PR

cgwalters:
Nothing to fix here.

### DIFF
--- a/src/filetree.rs
+++ b/src/filetree.rs
@@ -16,6 +16,12 @@ use std::path::Path;
 
 /// The prefix we apply to our temporary files.
 pub(crate) const TMP_PREFIX: &str = ".btmp.";
+// This module doesn't handle modes right now, because
+// we're only targeting FAT filesystems for UEFI.
+// In FAT there are no unix permission bits, usually
+// they're set by mount options.
+// See also https://github.com/coreos/fedora-coreos-config/commit/8863c2b34095a2ae5eae6fbbd121768a5f592091
+const DEFAULT_FILE_MODE: u32 = 0o700;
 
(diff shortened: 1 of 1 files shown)


src/filetree.rs
```diff
@@ -294,7 +294,8 @@ pub(crate) fn apply_diff(
     for pathstr in diff.additions.iter().chain(diff.changes.iter()) {
         let path = Path::new(pathstr);
         if let Some(parent) = path.parent() {
-            // TODO: care about directory modes?  We don't for FAT.
+            // Note that for FAT there are no unix permission bits.
+            // See also https://github.com/coreos/fedora-coreos-config/commit/8863c2b34095a2ae5eae6fbbd121768a5f592091
```
lucab, line 298: Minor: I was following the referenced PR/discussion and the world is settling on `0700`. I noted your `Perhaps we should be 0700 in all of these directories. (Though if the toplevel /boot/efi is 0700 then most of the other perms don't matter)` and I somehow agree that in order to avoid surprises/inconsistencies it may be good to align child directories on `0700`.
Did you forget to update the line below here, did you change your mind, or did I mix up something?

cgwalters: Well it's more that it doens't matter really what we pick here.  But I updated to use `0700` here too.

lucab:
/hold
/lgtm

coreos-ci is taking a short nap, this will need to be re-triggered afterwards.

travier (approved):

cgwalters:
/hold cancel
CI is back up

lucab (approved):
/lgtm

PR #55, opened 2020-10-02
state: Consistently kebab-case and validate a previous saved format

cgwalters:
I noticed we were missing `kebab-case` on a few structs (but
it doesn't matter since none of them have a name that would change).

To ensure we don't regress on parsing old file formats, add a JSON
file from a current system and test loading it.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -12,6 +12,7 @@ use std::collections::BTreeMap;
 pub(crate) const BOOTUPD_UPDATES_DIR: &str = "usr/lib/bootupd/updates";
 
 #[derive(Serialize, Deserialize, Clone, Debug, Hash, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
 pub(crate) struct ContentMetadata {
     /// The timestamp, which is used to determine update availability
     pub(crate) timestamp: DateTime<Utc>,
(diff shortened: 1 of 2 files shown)


src/model.rs
```diff
@@ -118,4 +121,14 @@ mod test {
         assert!(a.can_upgrade_to(&b));
         assert!(!b.can_upgrade_to(&a));
     }
+
+    /// Validate we're not breaking the serialized format.
+    #[test]
+    fn test_deserialize() -> Result<()> {
+        let data = include_str!("example-state-v0.json");
```
lucab, line 128: I'd suggest that we introduc a `tests/fixtures/` directory and move this one there. It would hopefully encourage a trend of adding more pristine fixtures to cover the interop surface.

cgwalters: fixtured.

lucab:
/hold

coreos-ci is still down thus missing a run on this one, but I'm confident this is fine to merge even without that. Deferring to Colin for when he prefers to let this land.

lucab (approved):
/lgtm

lucab (approved):
/lgtm

cgwalters:
/hold cancel
CI is up

lucab (approved):
/lgtm

Issue #38, opened 2020-09-24
implement adoption


cgwalters:
Today we handle the case where bootupd was used to create the disk image in the first place.

We also want to handle "adoption" - e.g. with an existing CoreOS system, ideally all the way back to an RHCOS 4.1 bootimage, we should be able to successfully update it to the latest EFI (and in the future, BIOS).

Example flow:
```
$ bootupctl status
Component EFI:
  Not installed via bootupd
  Detected known state: CoreOS disk image aleph א‎ version: 32.20200922.dev.3
$ bootupctl adopt
Component EFI:
  Version: coreos-aleph-32.20200922.dev.3
  Upgrade: Available: <version in ostree>
$ bootupctl update
Updated to <version in ostree>
```

See https://github.com/coreos/fedora-coreos-tracker/blob/master/internals/README-internals.md#aleph-version for more information on the aleph.

Handling systems *without* aleph is uglier, but I think in practice there aren't too many bootimages for OpenSHift 4.1 and 4.2 that will be in widespread use - we could hardcode them, or just punt and say "unknown" and offer `bootupctl adopt --force` and track it as an "unknown" version.

cgwalters:
PR in https://github.com/coreos/bootupd/pull/66

PR #56, opened 2020-10-02
Parse and print CoreOS aleph version

cgwalters:
This is prep for work on "adoption":
https://github.com/coreos/bootupd/issues/38

In order to (as safely as possible) take over management
of a system *not installed via bootupd* we want to gather
as much information as we can.

For example, we can use the imgid as the "version" of the
current bootloader data.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -88,6 +88,12 @@ pub(crate) struct ComponentStatus {
     pub(crate) updatable: ComponentUpdatable,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) enum OperatingSystem {
+    CoreOS { aleph_imgid: String },
+}
+
 /// Representation of bootupd's worldview at a point in time.
 /// This is intended to be a stable format that is output by `bootupctl status --json`
 /// and parsed by higher level management tools.  Transitively then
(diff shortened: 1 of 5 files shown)


lucab:
/hold

coreos-ci is still missing on this one.

src/model.rs
```diff
@@ -97,6 +97,8 @@ pub(crate) struct ComponentStatus {
 pub(crate) struct Status {
     /// Maps a component name to status
     pub(crate) components: BTreeMap<String, ComponentStatus>,
+    /// If a CoreOS system is detected, this will be the imgid from the "aleph version"
+    pub(crate) coreos_aleph_imgid: Option<String>,
```
lucab, line 101: Assuming that in the future you want to implement "unknown" as described at the bottom of https://github.com/coreos/bootupd/issues/38, should this be something like `install_label: Option<Adopted>` where Adopted is an tagged-kind enum? Something like:
```
enum Adopted {
  CoreOSAleph(String),
  ManuallyForced(String),
  Unknown,
}
```

cgwalters: I tweaked things to take a step in this direction.

cgwalters:
/hold cancel
CI is back

lucab (approved):
/lgtm

src/model.rs
lucab: Ack, SGTM. Although I got slightly confused with the new patch. My understanding was that with this field we were trying to answer the question "what was the original version of the bootloader at install time" while now it seems to have morphed into "current OS auto-detection".
I guess I'll wait for the rest of the pile of patches to get a better grasp of the whole picture.

PR #58, opened 2020-10-05
cargo: drop library crate, for now

lucab:
This removes the current library/binary crate split, leaving
only the application in place.
The original idea was to add the logic as a vendored library into
rpm-ostree for an emergency fix. However the urgency dropped and
this now lives as a self-standing application.
We may or may not consider re-adding it later on, but only for the
interop parts (paths, JSON, etc).

/cc @cgwalters


cgwalters (approved):
/lgtm

PR #60, opened 2020-10-06
main: Add more error context

cgwalters:
I was debugging a failure to write the statefile which
turned out to be me having
https://github.com/coreos/fedora-coreos-config/pull/659
locally.

Anyways let's add some error context to aid future debugging.


lucab (approved):
/lgtm

PR #61, opened 2020-10-06
Use a lot more systemd hardening features

cgwalters:
Just on general principle.  In particular the combination
of a readonly `/usr` and `WorkingDirectory=/usr/share/empty`
will give us a consistent clean error if we somehow accidentally
write to the current working directory.

### DIFF
--- a/systemd/bootupd.service
+++ b/systemd/bootupd.service
@@ -11,7 +11,17 @@ StartLimitBurst=10
 Type=notify
 Environment=BOOTUPD_VERBOSITY="-v"
 ExecStart=/usr/libexec/bootupd daemon $BOOTUPD_VERBOSITY
-# On general principle
+# This way our working directory isn't writable by default.
+WorkingDirectory=/usr
+# Various hardening flags just on general principle.  We need
+# to run as root, but let's avoid accidental damage.
 ProtectHome=yes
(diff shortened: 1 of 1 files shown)


systemd/bootupd.service
```diff
@@ -11,7 +11,17 @@ StartLimitBurst=10
 Type=notify
 Environment=BOOTUPD_VERBOSITY="-v"
 ExecStart=/usr/libexec/bootupd daemon $BOOTUPD_VERBOSITY
-# On general principle
+# This way our working directory isn't writable by default.
+WorkingDirectory=/usr/share/empty
```
lucab, line 15: Apparently this is a RPM-specific directory introduced in 2000 (but I couldn't find more info on that).

Neither LHS nor a Debian machine have it. I don't know how broad is the "distribution-independent" bit in the README, but this likely won't work on non-RPM systems.

cgwalters: I'd say non-Fedora systems (it's not a function of the package format so much as the `filesystem` package).  But yes.  I guess we could make our own.   It'd be really unfortunate to add a `#[cfg]` for this.

Eh, I just changed it to use `/usr` since we care more about read-only then we can about empty.

lucab (approved):
/lgtm

PR #62, opened 2020-10-06
Rename tests/e2e → tests/e2e-update

cgwalters:
I plan to add another test `e2e-adopt` for the case of going
from a known disk image not using bootupd initially.


kelvinfan001 (approved):
LGTM :)

PR #64, opened 2020-10-06
status: Fix indentation of CoreOS aleph image ID

cgwalters:
If it's indented it looks like it's part of the EFI component.


kelvinfan001 (approved):
LGTM

PR #66, opened 2020-10-06
 WIP: Implementation of adoption

cgwalters:
This shares a lot of similarity with `update` but I'm
trying to keep them distinct because we may need
to do something different in the future, and I think
we need to be careful when operating on state we didn't create.

Closes: #38


cgwalters:
Working on splitting out some prep work from this PR.

cgwalters:
Aaand the first actually did eat the `grub.cfg` which is exactly the kind of thing we need to handle in adoption vs update.  Fixed now, tested explicitly and now we also reboot into the updated system too.

cgwalters:
OK now this one just depends on https://github.com/coreos/bootupd/pull/72

cgwalters:
OK, lifting WIP on this!

lucab:
It looks reasonable overall. I've left a few minor comments and a suggestion to maybe tweak the trait slightly. I think this may be missing a `cargo fmt` pass too.

cgwalters:
Comments addressed!

cgwalters:
> I think this may be missing a cargo fmt pass too.

`cargo fmt` was clean here - which section of the code made you think it was missing that out of curiosity?

lucab (approved):
/lgtm

PR #67, opened 2020-10-07
ci: Remove unnecessary outer stage{}

cgwalters:
Jenkins doesn't seem to render nested stages, so drop the
outer one.  This makes CI look nicer and more useful
as we add more testing runs.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -32,35 +32,32 @@ codestyle: {
 
 // Build FCOS and do a kola basic run
 // FIXME update to main branch once https://github.com/coreos/fedora-coreos-config/pull/595 merges
-stage("FCOS") {
-  cosaPod(buildroot: true, runAsUser: 0, memory: "3072Mi", cpu: "4") {
-    stage("Build FCOS") {
-      checkout scm
-      unstash 'build'
-      // Note that like {rpm-,}ostree we want to install to both / and overrides/rootfs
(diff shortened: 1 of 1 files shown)


cgwalters:
See e.g. https://jenkins-coreos-ci.apps.ocp.ci.centos.org/blue/organizations/jenkins/github-ci%2Fcoreos%2Fbootupd/detail/PR-66/3/pipeline

lucab:
LGTM but @jlebon came up with this arrangement across all repos, I don't know if there is an explicit rationale for the outer stage stanza.

cgwalters:
Which other repo looks like this?  The [afterburn one](https://github.com/coreos/afterburn/blob/master/.cci.jenkinsfile) for example is tiny and just calls out to the standard workflow.

Other ones [like ostree](https://github.com/ostreedev/ostree/blob/master/.cci.jenkinsfile) use `parallel` inside which Jenkins explicitly handles.

I think I just messed this one up initially.

jlebon:
> LGTM but @jlebon came up with this arrangement across all repos, I don't know if there is an explicit rationale for the outer stage stanza.

Hmm, I don't think I usually use a top-level outer stage like this. In the end, all projects are 100% encouraged to re-organize the stages in ways that serve them better.

lucab:
My bad, I took the rpm-ostree as a single sample and inferred that we had the same stage-based setup in all other our repos too.

lucab (approved):
/lgtm

PR #68, opened 2020-10-07
daemon: Log all requests at debug level

cgwalters:
The requests aren't large, let's log the whole thing consistently
rather than having a copy/paste version for each request.  This
way we include the component too.


lucab (approved):
/lgtm

PR #69, opened 2020-10-07
coreos: Add timestamp to aleph version

cgwalters:
Prep for adoption; we use the aleph as a `ContentMetadata`
which wants both a timestamp and a version string.


lucab (approved):
/lgtm

PR #70, opened 2020-10-07
main: Gather status even if no components are installed

cgwalters:
Prep for adoption work.

Previously we just bailed out if nothing was installed,
but we want to return a status even if there are no
components installed.

Rework the component list to be a map so we can conveniently
look up things by name.


lucab (approved):
/lgtm

PR #71, opened 2020-10-07
model: Also validate serialized status

cgwalters:
We do want tooling to parse `bootupctl status --json`, so let's
validate we're not unexpectedly breaking it.

### DIFF
--- a/tests/fixtures/example-status-v0.json
+++ b/tests/fixtures/example-status-v0.json
@@ -1,0 +1,22 @@
+{
+  "components": {
+    "EFI": {
+      "installed": {
+        "timestamp": "2020-09-15T13:01:21Z",
+        "version": "grub2-efi-x64-1:2.04-23.fc32.x86_64,shim-x64-15-8.x86_64"
+      },
+      "interrupted": null,
+      "update": {
+        "timestamp": "2020-09-15T13:01:21Z",
+        "version": "grub2-efi-x64-1:2.04-23.fc32.x86_64,shim-x64-15-8.x86_64"
+      },
+      "updatable": "at-latest-version",
+      "adopted-from": null
+    }
+  },
+  "os": {
+    "core-o-s": {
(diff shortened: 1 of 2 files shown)


tests/fixtures/example-status-v0.json
```diff
@@ -0,0 +1,23 @@
...
+      "adopted-from": null
+    }
+  },
+  "adoptable": {},
+  "os": {
+    "core-o-s": {
+      "aleph_imgid": "fedora-coreos-32.20201007.dev.0-qemu.x86_64.qcow2"
+    }
+  }
+}
```
kelvinfan001, line 23: Should there be a newline character at the end of this file?

tests/fixtures/example-status-v0.json
```diff
@@ -0,0 +1,23 @@
...
+        "timestamp": "2020-09-15T13:01:21Z",
+        "version": "grub2-efi-x64-1:2.04-23.fc32.x86_64,shim-x64-15-8.x86_64"
+      },
+      "updatable": "at-latest-version",
+      "adopted-from": null
+    }
+  },
+  "adoptable": {},
+  "os": {
+    "core-o-s": {
```
lucab, line 19: Freeform labels as keys are usually a painful idea from the point of view of trying to have producers and consumers agreeing on the same schema.

With the larger context that you have in your mind about where you want to drive this, do you think it would be feasible to reflow this into something like:
```json
"os" : {
  "kind": "coreos",
  "content": "<imgid>"
}
```
?

cgwalters: Yes, though the idea of this PR is just to add validation for what we have *now*, then changes like https://github.com/coreos/bootupd/pull/72 are covered by CI and highlighted.  Let's adapt that one to also rework per your comment after this merges?

lucab (approved):
/lgtm

PR #72, opened 2020-10-07
 Drop operating system from status, print client side

cgwalters:
For now let's address concerns over how we represent
the operating system by dropping them from the status
JSON.  Instead detect CoreOS specifically client side
too and print the aleph there.

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -1,9 +1,7 @@
 use crate::component::{Component, ValidationResult};
 use crate::coreos;
 use crate::efi;
-use crate::model::{
-    ComponentStatus, ComponentUpdatable, ContentMetadata, OperatingSystem, SavedState, Status,
-};
+use crate::model::{ComponentStatus, ComponentUpdatable, ContentMetadata, SavedState, Status};
 use crate::{component, ipc};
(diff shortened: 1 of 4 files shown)


cgwalters:
Draft since this depends on https://github.com/coreos/bootupd/pull/71

cgwalters:
OK I reworked this, see commit msg/code.

cgwalters:
Rebased :surfer:

lucab:
I guess the longer term plan is to re-instate this logic, right? Otherwise a containerized client will end up going off-track.

cgwalters:
> I guess the longer term plan is to reinstate this logic, right? Otherwise a containerized client will end up going off-track.

Mmm...this is just informational.  The current client and server logic uses this internally but we're just showing it because it is relevant too.

A bit like how eventually it might make sense to have `rpm-ostree status -v` at least also chain to `bootupctl status` perhaps.  The daemon wouldn't rely on that in any way, it's just extra information.

cgwalters:
Any other comments on this?

lucab:
@cgwalters I think we can merge (although I'm a bit confused about the final goal, but I guess this won't hurt for now).

lucab (approved):
/lgtm

PR #73, opened 2020-10-08
model: Deny unknown saved state fields

cgwalters:
This is a big hammer and eventually what we'll really need
to do is introduce versioning or compatibility flags.

That said, for now it's better to ensure we're not reading
anything unexpected.

And in a scenario where bootupd is downgraded, it's not
a really big deal because it just means you can't update
the bootloader - which you probably don't want to do
if you're downgrading the operating system anyways.


lucab (approved):
/lgtm

PR #74, opened 2020-10-13
Fix clippy lints

cgwalters:
Nothing really important, may try to add clippy to CI at some point.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -25,6 +25,7 @@ impl MultiCall {
             let exe_path = std::path::PathBuf::from(arg0);
             exe_path.file_name().unwrap_or_default().to_os_string()
         };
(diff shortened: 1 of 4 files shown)


src/cli/mod.rs
```diff
@@ -27,7 +27,7 @@ impl MultiCall {
         };
         match exe_name.as_bytes() {
             b"bootupctl" => MultiCall::Ctl(bootupctl::CtlCommand::from_iter(args)),
-            b"bootupd" | _ => MultiCall::D(bootupd::DCommand::from_iter(args)),
+            _ => MultiCall::D(bootupd::DCommand::from_iter(args)),
```
lucab, line 30: Hmm, I think this is `wildcard_in_or_patterns`.
Would having the three cases explicitly on different arms (`bootupctl` - `bootupd` - `_`) make `clippy` happier?

PR #76, opened 2020-10-13
Release 0.1.3

cgwalters:


cgwalters:
I actually already pushed this to crates, not realizing that branch protection denied direct pushes to master (IMO a good thing).

cgwalters:
Force merging since this is already released.

lucab:
This should really be a 0.2.0 as we dropped the `lib` part which was previously public.

lucab:
Also, the lockfile for this release somehow got out of sync and is stuck at `0.1.1`: https://github.com/coreos/bootupd/blob/v0.1.3/Cargo.lock#L83

cgwalters:
OK https://github.com/coreos/bootupd/pull/79

PR #77, opened 2020-10-13
Drop requirement for BOOTUPD_ACCEPT_PREVIEW=1

cgwalters:
I'm increasingly confident in our testing, so let's
drop the requirement to set the environment variable.


lucab (approved):
/lgtm

PR #84, opened 2020-10-15
backend/statefile: add a write-lock guard for statefile updates

lucab:
This introduces a write-lock guard in order to to ensure that all
update operations attempted on the on-disk statefile have properly
acquired the system-wide lock first.

/cc @cgwalters

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -3,20 +3,11 @@ use crate::coreos;
 use crate::efi;
 use crate::model::{ComponentStatus, ComponentUpdatable, ContentMetadata, SavedState, Status};
 use crate::{component, ipc};
-use anyhow::{anyhow, bail, Context, Result};
-use fs2::FileExt;
-use openat_ext::OpenatDirExt;
+use anyhow::{anyhow, Context, Result};
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
-use std::io::prelude::*;
 use std::path::Path;
-
-/// Stored in /boot to describe our state; think of it like
-/// a tiny rpm/dpkg database.  It's stored in /boot
-pub(crate) const STATEFILE_DIR: &str = "boot";
-pub(crate) const STATEFILE_NAME: &str = "bootupd-state.json";
-pub(crate) const WRITE_LOCK_PATH: &str = "run/bootupd-lock";
 
 /// A message sent from client to server
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


cgwalters (approved):
Nice!  I agree the Rust pattern of having the mutex own the data is much better.

/lgtm

cgwalters:
Though this failed e2e in a way that makes me think there's a real bug:
```
# No components installed.
# Adoptable: EFI: fedora-coreos-32.20201015.dev.1-qemu.x86_64.qcow2
# CoreOS aleph image ID: fedora-coreos-32.20201015.dev.1-qemu.x86_64.qcow2
# Boot method: EFI
# -rw-r--r--. 1 root root 181 Oct 15 13:08 out.txt
# # No components installed.
# # Adoptable: EFI: fedora-coreos-32.20201015.dev.1-qemu.x86_64.qcow2
# # CoreOS aleph image ID: fedora-coreos-32.20201015.dev.1-qemu.x86_64.qcow2
# # Boot method: EFI
# error: File 'out.txt' doesn't match fixed string list 'Component EFI'
+ fatal 'test failed'
+ echo error: test failed
error: test failed
```

/lgtm cancel

lucab:
/hold

Oh dang, I think I had a silly copy-paste mistake. Let's see if the fixup on top agrees.

lucab:
```
+ /usr/bin/bootupctl backend install --src-root=/tmp/rootfs/ostree/deploy/fedora-coreos/deploy/2c874588f58a48872392d756088677d7205791cd4cf5a1aff73d42bb347cff0a.0 /tmp/rootfs

error: boot data installation failed: failed to acquire write lock: No such file or directory (os error 2)
```

This is because `install` was previously bypassing the write-lock, and it now fails to acquire. I think this because `/run` does not exist in the target sysroot, however it should really be using its own sysroot instead. I'll need to re-arrange the locking logic to take care of this split.

lucab:
CI turned green after the latest fixup, I've thus squashed all changes, this should be ready for the final pass.

PS: is Prow able to autosquash fixups when merging?

cgwalters:
> PS: is Prow able to autosquash fixups when merging?

Not currently AFAIK

src/bootupd.rs
```diff
@@ -32,29 +23,27 @@ pub(crate) enum ClientRequest {
...
+    let mut state = SavedState::default();
     for component in components.values() {
         let meta = component.install(source_root, dest_root)?;
         state.installed.insert(component.name().into(), meta);
     }
 
-    let sysroot = openat::Dir::open(dest_root)?;
-    update_state(&sysroot, &state).context("Failed to update state")?;
+    let mut state_guard =
+        SavedState::acquire_write_lock(source_root).context("failed to acquire write lock")?;
```
cgwalters, line 42: I'm not sure how much sense it makes though to grab any lock here.  The source root isn't being changed.

TBH I think the real bug here is that rpm-ostree is probably not mounting `/run` when it's running postprocessing scripts.

But for now...well, I guess if this works it doesn't hurt but I'd probably have said let's skip locking.

cgwalters (approved):
/lgtm

cgwalters:
https://github.com/coreos/rpm-ostree/pull/2275

PR #86, opened 2020-10-23
main: Fix sysroot lock target for install

cgwalters:
Depends https://github.com/coreos/fedora-coreos-config/pull/707
Followup fix for https://github.com/coreos/bootupd/pull/84#issuecomment-709414840


cgwalters:
OK this is right, but looking at this more I think it reveals a design bug - the target sysroot should be part of the lock guard.

cgwalters:
Closing in favor of https://github.com/coreos/bootupd/pull/87

PR #87, opened 2020-10-23
Rework sysroot/statefile locking

cgwalters:
Have the guard actually own the sysroot - matching how in
Rust the `Mutex<T>` type wraps what it guards.

Also add an `unlocked()` method to deal with the install case,
since there's no `/run` in the target root.

xref https://github.com/coreos/bootupd/pull/86#issuecomment-715351465
xref https://github.com/coreos/bootupd/pull/84#issuecomment-709414840

### DIFF
--- a/src/backend/statefile.rs
+++ b/src/backend/statefile.rs
@@ -21,16 +21,23 @@ impl SavedState {
     /// While ordinarily the daemon runs as a systemd unit (which implicitly
     /// ensures a single instance) this is a double check against other
     /// execution paths.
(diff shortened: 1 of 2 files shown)


lucab:
Oh, this is nice. Indeed I agree my previous  "install should use its own sysroot" locking approach was bogus, good catch.

In the offline case, should the underlying directory+lockfile be removed when unlocking?

cgwalters:
> In the offline case, should the underlying directory+lockfile be removed when unlocking?

Hmm.  Offline = unlocked yet we're creating a lockfile.  I'll fix that.

lucab (approved):
/lgtm

PR #89, opened 2020-10-26
component: Add error prefixing

cgwalters:
I think this is the error we're seeing in
https://github.com/coreos/fedora-coreos-config/pull/711#issuecomment-716780537
But let's add some error prefixing to be sure for the future.


jlebon:
Also opened https://github.com/coreos/bootupd/pull/90.

/lgtm

PR #90, opened 2020-10-26
Add some more error-prefixing

jlebon:
I think we're hitting issues in CI in this path right now.
See: https://github.com/coreos/fedora-coreos-config/pull/711.

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -101,11 +101,13 @@ impl Component for EFI {
             anyhow::bail!("No update metadata for component {} found", self.name());
         };
         let srcdir = component_updatedir(src_root, self);
-        let srcd = openat::Dir::open(&srcdir)?;
+        let srcd = openat::Dir::open(&srcdir)
+            .with_context(|| format!("opening src dir {}", srcdir.display()))?;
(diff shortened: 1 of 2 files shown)


cgwalters:
Hah https://github.com/coreos/bootupd/pull/89

src/efi.rs
```diff
@@ -95,17 +95,25 @@ impl Component for EFI {
...
 
     fn install(&self, src_root: &str, dest_root: &str) -> Result<InstalledContent> {
-        let meta = if let Some(meta) = get_component_update(src_root, self)? {
+        let meta = if let Some(meta) = get_component_update(src_root, self).with_context(|| {
+            format!(
+                "getting update metadata from {} for {}",
+                src_root,
+                self.name()
+            )
+        })? {
```
cgwalters, line 104: I think I prefer the patch from https://github.com/coreos/bootupd/pull/89 which would cover this path (there's not much that can fail other than parsing), but also OK to keep it.

jlebon: Sure, dropped!

cgwalters (approved):
/lgtm

PR #92, opened 2020-10-27
Use new openat-ext APIs for writing files

cgwalters:
Less code, not as error prone (e.g. forgetting to `flush()` a `BufWriter`)
and technically better using `O_TMPFILE`.

(And for symmetry also use openat for reading in the same file case)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -175,6 +175,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "drop_bomb"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


lucab:
It failed in CI with:
```
+ /usr/bin/bootupctl backend generate-update-metadata /
error: generating metadata failed: Invalid cross-device link (os error 18)
```

which I fear is linked to the last comment at https://github.com/cgwalters/openat-ext/pull/8#issuecomment-717238366.

cgwalters:
OK I reworked this one to target just the original case which definitely doesn't involve mount crossing.

kelvinfan001 (approved):
/lgtm

PR #93, opened 2020-10-27
component: More porting to openat

cgwalters:
Clean up this code to use the openat helpers too, just on
general principle of using fd-relative in more places.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,9 +417,9 @@ dependencies = [
 
 [[package]]
 name = "openat-ext"
-version = "0.1.7"
+version = "0.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ea8f6f2601742142683130af7d1b387f10e1c3d73121f069203d4139bbb27c4c"
+checksum = "73e85ee792caf10bf54dc2a844b9f5cbc4a05095360fdeba67ffe687912d2b8b"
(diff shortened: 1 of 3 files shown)


src/component.rs
```diff
@@ -84,16 +75,24 @@ pub(crate) fn component_updatedir(sysroot: &str, component: &dyn Component) -> P
         .join(component.name())
 }
 
+/// Returns the name of the JSON file containing a component's available update metadata installed
+/// into the booted operating system root.
+fn component_update_data_name(component: &dyn Component) -> PathBuf {
+    Path::new(&format!("{}.json", component.name())).into()
```
lucab, line 81: For future reference, direct `PathBuf::from<String>()` [exists](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#impl-From%3CString%3E).

Cargo.lock
```diff
@@ -417,9 +417,9 @@ dependencies = [
 
 [[package]]
 name = "openat-ext"
-version = "0.1.7"
+version = "0.1.8"
```
lucab, line 420: For context, I usually split commits wrt. manifest changes vs lockfile updates vs code changes.
While that makes a single commit not anymore atomic / bisectable, from past experiences it helps cherry-picks and backports.

lucab (approved):
/lgtm

Issue #91, opened 2020-10-27
Fix updating bootupd 0.1 systems


cgwalters:
There were some format cleanups between 0.1.2 and 0.2.0¹ - current bootupd will barf trying to update those.  I think we need a backcompat reader, or perhaps the simplest is to just delete earlier data and treat it the same as an adoption case.

To rephrase the current state is confusing because we handle updating *non-bootupd* systems and bootupd 0.2.0 systems but not bootupd 0.1.

¹ Also 0.1.3 which is ~equivalent but it's less confusing if we just talk about 0.2.x

cgwalters:
This issue gains some urgency since https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/32.20201004.3.0/x86_64/meta.json just hit stable which has 0.1.2 :cry:

PR #94, opened 2020-10-28
Parse old "0.1" format installed state format

cgwalters:
Since bootupd 0.1.1 shipped in FCOS stable
we need to support updating it, as well as
"pre-bootupd" systems and bootupd 0.2 systems.

I tested this manually; not yet adding an e2e yet for
this because our test suite is already really
expensive and the unit test covers most of it.

Closes: https://github.com/coreos/bootupd/issues/91


lucab (approved):
/lgtm

PR #97, opened 2020-11-02
Add a release-checklist + script

cgwalters:
Taken from the one currently in coreos-installer:
https://github.com/coreos/coreos-installer/blob/master/.github/ISSUE_TEMPLATE/release-checklist.md

But tweaked in various ways:

- I want to continue to use `git-evtag`, will try to teach `cargo-release` about that at some point.
- Rework the vendoring stuff to be a script and run beforehand

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-checklist.md
+++ b/.github/ISSUE_TEMPLATE/release-checklist.md
@@ -1,0 +1,65 @@
+# Release process
+
+The release process follows the usual PR-and-review flow, allowing an external reviewer to have a final check before publishing.
+
+In order to ease downstream packaging of Rust binaries, an archive of vendored dependencies is also provided (only relevant for offline builds).
+
+## Requirements
+
+This guide requires:
+
+ * a web browser (and network connectivity)
+ * `git`
+ * GPG setup and personal key for signing
+ * [git-evtag](https://github.com/cgwalters/git-evtag/)
+ * `cargo` (suggested: latest stable toolchain from [rustup][rustup])
+ * A verified account on crates.io
+ * Write access to this GitHub project
+ * Upload access to this project on GitHub, crates.io
+
+## Release checklist
+
+- Prepare local branch+commit
+  - [ ] `git checkout -b release`
+  - [ ] Bump the version number in `Cargo.toml`.  Usually you just want to bump the patch.
+  - [ ] Run `cargo build` to ensure `Cargo.lock` would be updated
+  - [ ] Commit changes `git commit -a -m 'Release x.y.z'`; include some useful brief changelog.
+
+- Prepare the release
+  - [ ] Run `./ci/prepare-release.sh`
+
+- Validate that `origin` points to the canonical upstream repository and not your fork:
+  `git remote show origin` should not be `github.com/$yourusername/$project` but should
+  be under the organization ownership.  The remote `yourname` should be for your fork.
+
+- Push a PR to create the release:
+  - [ ] `git push --set-upstream origin release`
+  - [ ] open a web browser and create a PR for the branch above
+  - [ ] make sure the resulting PR contains the commit
+  - [ ] in the PR body, write a short changelog with relevant changes since last release
+
+- [ ] get the PR reviewed, approved and merged
+
+- publish the artifacts (tag and crate):
+  - [ ] `git fetch origin && git checkout ${RELEASE_COMMIT}`
+  - [ ] verify `Cargo.toml` has the expected version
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+ * `cargo` (suggested: latest stable toolchain from [rustup][rustup])
+ * A verified account on crates.io
+ * Write access to this GitHub project
+ * Upload access to this project on GitHub, crates.io
+
+## Release checklist
+
+- Prepare local branch+commit
+  - [ ] `git checkout -b release`
+  - [ ] Bump the version number in `Cargo.toml`.  Usually you just want to bump the patch.
```
lucab, line 24: This is missing a version update in the lockfile update. (And a bump to the dev version, which I'd personally recommend). cargo-release takes care of both.

.github/ISSUE_TEMPLATE/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Prepare the release
+  - [ ] Run `./ci/prepare-release.sh`
+
+- Validate that `origin` points to the canonical upstream repository and not your fork:
+  `git remote show origin` should not be `github.com/$yourusername/$project` but should
+  be under the organization ownership.  The remote `yourname` should be for your fork.
+
+- Push a PR to create the release:
+  - [ ] `git push --set-upstream yourname`
```
lucab, line 35: In other release checklists we use `origin` for this. The rationale is that anyway the release captain needs access to it in order to push the tag, so it's okay to have the release branch there too, and it doubles as an early check for ACL.

(This remote mixing is also part of cause of the recent misunderstanding that forced us to re-tag coreos-installer and afterburn)

.github/ISSUE_TEMPLATE/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+  - [ ] open a web browser and create a PR for the branch above
+  - [ ] make sure the resulting PR contains the commit
+  - [ ] in the PR body, write a short changelog with relevant changes since last release
+
+- [ ] get the PR reviewed, approved and merged
+
+- publish the artifacts (tag and crate):
+  - [ ] `git fetch origin && git checkout ${RELEASE_COMMIT}`
+  - [ ] verify `Cargo.toml` has the expected version
```
lucab, line 44: This is ambiguous and error-prone (we arleady went though mistakes here in previous release-checklist revisions). 
There exist multiple commits where the manifest contains the expected version (see comment above wrt dev versions).

ci/prepare-release.sh
```diff
@@ -0,0 +1,20 @@
...
+version=$(cargo read-manifest | jq -r .version)
+commit=$(git rev-parse HEAD)
+
+# Generate a vendor tarball of sources to attach to a release
+# in order to support offline builds.
+vendor_dest=target/${name}-${version}-vendor.tar.gz
+rm vendor ${vendor_dest} -rf
+cargo vendor
+# Trim off Windows pre-built libraries that make the vendor/ dir much larger
+find vendor/ -name '*.a' -delete
```
lucab, line 15: Did you try building the resulting vendor bundle?

While I share your feeling, I fear this may invalidate lockfile checksumming.

.github/ISSUE_TEMPLATE/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+  - [ ] `git-evtag sign v${RELEASE_VER}`
+  - [ ] `git push --tags origin v${RELEASE_VER}`
+  - [ ] `cargo publish`
+
+- publish this release on GitHub:
+  - [ ] find the new tag in the [GitHub tag list](https://github.com/coreos/coreos-installer/tags) and click the triple dots menu, and create a release for it
```
lucab, line 50: s/coreos-installer/bootupd/

.github/ISSUE_TEMPLATE/release-checklist.md
cgwalters: Hmm, OK.

ci/prepare-release.sh
cgwalters: Yep I did.

lucab: Nice!

lucab:
This deviates a bit from our usual flow (specifically, it is not using a dev version outside of releases), but I think we should merge this improvement for now and discuss somewhere else why/if dev-versions are better.

/lgtm

Issue #103, opened 2020-11-11
Don't require adoption for FCOS


cgwalters:
I think we should be confident enough in our process that `bootupctl update` should detect the FCOS case and update it transparently.

PR #104, opened 2020-11-11
Update all {F,RH}CoreOS systems that have an aleph version

cgwalters:
Came up in review in https://github.com/coreos/fedora-coreos-docs/pull/203
Basically it's confusing to users that they need to understand
the difference between `update` and `adopt-and-update`.

From bootupd's perspective these are quite different things;
the first case is a completely known quantity, the second isn't.

However...we can be pretty confident that we can update systems
that have a CoreOS aleph version (i.e. they were generated by coreos-assembler),
since we haven't changed how the bootloader is installed there
really much at all.

This means that for now RHCOS 4.{1,2} bootimages that were
generated via Anaconda will still require `adopt-and-update`,
but detecting and validating that can come as a second phase.

The high level logic here is that the status gains a new
`confident: bool` flag (which corresponds right now to
"have CoreOS aleph").  The client side looks at this
and automatically bridges the `update` CLI to
`adopt-and-update`.

Closes: https://github.com/coreos/bootupd/issues/103

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -95,6 +95,16 @@ pub(crate) struct ComponentStatus {
     pub(crate) adopted_from: Option<ContentMetadata>,
 }
 
+/// Information on a component that can be adopted
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct Adoptable {
+    /// A synthetic version
+    pub(crate) version: ContentMetadata,
+    /// True if we are likely to be able to reliably update this system
+    pub(crate) confident: bool,
+}
+
 /// Representation of bootupd's worldview at a point in time.
 /// This is intended to be a stable format that is output by `bootupctl status --json`
 /// and parsed by higher level management tools.  Transitively then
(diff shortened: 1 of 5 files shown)


cgwalters:
OK, this one has green CI.

src/model.rs
```diff
@@ -106,7 +116,7 @@ pub(crate) struct Status {
     /// Maps a component name to status
     pub(crate) components: BTreeMap<String, ComponentStatus>,
     /// Components that appear to be installed, not via bootupd
-    pub(crate) adoptable: BTreeMap<String, ContentMetadata>,
+    pub(crate) adoptable: BTreeMap<String, Adoptable>,
```
lucab, line 119: I think this change breaks the "stable format" premise stated in the docstring, but current CI does not catch this because the entry/map in `example-status-v0.json` fixture is empty.

Is that a problem, or is this breaking change still ok? To my understanding this data-structure is not persisted to disk, and the change only affects client compatibility.

lucab (commented):
Patch looks good, but I do have a doubt regarding a (possibly unintended?) breaking change.

lucab:
/hold
/lgtm

Leaving this pre-stamped in case the breaking change was intended and confirmed ok.

src/model.rs
cgwalters: Yeah, this is a valid concern and you're right this sneaks by our tests today.

In the end, while we have `bootupctl status --json` I think the project is still in a "stabilizing" state - there aren't any known consumers of the JSON yet.  

Hmm, probably we should add a simple "is up to date" API, since that's all that most consumers would care about.

cgwalters:
Will address that in followup
/hold cancel

PR #105, opened 2020-11-13
Add bootupctl status --print-if-available

cgwalters:
This is a middle ground between:

1) Parse arbitrary text output from `bootupctl status`; this could
   change at any time.
2) Parse JSON from `bootupctl status --json`; parsing this requires
   knowlege of bootupd and the JSON format still isn't 100% fixed
   in stone either.

Projects like https://github.com/openshift/machine-config-operator/
will basically want a flow that just allows them to check for
outstanding updates and apply them one node at a time.

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -214,6 +214,24 @@ pub(crate) fn status() -> Result<Status> {
     Ok(ret)
 }
 
+pub(crate) fn print_status_avail(status: &Status) -> Result<()> {
+    let mut avail = Vec::new();
+    for (name, component) in status.components.iter() {
+        if let ComponentUpdatable::Upgradable = component.updatable {
+            avail.push(name.as_str());
+        }
+    }
+    for (name, adoptable) in status.adoptable.iter() {
+        if adoptable.confident {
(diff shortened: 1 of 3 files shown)


lucab:
> Projects like https://github.com/openshift/machine-config-operator/ will basically want a flow that just allows them to check for outstanding updates and apply them one node at a time.

Sorry if this a naïf question, but does this mean we maybe want those projects to only consume the exit code of a `bootupctl status needs-update`, without output capturing in the happy path?

cgwalters:
> Sorry if this a naïf question, but does this mean we maybe want those projects to only consume the exit code of a bootupctl status needs-update

The thing I don't like about that is it masks errors.  There's the approach of testing for a specific exit code (aka `rpm-ostree --exit-unchanged-77`) but feels less elegant than this approach.

lucab:
/lgtm

@cgwalters yes, I was exactly thinking of `rpm-ostree`. Something like those codes (0/1/77) looks nice to me if we only need to convey a tri-state yes/no/err-maybe.
But I'm equally fine if you prefer this, I guess the consumers then will have to consume and check stdout.

Issue #108, opened 2020-11-17
Add support for `bootupctl update --from-pending`


cgwalters:
Splitting this thread from https://github.com/coreos/bootupd/issues/8#issuecomment-708375600

For rpm-ostree systems we should support a flow like:
```
$ rpm-ostree upgrade  # queue a pending deployment
$ bootupctl update --from-pending
```

(or possibly `bootupctl update` does that by default if it detects a pending deployment; not sure)

This flow would be used by orchestrator systems like zincati/MCO.

cgwalters:
Or maybe it should just be the default if we detect a pending deployment.

cgwalters:
Ah, typing "... fix <num>" also means "Closes: <num>".

PR #109, opened 2020-11-17
Pass sysroot through updates

cgwalters:
In order to tackle https://github.com/coreos/bootupd/issues/108
we need to switch to not assuming `/` as the update source.
We do provide the system root as a string in some cases but
not all.  Let's go fd-relative.


lucab (approved):
/lgtm

PR #110, opened 2020-11-17
Thread sysroot through adopt+update flow

cgwalters:
More prep for https://github.com/coreos/bootupd/issues/108
Similar to the previous PR for updates.


lucab (approved):
/lgtm

PR #111, opened 2020-11-18
Thread sysroot through update query flow

cgwalters:
More incremental work on https://github.com/coreos/bootupd/issues/108
In some cases we're using both an absolute path and a dfd;
fixing that gets involved because we need to use it for child
processes too, which will require some more prep work.


lucab (approved):
/lgtm

PR #116, opened 2020-11-30
WIP: Handle ro /boot

cgwalters:


src/bootupd.rs
```diff
@@ -93,6 +93,27 @@ pub(crate) enum ComponentUpdateResult {
...
+fn ensure_writable_mount<P: AsRef<Path>>(p: P) -> Result<()> {
+    use nix::sys::statvfs;
+    let p = p.as_ref();
+    let stat = statvfs::statvfs(p)?;
+    if !stat.flags().contains(statvfs::FsFlags::ST_RDONLY) {
+        return Ok(());
+    }
+    let status = std::process::Command::new("mount")
+        .args(&["-o", "remount,rw"])
+        .arg(p.as_os_str())
```
kelvinfan001, line 105: I think we can directly pass `p` here without `as_os_str()` since `Path` implements the `AsRef<OsStr>` trait right?

kelvinfan001 (commented):
I think we would also need to `ensure_writable_efi()` since we want to [mount `/boot/efi` as read-only](https://github.com/coreos/fedora-coreos-config/pull/659/files) as well right? This should probably be done in `efi.rs`? Also, we'd need to tweak the kola tests a little bit so e.g. https://github.com/coreos/bootupd/blob/0be1b276a99a3496135020562149109d7a710b31/tests/kola/test-bootupd#L118 would work. Mind if I just create another PR to do the above?

cgwalters:
Moving to https://github.com/coreos/bootupd/pull/118

PR #118, opened 2020-12-01
Handle ro /boot

kelvinfan001:
Minor changes from https://github.com/coreos/bootupd/pull/116
Tested on [FCOS with /boot and /boot/efi mounted ro](https://github.com/coreos/fedora-coreos-config/pull/659).

### DIFF
--- a/tests/kola/test-bootupd
+++ b/tests/kola/test-bootupd
@@ -19,7 +19,9 @@ function cleanup () {
   fi
 }
 
-bootefidir=/boot/efi/EFI
+bootmount=/boot
+bootefimount=${bootmount}/efi
+bootefidir=${bootefimount}/EFI
 bootupdir=/usr/lib/bootupd/updates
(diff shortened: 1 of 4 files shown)


tests/kola/test-bootupd
```diff
@@ -115,6 +117,9 @@ bootupctl update | tee out.txt
 assert_file_has_content_literal out.txt 'No update available for any component'
 assert_not_file_has_content_literal out.txt 'Updated EFI'
 
+# Ensure /boot and /boot/efi can be written to
+mount -o remount,rw ${bootmount}
+mount -o remount,rw ${bootefimount}
```
cgwalters, line 122: This is OK for now, but in the future we can clean this up by having the test do something like re-exec itself in a new mount namespace so we're not affecting the main one.

cgwalters (approved):
Thanks!
/lgtm

PR #127, opened 2020-12-21
Ensure ESP is mounted

kelvinfan001:
Following coreos/fedora-coreos-config#794,
the ESP will no longer be mounted by default in FCOS. Mount the
ESP at a directory with a random name in bootupd's sandboxed /tmp
ourselves before reading from/writing to it.

Note that ESPs on redundant disks is not yet supported, this commit
simply mounts the device with part-label `EFI-SYSTEM` if one exists.

`ProtectClock=yes` in `bootupd.service` is preventing the ESP from
being mounted, remove that for now.


kelvinfan001:
Kola tests pass with a local FCOS build with https://github.com/coreos/fedora-coreos-config/pull/794.

bgilbert:
> No longer `ProtectClock=yes` in `bootupd.service` to allow bootupd to mount the ESP.

Hmm, could you expand on this?  Why do these conflict?

src/efi.rs
```diff
@@ -33,16 +35,47 @@ impl EFI {
...
     }
     fn open_esp(&self) -> Result<openat::Dir> {
+        self.ensure_mounted_esp()?;
         let sysroot = openat::Dir::open("/")?;
         let esp = sysroot.sub_dir(&self.esp_path())?;
         Ok(esp)
     }
+
+    fn ensure_mounted_esp(&self) -> Result<()> {
+        let mount_point = &Path::new("/").join(MOUNT_PATH);
```
bgilbert, line 56: Probably we should create a temporary directory to use as a mountpoint, and then clean it up afterward.

src/efi.rs
```diff
@@ -33,16 +35,47 @@ impl EFI {
         Path::new(MOUNT_PATH).join("EFI")
     }
 
+    fn esp_device(&self) -> PathBuf {
+        Path::new("/dev/disk/by-partlabel/").join(ESP_PART_LABEL)
```
bgilbert, line 39: This still assumes there'll be a canonical ESP.  :slightly_smiling_face: 

RAIDed boot disks won't even have a partition with this label.  Instead we should use libblkid or similar to walk any filesystem with the partition GUID `c12a7328-f81f-11d2-ba4b-00a0c93ec93b`.

bgilbert (commented):
Unfortunately there are deeper architectural consequences to having multiple ESPs.  Every operation will need to iterate over all ESPs, and cannot assume that the ESPs are all in the same state.  (There's a reasonable argument that if the ESPs get desynced, fixing it isn't our responsibility, but it'd be better UX if we handled it anyway.)

kelvinfan001:
> Hmm, could you expand on this? Why do these conflict?

Admittedly, it was just trial and error. The tests passed when I removed this.

src/efi.rs
kelvinfan001: I thought there is still a lot of work to be done related to supporting redundant disks for `/boot` and `/boot/efi`. I was thinking this PR would at least let bootupd support the case where we're not using redundant disks, but the EFI is also not mounted (please correct me if I'm wrong!).

bgilbert: Redundant disks for `/boot` should be transparent to bootupd.  But for `/boot/efi`, you're right, and yes, this PR should be sufficient for the non-redundant case.  Let's update the commit message to reflect that goal, and also make sure we have a good error message when we encounter an unsupported environment.

kelvinfan001: ah ok, that makes sense! Thanks for the explanation :)

src/efi.rs
cgwalters: We should probably add `PrivateTmp=yes` too.

cgwalters: Yeah if we use `PrivateTmp` then we can rely on systemd to clean up our tempdir.
Then I'd use https://docs.rs/tempfile/3.1.0/tempfile/struct.TempDir.html#method.into_path (i.e. make a tempdir, mount on it and then return the path).

systemd/bootupd.service
```diff
@@ -19,7 +19,6 @@ ProtectHome=yes
 ReadOnlyPaths=/usr
 PrivateTmp=yes
 PrivateNetwork=yes
-ProtectClock=yes
```
cgwalters, line 22: Probably something going wrong in the seccomp policy systemd makes.  It's...about time...we looked at that.

cgwalters (commented):
Thanks for working on this!

src/efi.rs
cmurf: What do you think about only updating ESP's that are listed in NVRAM? Non-UEFI is trickier, risky to step on all BIOS Boot or MBR gaps with grub-install.

bgilbert: For CoreOS, at least, we can make some simplifying assumptions.  We own the whole system, so we can update every ESP and BIOS Boot partition we see.  CoreOS boots via Default Boot Behavior and doesn't set Boot NVRAM variables at all.

If bootupd is going to generalize beyond CoreOS, of course, those assumptions may no longer be workable.

cmurf: Yep, I'm thinking the idea to rebase Silverblue on CoreOS. That use case includes dual boot, so I think "fair game" includes MBR gap, BIOS Boot, and ESP:`/EFI/fedora/` on the drive that contains the current active `/` but not anything else. i.e. EFI/fedora is owned but not the entire ESP. There's low, but non-zero, chance of USB sticks, SD Cards, external drives or even other internal drives, are present and used to boot something else; and also can't be touched.

src/efi.rs
```diff
@@ -22,27 +23,64 @@ use crate::ostreeutil;
...
     fn open_esp(&self) -> Result<openat::Dir> {
+        self.ensure_mounted_esp()?;
         let sysroot = openat::Dir::open("/")?;
         let esp = sysroot.sub_dir(&self.esp_path())?;
         Ok(esp)
     }
+
+    fn ensure_mounted_esp(&self) -> Result<()> {
+        let mount_point = &Path::new("/").join(&*MOUNT_PATH);
+        let output = std::process::Command::new("mountpoint")
```
cgwalters, line 65: This isn't bad, but now that we know *our* process is the only one doing the mount (in our mount namespace), I think it'd be better to have a something like `Arc<Mutex<Option<PathBuf>>>` or...actually there's [a lazy_init crate for that](https://docs.rs/lazy-init/0.4.0/lazy_init/) even.

Basically do the tempdir creation + mount as a single locked unit instead.

tests/kola/data/libtest.sh
```diff
@@ -81,3 +81,10 @@ assert_not_file_has_content_literal () {
     done
 }
 
+# Mount the EFI partition at a temporary location.
+tmpefimount=/tmp/efi
```
cgwalters, line 85: I know this is just the test suite, but I try as a general rule to avoid predictably named temporary files.  We could have this code mirror the main logic and use e.g. `$(mktemp -d)` and "return" it by printing, so
callers do e.g.:
`tmpefimount=$(mount_tmp_efi)`
etc.

Not a blocker, just noting.

src/efi.rs
```diff
@@ -14,6 +14,7 @@ use anyhow::{bail, Context, Result};
 use openat_ext::OpenatDirExt;
 
 use chrono::prelude::*;
+use structopt::lazy_static::lazy_static;
```
cgwalters, line 17: We should add the dependency to `Cargo.toml` directly; it looks like a bug that structopt is re-exporting it.

cgwalters (approved):
One minor must fix for the crate ref, everything else is optional or can be followup.

Thanks for working on this!

src/efi.rs
cgwalters: Right, so far bootupd is Fedora CoreOS specific but the idea from the start is it could generalize indeed to other rpm-ostree systems (i.e. not necessarily coreos-assembler generated) as well as even not rpm-ostree systems (traditional yum/dnf/zypper/apt).

It's still not clear to me that the latter effort is really worth it but I don't want to exclude it.

src/efi.rs
```diff
@@ -22,27 +23,64 @@ use crate::ostreeutil;
...
-pub(crate) const MOUNT_PATH: &str = "boot/efi";
+/// The ESP partition label
+pub(crate) const ESP_PART_LABEL: &str = "EFI-SYSTEM";
+
+#[macro_use]
+lazy_static! {
+    /// The path to a temporary ESP mount
+    static ref MOUNT_PATH: PathBuf = {
+        // Create new directory in /tmp with randomly generated name at runtime for ESP mount path.
+        tempfile::tempdir_in("/tmp").expect("Failed to create temp dir for EFI mount").into_path()
```
travier, line 34: Any particular reason we're not using /run that is root:root owned instead?

kelvinfan001: Not really any particular reason. Just read up on `/tmp` vs `/run`, I'm guessing `/run` would be more appropriate?

cmurf: Or /run/boot_efi/$VOLUME-ID to account for the multiple ESP case?

kelvinfan001: This PR currently doesn't support the multiple ESP case and will output an error if there's no single device with the label `EFI-SYSTEM`, but I guess it'll still be clearer to indicate which ESP even if we only have one?

cgwalters (approved):
/lgtm

cgwalters:
Moved the ProtectClock issue to https://github.com/systemd/systemd/issues/18142

src/efi.rs
jlebon: Opened https://github.com/coreos/bootupd/issues/132 to track this.

PR #131, opened 2021-01-06
Release 0.2.5

cgwalters:
This adds support for having the ESP not be mounted by
default, which we are going to do in Fedora CoreOS at least.


lucab (approved):
/lgtm

PR #149, opened 2021-02-03
doc: Add license_finder permitted licenses

cgwalters:
https://lists.fedoraproject.org/archives/list/rust@lists.fedoraproject.org/thread/UE4FU27OAHXQ2EOV5OFSXGHELUWZYSJM/
discusses license compliance as something a downstream
distribution should be doing.  Yes.  But also, this is something
that *upstreams* should be doing as part of their CI.

Looking around I found https://github.com/pivotal/LicenseFinder
which looks decent; it handles multiple languages, has
a container image, is FOSS itself etc.

I ran through it for bootupd and this lists the approvals.

The next step here is to validate this in CI.

### DIFF
--- a/doc/dependency_decisions.yml
+++ b/doc/dependency_decisions.yml
@@ -1,0 +1,37 @@
+---
+- - :permit
+  - MIT OR Apache-2.0
+  - :who: 
+    :why: 
+    :versions: []
+    :when: 2021-02-03 19:31:28.263225624 Z
+- - :permit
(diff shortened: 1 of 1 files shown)


lucab:
Self-note: tested with `docker run --rm -it -v "${PWD}:/scan" licensefinder/license_finder:6.10.1 bash -lc 'cd /scan && /LicenseFinder/bin/license_finder'`

lucab (approved):
/lgtm

PR #163, opened 2021-03-08
Add and use fn-error-context for more reliable error traces

cgwalters:
Effectively what we're doing here is creating a human-readable subset
of the stack trace.  This is nicer than having the calling functions
add `with_context()` because it's more verbose (gets duplicative at
each call site), easy to forget, etc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,6 +87,7 @@ dependencies = [
  "chrono",
  "clap",
  "env_logger",
+ "fn-error-context",
  "fs2",
  "hex",
  "lazy_static",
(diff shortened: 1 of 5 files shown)


cgwalters:
The background context[¹](https://twitter.com/cgwalters/status/1364988228114276359) here was a team discussion around errors, in particular using `unwrap()` in unit tests versus having them return `Result<()>` and the tradeoffs in debuggability.

Docs link to the crate I found: https://docs.rs/fn-error-context/0.1.1/fn_error_context/

PR #184, opened 2021-05-03
tests/e2e-update: do not rebuild binaries

lucab:
Avoid rebuilding the project from source, as binaries have already
been built in previous steps.

### DIFF
--- a/tests/e2e-update/e2e-update.sh
+++ b/tests/e2e-update/e2e-update.sh
@@ -80,7 +80,6 @@ if test -z "${e2e_skip_build:-}"; then
     echo "Building starting image"
     rm -f ${overrides}/rpm/*.rpm
     add_override grub2-2.04-22.fc32
-    (cd ${bootupd_git} && runv make && runv make install DESTDIR=${overrides}/rootfs)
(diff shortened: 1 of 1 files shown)


lucab:
This still fails in CI but in a different way (possibly due to copying binaries across different environments):
```
+ install_uefi
+ /usr/bin/bootupctl backend install --src-root=/tmp/rootfs/ostree/deploy/fedora-coreos/deploy/3cff9961aa0aec9c34db5cfb341f35c40be181d67c2814a89ca6c364b55ea729.0 /tmp/rootfs

/usr/bin/bootupctl: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /usr/bin/bootupctl)
```

cgwalters:
This has the same problem as https://github.com/coreos/rpm-ostree/pull/2806#issuecomment-831289329

IOW with this PR we silently stop testing the PR code at build time.  I think I'd prefer to wait for f34 cosa and leave CI broken rather than silently testing half of things.

lucab:
Ack, feel free to take over with the proper fix for this.

PR #204, opened 2021-07-28
workflows: bump toolchains; restrict repository access; fix build/lint warnings

bgilbert:
OCP 4.9 has Rust 1.49.


travier (approved):
LGTM (we still need to fix CI here)

PR #209, opened 2021-08-16
tests/e2e-update: Bump to f34 grub

cgwalters:
Need to tweak this to auto-detect gold versus updates, or
probably better just synthesize an update.


cgwalters:
Going to roll this into https://github.com/coreos/bootupd/pull/203

PR #228, opened 2021-10-01
ci: Don't run basic scenarios

cgwalters:
We're downgrading to an old grub2 build that is broken, so
we can't actually boot uefi after doing our hacks.

The correct thing here is to change bootupd CI to generate
synthetic RPMs instead of downloading old builds.  Will do that
at some point...


cgwalters:
https://github.com/coreos/bootupd/issues/229

bgilbert (approved):
WFM for now.

Issue #242, opened 2021-11-15
ci is broken again


cgwalters:
It's probably due to the Fedora 35 update, likely need to revise our pinned grub builds.

PR #250, opened 2021-11-29
ci: Bump to F35 grub

cgwalters:
This is the F35 equivalent of
https://github.com/coreos/bootupd/pull/203

Closes: https://github.com/coreos/bootupd/issues/242


cgwalters:
OK, this was a bit confusing because there are two separate tests, and *both* broke for unrelated reasons.

But now we're passing CI again!

lucab (approved):

PR #310, opened 2022-04-05
Replace structopt with clap 3.1

bgilbert:
structopt has been merged into clap.


cgwalters (approved):
Awesome, thanks so much for this!  This patch series looks like a good reference point for future porting.

Issue #333, opened 2022-05-11
CI is failing


bgilbert:
`e2e-update.sh` is [failing](https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/blue/organizations/jenkins/bootupd/detail/PR-332/1/pipeline):

```
[2022-05-11T01:14:00.815Z] + env COSA_DIR=/home/jenkins/agent/workspace/bootupd_PR-332 ./tests/e2e-update/e2e-update.sh
[2022-05-11T01:14:00.815Z] fatal: unsafe repository ('/home/jenkins/agent/workspace/bootupd_PR-332' is owned by someone else)
[2022-05-11T01:14:00.815Z] To add an exception for this directory, call:
[2022-05-11T01:14:00.815Z] 
[2022-05-11T01:14:00.815Z] 	git config --global --add safe.directory /home/jenkins/agent/workspace/bootupd_PR-332
script returned exit code 128
```

xref https://github.com/coreos/coreos-installer/pull/855

PR #334, opened 2022-05-13
ci: fix e2e-update test with current Git

bgilbert:
Tell Git that the current directory is safe, even though it's not owned by the current user.

Fixes https://github.com/coreos/bootupd/issues/333.


miabbott (approved):
LGTM

PR #346, opened 2022-07-18
Add `package.metadata.vendor-filter`

cgwalters:
To reduce the size of our vendor tarball, xref https://github.com/coreos/fedora-coreos-tracker/issues/1240


bgilbert:
`prepare-release.sh` would also need an update, right?

cgwalters:
> prepare-release.sh would also need an update, right?

Yes, thank you! :pray:

PR #379, opened 2022-09-30
efi: Rework/cleanup ESP mounting code

cgwalters:
build(deps): bump serde from 1.0.139 to 1.0.140

Bumps [serde](https://github.com/serde-rs/serde) from 1.0.139 to 1.0.140.
- [Release notes](https://github.com/serde-rs/serde/releases)
- [Commits](https://github.com/serde-rs/serde/compare/v1.0.139...v1.0.140)

---
updated-dependencies:
- dependency-name: serde
  dependency-type: direct:production
  update-type: version-update:semver-patch
...

Signed-off-by: dependabot[bot] <support@github.com>
---

Use inline `format!` in a few places

Just a minor cleanup.

---

Update lockfile

---

efi: Inline helper for path

It's not necessary to have a separate function for this; it
was called twice pointlessly.

---

efi/install: Don't pointlessly try to append destdir to mount path

I was doing a larger cleanup and was really confused by this bit of
code - we do `Path::new(destdir).join(/tmp/tmp.efiXXXXXX`) and
I couldn't understand how that worked until I realized that
`Path::join` will ignore the prefix for absolute paths.

This code has never actually used the destination root at all; we
look for the ESP in `/dev/disk/by-label`, which has nothing to
do with any target filesystem root.

We should just drop the target root command line argument here
at all, and document that we instead require an ESP.

But for now, drop the dead code.

---

efi: Rework/cleanup ESP mounting code

Right now we hardcode finding the ESP by the label we chose
for Fedora CoreOS.

This is preparatory work for generalizing that to support finding
the ESP on systems installed via Anaconda and ideally others.

---


jmarrero (approved):
lgtm

PR #386, opened 2022-10-13
ci: migrate to new directory and method names

jlebon:
The previous `fcos*` ones are deprecated.
See: https://github.com/coreos/coreos-ci-lib/pull/122

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -32,7 +32,7 @@ codestyle: {
 
 // Build FCOS and do a kola basic run
 // FIXME update to main branch once https://github.com/coreos/fedora-coreos-config/pull/595 merges
-cosaPod(runAsUser: 0, memory: "3072Mi", cpu: "4") {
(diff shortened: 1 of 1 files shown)


lucab:
The CI here was failing due hitting memory ceiling. I pushed a commit on top of the PR branch to tweak memory request, as we did in other upstream projects after https://github.com/coreos/coreos-ci-lib/pull/116.

jlebon:
I think 3G should normally be enough here. But https://github.com/coreos/coreos-ci-lib/pull/118 introduced a regression where reprovisioning tests (which require 4G) are run even if the caller only wanted to run a subset of tests. https://github.com/coreos/coreos-ci-lib/pull/120 will fix this.

lucab:
@jlebon ack; in the meanwhile the CI here is green. Do you want me drop the additional commit here, or to stamp and merge as is?

lucab (approved):

PR #396, opened 2022-11-01
Statically enable bootupd.socket

cgwalters:
The fact that we're a systemd unit is mostly an implementation detail so that we get a natural locking mechanism and can reuse systemd sandboxing.

If you have the software installed, we want the socket to be enabled.

Having to work through adding this into presets is just compounding pain for shipping it in multiple places.

Instead, install a static symlink.


jmarrero (approved):
lgtm

PR #412, opened 2022-12-12
efi: Reuse existing mountpoint if available

cgwalters:
This is an even better fix for https://github.com/coreos/bootupd/issues/367

Basically in coreos-assembler today we're already mounting the ESP, so let's just reuse that existing mount if it's present.  This is also what we want for other systems that pre-mount the ESP.

(At least for now - long term to support redundant ESPs we'll need
 to fully own all of this, but that can come later)

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -23,6 +23,9 @@ use crate::ostreeutil;
 use crate::util;
 use crate::util::CommandRunExt;
 
+/// Well-known paths to the ESP that may have been mounted external to us.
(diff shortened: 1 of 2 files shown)


jlebon (approved):

jlebon:
The test is trying to override with an old f35 grub2, which requires libgomp, which is no longer in FCOS after the f37 rebase. `cosa fetch` won't fetch it, and `cosa build` doesn't have Internet access. Easiest would be to bump the grub2 override to f37.

cgwalters:
Ah right thanks! 

OK let's try https://github.com/coreos/bootupd/pull/416

Issue #53, opened 2020-10-01
implement BIOS (grub)


cgwalters:
We have a lot of layers of abstraction predicated on the idea that there are multiple things to update, but we only update UEFI today.

In practice we also should update the BIOS MBR since there are reasons to do so sometimes, and it helps finish the story.

cgwalters:
Looking at https://github.com/coreos/coreos-assembler/blob/a356ccc729f4b20055fd887e659eeccc9429f805/src/create_disk.sh#L381 I notice we're actually using the assembler (Fedora) grub which may be distinct from the target system.

When we go to implement this I suspect bootupd will need to grow support for running `grub-install` from the source root in a basic container (`bwrap`?)

cgwalters:
A simple way to say this is we're ultimately just maintaining a little bit of metadata about which version of `grub2-install` was run.  There's a lot of layers of abstraction on top of that.

- Generating `/usr/lib/bootupd/updates/BIOS.json` will just have the version and timestamp of `grub2-tools` (much like how for EFI we have the combined version and timestamp of `grub2-efi-x64` and `shim-x64`)
- For this all we need to do on updates is run `grub2-install`; it should take care of updating the partition and installing files into `/boot`.  We shouldn't try to copy files on our own; this is in contrast to what we do for EFI in `/usr/lib/bootupd/updates/EFI/`
- We probably shouldn't try to track the checksum of the files.  For EFI, it's bootupd that copies the files.  But it's grub2-install that copies the files.  We wouldn't have the functionality of `bootupctl validate` in this model, but that's just an optional nice-to-have anyways.

cgwalters:
Ideally, just adding a new `src/bios.rs` and implementing `Component` will be enough for this!  There's lots of layers of abstraction in bootupd to support this.  But, this will be a "reality check" :smile:

PR #420, opened 2022-12-27
bios.rs: implement BIOS (grub)

HuijingHei:
bootupd: Prepare bootupd update directory for both components
- Prepare bootupd update directory which will save component metadata
files for both components.

---

Add bios.rs and implement generate_update_metadata
- This will generate `/usr/lib/bootupd/updates/BIOS.json`, which will have the version and timestamp of `grub2-tools`

---

bios.rs: implement BIOS (grub) install function
- Add `--device </dev/vda>` option to install BIOS bootloader (grub)
- If call function without `--device`, will only install EFI.

---

bios.rs: add `run_update` to support bootupctl update
- This is the same action as `install`
---

bios.rs: return `Skipped` when run `bootupctl validate` for BIOS

---
ppc64: fix build warning
- The fix looks really ugly. Just realize bootupd is not supported on ppc yet, maybe we will support in future.

---

aarch64: fix build warning

---

Fix https://github.com/coreos/bootupd/issues/53

### DIFF
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -1,5 +1,6 @@
 use std::io::prelude::*;
 use std::path::Path;
+use std::process::Command;
 
 use crate::component::*;
 use crate::model::*;
@@ -14,6 +15,40 @@ pub(crate) const GRUB_BIN: &str = "usr/sbin/grub2-install";
 pub(crate) struct Bios {}
 
 impl Bios {
+    // get target device for running update
+    fn get_device(&self) -> Result<String> {
+        let mut cmd: Command;
+        #[cfg(target_arch = "x86_64")]
+        {
+            // find /boot partition
+            let boot_dir = Path::new("/").join("boot");
+            cmd = Command::new("findmnt");
+            cmd.arg("--noheadings")
+                .arg("--output")
+                .arg("SOURCE")
+                .arg(boot_dir);
+            let partition = util::cmd_output(&mut cmd)?;
+
+            // lsblk to find parent device
+            cmd = Command::new("lsblk");
+            cmd.arg("--paths")
+                .arg("--noheadings")
+                .arg("--output")
+                .arg("PKNAME")
+                .arg(partition.trim());
+        }
+
+        #[cfg(target_arch = "powerpc64")]
+        {
+            // get PowerPC-PReP-boot partition
+            cmd = Command::new("realpath");
+            cmd.arg("/dev/disk/by-partlabel/PowerPC-PReP-boot");
+        }
+
+        let device = util::cmd_output(&mut cmd)?;
+        Ok(device)
+    }
+
     // Run grub2-install
     fn run_grub_install(&self, dest_root: &str, device: &str) -> Result<()> {
         let grub_install = Path::new("/").join(GRUB_BIN);
@@ -101,15 +136,21 @@ impl Component for Bios {
     }
 
     fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>> {
-        todo!();
+        get_component_update(sysroot, self)
     }
 
-    fn run_update(
-        &self,
-        sysroot: &openat::Dir,
-        current: &InstalledContent,
-    ) -> Result<InstalledContent> {
-        todo!();
+    fn run_update(&self, sysroot: &openat::Dir, _: &InstalledContent) -> Result<InstalledContent> {
+        let updatemeta = self.query_update(sysroot)?.expect("update available");
+        let device = self.get_device()?;
+        let device = device.trim();
+        self.run_grub_install("/", device)?;
+
+        let adopted_from = None;
+        Ok(InstalledContent {
+            meta: updatemeta,
+            filetree: None,
+            adopted_from,
+        })
     }
 
     fn validate(&self, current: &InstalledContent) -> Result<ValidationResult> {
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -1,0 +1,71 @@
+use std::io::prelude::*;
+use std::path::Path;
+
+use crate::component::*;
+use crate::model::*;
+use anyhow::{bail, Result};
+
+use crate::util;
+
+// grub2-install file path
+pub(crate) const GRUB_BIN: &str = "usr/sbin/grub2-install";
+
+#[derive(Default)]
+pub(crate) struct Bios {}
+
+impl Component for Bios {
+    fn name(&self) -> &'static str {
+        "BIOS"
+    }
+
+    fn install(&self, src_root: &openat::Dir, dest_root: &str) -> Result<InstalledContent> {
+        todo!();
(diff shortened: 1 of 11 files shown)


src/bios.rs
```diff
@@ -0,0 +1,115 @@
...
+            bail!("Failed to invoke rpm -qf");
+        }        
+
+        let meta = util::parse_metadata(rpmout.stdout).unwrap();
+        write_update_metadata(sysroot_path, self, &meta)?;
+        Ok(meta)
+    }
+
+    fn query_adopt(&self) -> Result<Option<Adoptable>> {
+        unreachable!()
```
cgwalters, line 89: One minor thing here is to use the `todo!()` - it's intended for this.

src/bios.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>> {
+        get_component_update(sysroot, self)
+    }
+
+    fn run_update(
+        &self,
+        sysroot: &openat::Dir,
+        current: &InstalledContent,
+    ) -> Result<InstalledContent> {
+        unreachable!()
```
cgwalters, line 109: I think performing an update here is exactly the same as a new install.

src/bios.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    fn validate(&self, current: &InstalledContent) -> Result<ValidationResult> {
+        unreachable!()
```
cgwalters, line 113: As discussed we could probably keep this unimplemented, though it should return an error or so?

src/util.rs
```diff
@@ -83,3 +87,38 @@ pub(crate) fn ensure_writable_mount<P: AsRef<Path>>(p: P) -> Result<()> {
     }
     Ok(())
 }
+
+/// Parse the output of `rpm -q`
+pub(crate) fn parse_metadata(stdout: Vec<u8>) -> Result<ContentMetadata> {
```
cgwalters, line 92: Let's call this `parse_rpm_metadata()` perhaps?  We could also land this as a preparatory PR now!

src/bios.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn generate_update_metadata(&self, sysroot_path: &str) -> Result<ContentMetadata> {
+        let grub_install = Path::new(sysroot_path).join(GRUB_BIN);
+        if !grub_install.exists() {
+            bail!("Failed to find {:?}", grub_install);
+        }
+        
+        // Query the rpm database and list the package and build times for all the
+        // files in the EFI system partition. If any files are not owned it is considered
+        // and error condition.
+        let rpmout = {
```
cgwalters, line 71: Now that I look at this we could probably extract this part to a helper function too.  Do you want to try that too?  Or I can do it if you like.

cgwalters: SGTM!

src/util.rs
```diff
@@ -83,3 +88,62 @@ pub(crate) fn ensure_writable_mount<P: AsRef<Path>>(p: P) -> Result<()> {
...
+
+    match path.file_name().expect("filename").to_str() {
+        Some("EFI") => {
+            let efidir = openat::Dir::open(path)?;
+            c.args(filenames(&efidir)?.drain().map(|mut f| {
+                f.insert_str(0, "/boot/efi/EFI/");
+                f
+            }));
+        }
+        Some("grub2-install") => {
```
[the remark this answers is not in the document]
HuijingHei: That makes sense! I will try, thanks!

HuijingHei:
Will update based on https://github.com/coreos/bootupd/pull/423

src/bios.rs
HuijingHei: Agree, will try

src/bios.rs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        let boot_dir = Path::new("/").join("boot");
+
+        #[cfg(target_arch = "x86_64")]
+        let r = std::process::Command::new(grub_install)
+            .args(&["--target", "i386-pc"])
+            .args(&["--boot-directory", boot_dir.to_str().unwrap()])
+            .args(&["--modules", "mdraid1x"])
+            .arg(&device)
```
HuijingHei, line 123: Not sure how to get `device` name for update, any suggestions?

cgwalters: Good question!  We could look at the backing device of `/boot` (or `/`) as a default.  (For example, forking off `findmnt -J /boot`, then we need to find its parent device.  There's a lot of related code in https://github.com/coreos/coreos-installer/blob/main/src/bin/rdcore/rootmap.rs

HuijingHei: Good suggestion, will look at it, thanks!

cgwalters:
> error: Server sent an empty reply

Try using `journalctl -u bootupd` to get more information

src/bios.rs
```diff
@@ -0,0 +1,161 @@
...
+        Ok(meta)
+    }
+
+    fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>> {
+        get_component_update(sysroot, self)
+    }
+
+    #[allow(unused_variables)]
+    fn query_adopt(&self) -> Result<Option<Adoptable>> {
+        todo!();
```
cgwalters, line 132: In theory what we should do here is query for a BIOS partition - possibly if we booted from one.  But for now...I think we can just do:

```
            return Ok(None);
```
or so.

src/efi.rs
```diff
@@ -257,10 +262,6 @@ impl Component for Efi {
 
             // Fork off mv() because on overlayfs one can't rename() a lower level
             // directory today, and this will handle the copy fallback.
-            let parent = dest_efidir
```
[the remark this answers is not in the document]
HuijingHei: SGTM, separated PR is https://github.com/coreos/bootupd/pull/427, will rebase when it is merged.

src/bios.rs
HuijingHei: Return `Skipped` for `BIOS` instead, does that make sense?
```
[root@cosa-devsh ~]# bootupctl validate
Skipped: BIOS
Validated: EFI
```

src/bios.rs
HuijingHei: Nice catch, thanks! Will update in future patch.

HuijingHei:
Wait, I am testing on `ppc64le`, and there are some errors need to be fixed.

cgwalters (approved):
Nice work on this!

I hadn't even realized when we started this how complex adding ppc64le would be from the conditional-compilation perspective.  Thanks for plowing through all those conditionals.

I think there's some smaller things to clean up here but we can do that post-merge

Thanks so much for your work on this!

HuijingHei:
Thanks @jlebon! Also thanks @cgwalters for lots of help!

PR #423, opened 2023-01-12
util.rs: move some code to function to reuse

HuijingHei:
- Move code to func `parse_rpm_metadata` to parse the output of `rpm -q`
- Move code to func `rpm_query` to query rpm database, list the package and build times

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -5,7 +5,6 @@
  */
 
 use std::cell::RefCell;
-use std::collections::{BTreeMap, BTreeSet};
 use std::io::prelude::*;
 use std::os::unix::io::AsRawFd;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


src/efi.rs
```diff
@@ -267,56 +264,17 @@ impl Component for Efi {
...
-            let mut c = ostreeutil::rpm_cmd(sysroot_path);
-            c.args(&["-q", "--queryformat", "%{nevra},%{buildtime} ", "-f"]);
-            c.args(util::filenames(&src_efidir)?.drain().map(|mut f| {
-                f.insert_str(0, "/boot/efi/EFI/");
-                f
-            }));
-            c
-        }
-        .output()?;
+        let mut rpmout = util::rpm_query(sysroot_path, &dest_efidir).unwrap();
```
cgwalters, line 270: This should use `?` instead of `unwrap()`.

cgwalters (commented):
Thanks!  One minor bit, otherwise looks good!

cgwalters (approved):

PR #426, opened 2023-01-23
Copy xtask packaging bits from bootc

cgwalters:
This copies the logic from https://github.com/containers/bootc/pull/36/commits/b274b315c24de933e67a538a5fccbcf904854af8 into this project with the same motivation of enabling COPR builds.


jmarrero (approved):
lgtm

PR #427, opened 2023-02-01
bootupd: Prepare bootupd update directory for both components

HuijingHei:
Prepare bootupd update directory which will save component metadata files for both components.


HuijingHei:
Close this, and will add this to https://github.com/coreos/bootupd/pull/420

PR #433, opened 2023-02-25
bios: Always add `part_gpt`

cgwalters:
This fixes https://github.com/containers/bootc/issues/64

I didn't chase through what was going wrong in grub; there's a whole lot of layers here.  But for the CoreOS use case we know the modules we need, so we should specify them.


jmarrero (approved):
lgtm

PR #439, opened 2023-03-15
README.md: Update status to reflect current reality

cgwalters:
Came up re https://github.com/coreos/fedora-coreos-tracker/issues/1441

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,18 @@ that's for tools like `grubby` and `ostree`.
 
 ## Status
 
-Currently a work in progress and is not ready to ship for production
-updates, but early feedback on the design is appreciated!
+bootupd supports updating GRUB and shim for UEFI firmware on
+x86_64 and aarch64, and GRUB for BIOS firmware on x86_64.
+The project is [deployed in Fedora CoreOS](https://docs.fedoraproject.org/en-US/fedora-coreos/bootloader-updates/) and derivatives,
+and is also used by the new [`bootc install`](https://github.com/containers/bootc/#using-bootc-install)
+functionality.  The bootupd CLI should be considered stable.
+
+bootupd does not yet perform updates in a way that is safe
+against a power failure at the wrong moment, or
+against a buggy bootloader update that fails to boot
+the system.
+
+Therefore, by default, bootupd updates the bootloader only when manually instructed to do so.
 
 ## Relationship to other projects
 


README.md
```diff
@@ -31,8 +31,10 @@ that's for tools like `grubby` and `ostree`.
 
 ## Status
 
-Currently a work in progress and is not ready to ship for production
-updates, but early feedback on the design is appreciated!
+This project is deployed in e.g. Fedora CoreOS and derivatives, and is also used by the new
+`bootc install` functionality https://github.com/containers/bootc/#using-bootc-install
```
jlebon, line 35: Minor/optional:

```suggestion
[`bootc install` functionality](https://github.com/containers/bootc/#using-bootc-install).
```

jlebon (approved):

bgilbert:
Should we mention that bootupd doesn't yet perform updates automatically?  Folks may assume otherwise if we don't say anything.

cgwalters:
The FCOS side here is already covered by https://docs.fedoraproject.org/en-US/fedora-coreos/bootloader-updates/ right?

bgilbert:
Sure.  But the upstream project status should presumably reflect the upstream portion of this.

Disabled auto-merge to avoid this comment creating a race condition.

jlebon:
Maybe something like:

> As mentioned lower down in the [Q&A](#questions-and-answers), there is no method currently to transactionally update the bootloader. As such, bootupd itself does not by default update the bootloader.

?

On that topic, ISTM like bootupd could ship the systemd unit in https://docs.fedoraproject.org/en-US/fedora-coreos/bootloader-updates/ itself as e.g. `bootupd-simple-update.service` or something? Then it's even easier for users to enable that if they're OK with its semantics.

bgilbert:
I agree that it'd be good to ship the unit upstream, even if disabled by default.

How about text like this:

> bootupd supports updating GRUB and shim for UEFI firmware on x86_64 and aarch64, and GRUB for BIOS firmware on x86_64.  The project is deployed in Fedora CoreOS and derivatives, and is also used by the new
[`bootc install`](https://github.com/containers/bootc/#using-bootc-install) functionality.  The bootupd CLI should be considered stable.
>
> bootupd does not yet perform updates in a way that is safe against a power failure at the wrong moment, or against a buggy bootloader update that fails to boot the system.  Therefore, by default, bootupd updates the bootloader only when manually instructed to do so.

cgwalters:
LGTM, updated to that text.

cgwalters:
Nightly Rust failure is https://github.com/rust-lang/rust/issues/109199#issuecomment-1471205592
I don't see a lot of value in testing with nightly here, probably will remove it or at least make it non-gating.

And this is a network flake
/override continuous-integration/jenkins/pr-merge

PR #447, opened 2023-03-21
Update to nix 0.26

cgwalters:
This one is important to keep up with, it's like part of a libc.


cgwalters:
Hmm, I must be doing something wrong here in the iov porting, but not seeing what...punting this one for now.

Issue #229, opened 2021-10-01
rework CI to generate synthetic RPMs


cgwalters:
See https://github.com/coreos/bootupd/pull/228 and https://github.com/coreos/bootupd/pull/203

Basically to test upgrades, we're downgrading to an older grub build which is broken.  We're just testing that the synthetic process of changing files actually works, but we're not currently testing rebooting into it.

Instead, we should download the working RPM and tweak it by:

 - adding files (safe)
 - Figuring out a way to mutate e.g. shim or grub binaries in a benign way that doesn't affect functionalty (this will (hopefully) not work with Secure Boot though), but at least we can reboot in non-SB UEFI
 -

bgilbert:
Once we have this, we can revert #228.

dralley:
So, the `rpm` crate that you already optionally depend on (which I maintain) has the ability to parse and build RPMs.

If you have a use case for rebuilding an RPM with modifications, I can see if I can accommodate that.  Relatedly I've heard some ideas from kernel folks that it would be nice to be able to build different variants of the kernel package with different sets of kernel modules, or perhaps rebuild in-place to strip out unneeded modules.  Kinda similar?

PR #462, opened 2023-04-11
Don't adopt systemd boot loaders

dbnicholson:
Skip adoption if systemd-boot or systemd-stub are used since `bootctl` will handle updating them. The 2nd commit is IMO correct but more controversial. I needed it to test that our sd-boot system would be detected and skipped. It did work in that case, though:

```
Apr 11 10:56:49 endless bootupd[3086]: [DEBUG bootupd::efi] Reusing existing "/boot"
Apr 11 10:56:49 endless bootupd[3086]: [TRACE bootupd::efi] No CoreOS aleph detected
Apr 11 10:56:49 endless bootupd[3086]: [TRACE bootupd::efi] Skipping adoption for "systemd-boot 251.2+dev56.9738984-22bem1"
```

I'm not that familiar with writing rust tests, but I think the UTF-16 string function could be written to take an optional root and then a test that writes various stuff into `<testroot>/sys/firmware/efi/efivars/<somevar>` could be added.

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -12,6 +12,7 @@ use std::process::Command;
 
 use anyhow::{bail, Context, Result};
 use openat_ext::OpenatDirExt;
+use widestring::U16CString;
 
 use crate::component::*;
 use crate::filetree;
@@ -26,6 +27,10 @@ pub(crate) const ESP_MOUNTS: &[&str] = &["boot/efi", "efi"];
 /// The ESP partition label on Fedora CoreOS derivatives
 pub(crate) const COREOS_ESP_PART_LABEL: &str = "EFI-SYSTEM";
 pub(crate) const ANACONDA_ESP_PART_LABEL: &str = "EFI\\x20System\\x20Partition";
+
+/// Systemd boot loader info EFI variable names
+const LOADER_INFO_VAR_STR: &str = "LoaderInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
+const STUB_INFO_VAR_STR: &str = "StubInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
 
 #[derive(Default)]
 pub(crate) struct Efi {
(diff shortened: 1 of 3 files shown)


src/efi.rs
```diff
@@ -105,6 +110,70 @@ impl Efi {
     }
 }
 
+/// Convert a nul-terminated UTF-16 byte array to a String.
+fn string_from_utf16_bytes(slice: &[u8]) -> String {
+    // For some reason, systemd appends 3 nul bytes after the string.
+    // Drop the last byte if there's an odd number.
+    let size = slice.len() / 2;
+    let v: Vec<u16> = (0..size)
+        .map(|i| u16::from_ne_bytes([slice[2 * i], slice[2 * i + 1]]))
```
cgwalters, line 119: I think we can use https://doc.rust-lang.org/std/primitive.slice.html#method.chunks here

src/efi.rs
```diff
@@ -105,6 +110,70 @@ impl Efi {
     }
 }
 
+/// Convert a nul-terminated UTF-16 byte array to a String.
+fn string_from_utf16_bytes(slice: &[u8]) -> String {
+    // For some reason, systemd appends 3 nul bytes after the string.
```
cgwalters, line 115: Hmm, messy.  But I wonder if we should care about this?  Maybe instead we return a `Vec<u8>` since in the end all we care about is if it starts with the ASCII string `systemd` right?  

OK as is though.

cgwalters (approved):
Looks sane overall, just a few minor nits

cgwalters (approved):
I think this is fine as is, we can do followups later.

cgwalters:
Tried rebasing this just to rerun CI and see if the failure is somehow fixed in main but I think the problem is rooted in https://github.com/coreos/bootupd/issues/229

src/efi.rs
dbnicholson: You can't just send an array of bytes (which you already have from reading the file contents) back since they're be misinterpreted as UTF-8 since every other byte is `nul`. That's why you have to convert it to `Vec<u16>` and parse it as UTF-16. I don't believe there's any spec that says you have to write UTF-16 strings to EFI variables. I think that's just a convention systemd follows since that's how other vendors do it. I think I read a comment in the systemd code that they append 3 `nul` bytes so that in case they start parsing in the middle of the buffer they'll still end up with a valid C string. However, stripping an odd numbered trailing byte seems like good practice.

When I wrote roughly the same code in Python where I didn't have to change types, I walked backwards from the end of the buffer until I found no more pairs of `nul` bytes. Here I let `U16CString` handle that part.

src/efi.rs
dbnicholson: Oh, that does look nicer. My rust noobness shows through.

src/efi.rs
```diff
@@ -105,6 +110,70 @@ impl Efi {
...
+
+/// Read a nul-terminated UTF-16 string from an EFI variable.
+fn read_efi_var_utf16_string(name: &str) -> Option<String> {
+    let efivars = Path::new("/sys/firmware/efi/efivars");
+    if !efivars.exists() {
+        log::warn!("No efivars mount at {:?}", efivars);
+        return None;
+    }
+    let path = efivars.join(name);
+    if !path.exists() {
```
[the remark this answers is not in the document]
dbnicholson: I didn't notice that.

src/efi.rs
cgwalters: Yes...hm.  What about going the other way, and converting from `systemd` in UTF-8 to UTF-16, and then using https://doc.rust-lang.org/std/primitive.slice.html#method.starts_with ?

src/efi.rs
cgwalters: Your code is great honestly!  Learning a language (programming or not) is always a journey; even the relatively small `std` in Rust has a lot of surface area and idioms.

Issue #445, opened 2023-03-20
add extra message about starting point when performing update 


dustymabe:
It would be nice if we could get another piece of information when running an update. Right now:

```
[core@dustymabe ~]$ sudo bootupctl update
Updated EFI: grub2-efi-aa64-1:2.06-88.fc37.aarch64,shim-aa64-15.6-2.aarch64
```

It would be nice when looking at issues later to also know what version the update started at.. i.e. something like:

```
[core@dustymabe ~]$ sudo bootupctl update
Initial Installed: grub2-efi-aa64-1:2.06-2.fc34.aarch64,shim-aa64-15.4-4.aarch64
Updated EFI: grub2-efi-aa64-1:2.06-88.fc37.aarch64,shim-aa64-15.6-2.aarch64
```

PR #472, opened 2023-05-05
Print previous version too when updating

cgwalters:
This was easy.

Closes: https://github.com/coreos/bootupd/issues/445


travier:
Updated and tweaked in https://github.com/coreos/bootupd/pull/642

PR #479, opened 2023-05-30
build-sys: Use new tier = 2 for cargo-vendor-filterer

cgwalters:
While the architecture list here was right today, there's no reason not to just expand our vendoring set.


bgilbert (approved):
LGTM

Issue #467, opened 2023-04-24
Internal error when `bootupctl adopt-and-update` on Fedora Silverblue 38


xlejo:
Hi!

When I run the command `bootupctl adopt-and-update` on Fedora Silverblue 38 (build with unified core) shows the next error.

```
# bootupctl adopt-and-update
error: internal error: Failed adopt and update: applying filesystem changes: removing fedora/BOOTIA32.CSV: No such file or directory (os error 2)
```

Seems to look for 32-bit binaries. If I run it multiple times, the file to remove changes.

```
# bootupctl adopt-and-update                     
error: internal error: Failed adopt and update: applying filesystem changes: removing fedora/mmia32.efi: No such file or directory (os error 2)
# bootupctl adopt-and-update
error: internal error: Failed adopt and update: applying filesystem changes: removing BOOT/BOOTIA32.EFI: No such file or directory (os error 2)
```

I search the files in `/boot/efi/EFI/fedora` but I don't find it (all 64-bit binaries).

```
# ls -l /boot/efi/EFI/fedora/
total 7404
-rwx------. 1 root root     110 Aug 18  2022 BOOTX64.CSV
drwx------. 2 root root    4096 Aug 18  2022 fonts
drwx------. 2 root root    4096 Apr 21 09:14 fw
-rwx------. 1 root root   64368 Dec 31  1979 fwupdx64.efi
-rwx------. 1 root root     144 Aug 18  2022 grub.cfg
-r-x------. 1 root root 3530048 Apr 24 09:23 grubx64.efi
-r-x------. 1 root root  857248 Apr 24 09:23 mmx64.efi
-r-x------. 1 root root  946712 Apr 24 09:23 shim.efi
-r-x------. 1 root root  946712 Apr 24 09:23 shimx64.efi
-rwx------. 1 root root 1204496 Aug 18  2022 shimx64-fedora.efi 
```

This is the output of `strace bootupctl adopt-and-update`

```
# strace bootupctl adopt-and-update
execve("/usr/bin/bootupctl", ["bootupctl", "adopt-and-update"], 0x7ffc1eefb488 /* 22 vars */) = 0
brk(NULL)                               = 0x557a2ef1f000
arch_prctl(0x3001 /* ARCH_??? */, 0x7ffeae8fd960) = -1 EINVAL (Invalid argument)
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
[...]

PR #528, opened 2023-09-14
Fix removal in non-existent files

xlejo:
Tested in Fedora Silverblue 38 upgraded from 34 (upgrades from 36 isn't affected).

Fixes https://github.com/coreos/bootupd/issues/467.

### DIFF
--- a/src/filetree.rs
+++ b/src/filetree.rs
@@ -337,7 +337,7 @@ pub(crate) fn apply_diff(
     if !opts.skip_removals {
         for path in diff.removals.iter() {
             destdir
-                .remove_file(path)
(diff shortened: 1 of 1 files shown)


cgwalters:
Why is Silverblue shipping `shim-ia32`?  That seems like one root bug here.  Is it being somehow removed later?  

Oh, I think I see; Anaconda may be just ignoring it when copying the bootloader data?

Anyways, I guess this is OK (and thank you so much for working on it!).  It's a relatively safe change, but again IMO the real bug lies in the SB config.

cgwalters (approved):

PR #535, opened 2023-09-20
install: Add support for selecting components

cgwalters:
So one can use e.g. `--component=EFI` to only install UEFI.


cgwalters:
I think this one is good to merge if someone could look!

jmarrero (approved):
lgtm

Issue #468, opened 2023-04-24
how to workaround rpm dependency?


akdev1l:
Hi bootupd team,

I love the idea behind bootc so I've been trying to create a bootc image based on archlinux. 

As part of creating a bootc image `bootupd backend generate-metadata` is called but this fails on arch linux as there is not rpm binary. 

Is there a way I can work around this dependency? I am happy to hack around this in my image as it is just a proof of concept but not sure what to do

Also the readme suggest bootupd should be OS agnostic but looking for `rpm` means it is not really agnostic :-( are there plans to remove this dependency?

cgwalters:
Absolutely, we can support other backends here in bootupd.  It should be relatively straightforward to do.  Do you know any Rust?

akdev1l:
I know a little bit of rust, I can probably try and get it working on arch linux if you can provide some guidance on:

1. what is bootupd expecting from the package manager?
2. what is the purpose of the metadata being generated?
3. how does it relate to the grub configuration?

we were able to hack around this dependency by reusing generated files from fedora but naturally the resulting image couldn't boot (I think because the metadata pointed to Fedora-specific grub configuration?)

cgwalters:
We basically just need version information.  I think instead of hardcoding knowledge of rpm in bootupd, we could try having an external script that can be provided by the OS output it?

bigpod98:
@cgwalters @akdev1l any movement on this

symbioquine:
@cgwalters I'm curious about this too. Can you provide a bit more detail about what you mean when you say "We basically just need version information"?

Specifically, what I'm wondering is what it would take to boot existing container images (say Arch Linux) without any of the package layering stuff? I don't mind needing to build a custom image in order to install things. It looks like someone has actually already been [trying to get that working](https://github.com/M1cha/bootc-archlinux). Maybe you can comment about whether that would be expected to work?

cgwalters:
I was poking again at this; looking specifically at Debian for example, it looks like the deb already in `/usr`: `/usr/lib/grub/x86_64-efi/monolithic/grubx64.efi`.

So this one then also relates a lot to https://github.com/coreos/bootupd/issues/766 in that we should potentially just be able to take over or move what happens in the maintainer scripts? I am not finding offhand how that happens, I see at least

```
cat /var/lib/dpkg/info/grub-efi-amd64-signed.postinst 
#! /bin/sh
set -e

. /usr/share/debconf/confmodule

case $1 in
    configure)
[...]

PR #536, opened 2023-09-21
Import static grub configuration

cgwalters:
Pull this from coreos-assembler; prep for moving ownership of it into bootupd.

In the short term, this will allow install tools to copy it manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,3 +44,6 @@ install: install-units
 	ln -f ${DESTDIR}$(LIBEXECDIR)/bootupd ${DESTDIR}$(PREFIX)/bin/bootupctl
 	install -d "${DESTDIR}$(PREFIX)/lib/systemd/system/multi-user.target.wants"
 	ln -s ../bootupd.socket "${DESTDIR}$(PREFIX)/lib/systemd/system/multi-user.target.wants"
+
+install-grub-static:
+	install -D -t ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static src/grub2/*.cfg
--- a/contrib/packaging/bootupd.spec
+++ b/contrib/packaging/bootupd.spec
@@ -44,11 +44,21 @@ replace-with = "vendored-sources"
 directory = "vendor"
 EOF
 
+%package grub2-static
+Summary: Static grub2 configuration files for CoreOS-alike systems
+
+%description grub2-static
+This package includes static grub2 configuration files.
+
+%files grub2-static
+%{_prefix}/lib/bootupd
+
 %build
 cargo build --release
 
 %install
 %make_install INSTALL="install -p -c"
+make install-grub-static DESTDIR=%{?buildroot} INSTALL="%{__install} -p"
 
 %post        -n %{crate}
 %systemd_post bootupd.service bootupd.socket
--- a/src/grub2/README.md
+++ b/src/grub2/README.md
@@ -1,0 +1,3 @@
+# Static GRUB configuration files
+
+These static files were taken from https://github.com/coreos/coreos-assembler/blob/5824720ec3a9ec291532b23b349b6d8d8b2e9edd/src/grub.cfg
--- a/src/grub2/grub-static-efi.cfg
+++ b/src/grub2/grub-static-efi.cfg
@@ -1,0 +1,19 @@
+if [ -e (md/md-boot) ]; then
+  # The search command might pick a RAID component rather than the RAID,
+  # since the /boot RAID currently uses superblock 1.0.  See the comment in
+  # the main grub.cfg.
+  set prefix=md/md-boot
+else
+  if [ -f ${config_directory}/bootuuid.cfg ]; then
+    source ${config_directory}/bootuuid.cfg
+  fi
+  if [ -n "${BOOT_UUID}" ]; then
+    search --fs-uuid "${BOOT_UUID}" --set prefix --no-floppy
+  else
+    search --label boot --set prefix --no-floppy
+  fi
+fi
+set prefix=($prefix)/grub2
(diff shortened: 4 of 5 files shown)


src/grub2/grub-static-coreos.cfg
```diff
@@ -0,0 +1,97 @@
...
+    insmod vga
+    insmod video_bochs
+    insmod video_cirrus
+  fi
+}
+
+# tracker: https://github.com/coreos/fedora-coreos-tracker/issues/805
+if [ -f $prefix/platform.cfg ]; then
+  source $prefix/platform.cfg
+fi
```
cgwalters, line 63: Need to change coreos-assembler to do this, see https://github.com/coreos/coreos-assembler/blob/5824720ec3a9ec291532b23b349b6d8d8b2e9edd/src/create_disk.sh#L83

cgwalters:
There's zero risk to this, it's just shipping some files in a subpackage.

The larger set of work is adding support for importing this into being "owned" by bootupd.  But that can come *after* centralizing the static config into this package, where it can be reused by both osbuild and anaconda style flows.

jlebon:
Hmm, if we're going to move the GRUB configs, I think they belong better in e.g. https://github.com/coreos/fedora-coreos-config instead. There's a lot of integration points between the two and other distros that want bootupd likely will have their own GRUB config. (I guess we could set up some inheritance here with `source`, though not sure it's worth the complexity.)

For the update case, we could easily ship an overlay somewhere under e.g. `/usr/lib/grubcfgs/` and bootupd could learn to look there.

cgwalters:
> Hmm, if we're going to move the GRUB configs, I think they belong better in e.g. https://github.com/coreos/fedora-coreos-config instead. 

The main problem is that doesn't come as an RPM today, and that causes friction with tooling currently oriented towards that as primary input sources (e.g. `rpm-ostree compose image` and Image Builder namely).  It's not that RPMs are amazing or anything, but from the pungi side (which sagano is aiming to integrate with) adding more git repositories complicates the picture and current "hermetic" nature of things.

jlebon:
We're already going to have to solve the RPM problem for a lot of overlay files we have in https://github.com/coreos/fedora-coreos-config so I'd just glob it as part of that effort. If the goal of bootupd is distro-independence, then IMO it doesn't make sense to have a GRUB config with opinionated things like `bootuuid`, Ignition firstboot stuff, our RAID1 setup, user dropins, etc...

cgwalters:
Today, Image Builder has a separate copy of a grub config it uses, see https://github.com/osbuild/osbuild/blob/e48360e01c84ef702b6e5d34a26cedddc55a60c0/stages/org.osbuild.grub2.legacy#L204

And it'd be a giant effort to somehow wedge github.com/coreos/fedora-coreos-config into that.  And as far as an RPM package place, given we already have bootupd and the plan is to move ownership there, it really seems simpler to move it here for now.

Other distributions can just ignore it (it's not even installed by default!).  Other ostree/container-native distributions don't have the problem with RPM-obsessed build systems :smile:

cgwalters:
OK I dropped out the Ignition glue from this PR, so what we can do is use this even more minimal static version for Sagano work, and rework the version that lives in FCOS to write its additions into `platform.cfg` instead.

>  it doesn't make sense to have a GRUB config with opinionated things like bootuuid, Ignition firstboot stuff, our RAID1 setup, user dropins, etc...

bootuuid is used by [bootc install](https://github.com/containers/bootc/blob/e67c89504b816a3cdff6305fa96e63d4fafad889/lib/src/bootloader.rs#L16) today.

Ignition firstboot: dropped!

RAID1: eh...it aligns with also supporting mirrored EFI which we also want in bootupd.

user dropins: Also something desired for anyone using a static grub config I think.


IOW I don't want to block this demo I'm trying to put together of installing a Sagano container w/Anaconda (and bootupd) on refactoring the FCOS structure entirely by packaging fedora-coreos-config.  And even if we did that, the converse is true that Sagano grub.cfg shouldn't have Ignition, so we still need two versions.

Hence I think this model of changing FCOS to inherit from this version should work out well.

cgwalters:
Incidentally also, this project is still currently hardcoded to work with RPM https://github.com/coreos/bootupd/issues/468 which is a far larger distro-independence problem, and one that will be hard to get away from.  There's just going to be some ugly distro stuff in this project probably forever.

jlebon (approved):
Yeah, that looks better. A bit apprehensive about setting up more inheritance chains here for this but I'm cool also evaluating the full impact when we get to this in FCOS land if this is considered demo/WIP material.

Again, what I think we should aim for is to have GRUB logic located closer to the code that leverages it. E.g. the thing that writes the bootuuids is called by f-c-c. A lot, but not all, of password and Ignition integration live in f-c-c. Once we get to the packaging hurdle, these things would go together.

Having the update *mechanism* live in bootupd sounds great, and then f-c-c can leverage that by placing its file in some predefined directory or something?

cgwalters:
> E.g. the thing that writes the bootuuids is called by f-c-c. 

Only in a "golden image" case.  The Anaconda and bootc paths want to ensure that the system is set up properly with uuids from the start.

(And actually, I want to change Anaconda to use `bootc install-to-filesystem` which would then mean it's only bootc)

cgwalters:
Also thanks for the review and debate here!  It's a tricky topic we're navigating and I think while the debate felt contentious at first I actually think the result ended up working well!  Or at least it hopefully will :smile:

cgwalters:
Also xref https://github.com/osbuild/osbuild/blob/4e5de3808e2bf56e1060937c4d5e1796df441fa4/stages/org.osbuild.grub2#L280

PR #540, opened 2023-09-29
Packagesystem cleanup

cgwalters:
Refactor rpm bits into packagesystem module

Prep for cleaning this up and making it more generic
to support non-rpm systems in a pluggable way.

---

De-duplicate package querying

There's no reason to have two separate functions here.

---

packagesystem: Add a unit test

On general principle.

---


jmarrero (approved):
lgtm

PR #541, opened 2023-09-29
backend: Make --generate-update-metadata require a sysroot of /

cgwalters:
As of right now all known callers of this pass `/`.  More generally we now expect bootupd to run in a container inside its target root; never outside of it.

xref https://github.com/coreos/coreos-assembler/pull/3631


jmarrero (approved):
lgtm

PR #543, opened 2023-10-03
 Add support for installing static grub configs

cgwalters:
grub: Support multiple platform files

In practice, just one should be enough but let's expand
in case.

---

Add support for installing static grub configs

Currently these are duplicated in osbuild and coreos-assembler.
We will aim to deduplicate them here.

Ideally we'd add support for "day 2" updates of these; I started
on a patch for that but it's sadly messy.

This is an incremental improvement.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,4 +46,5 @@ install: install-units
 	ln -s ../bootupd.socket "${DESTDIR}$(PREFIX)/lib/systemd/system/multi-user.target.wants"
 
 install-grub-static:
-	install -D -t ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static src/grub2/*.cfg
+	install -m 644 -D -t ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static src/grub2/*.cfg
+	install -m 755 -d ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static/configs.d
--- a/src/grubconfigs.rs
+++ b/src/grubconfigs.rs
@@ -1,0 +1,111 @@
+use std::fmt::Write;
+use std::os::unix::prelude::OsStrExt;
+use std::path::{Path, PathBuf};
+
+use anyhow::{anyhow, Context, Result};
+use fn_error_context::context;
+use openat_ext::OpenatDirExt;
+
+/// The subdirectory of /boot we use
+const GRUB2DIR: &str = "grub2";
+const CONFIGDIR: &str = "/usr/lib/bootupd/grub2-static";
+const DROPINDIR: &str = "configs.d";
+
+#[context("Locating EFI vendordir")]
+pub(crate) fn find_efi_vendordir(efidir: &openat::Dir) -> Result<PathBuf> {
+    for d in efidir.list_dir(".")? {
+        let d = d?;
+        if d.file_name().as_bytes() == b"BOOT" {
+            continue;
+        }
+        let meta = efidir.metadata(d.file_name())?;
+        if !meta.is_dir() {
+            continue;
+        }
+        return Ok(d.file_name().into());
+    }
+    anyhow::bail!("Failed to find EFI vendor dir")
+}
+
+/// Install the static GRUB config files.
+#[context("Installing static GRUB configs")]
+pub(crate) fn install(target_root: &openat::Dir, efi: bool) -> Result<()> {
+    let bootdir = &target_root.sub_dir("boot").context("Opening /boot")?;
+
+    let mut config = std::fs::read_to_string(Path::new(CONFIGDIR).join("grub-static-pre.cfg"))?;
+
+    let dropindir = openat::Dir::open(&Path::new(CONFIGDIR).join(DROPINDIR))?;
+    // Sort the files for reproducibility
+    let mut entries = dropindir
+        .list_dir(".")?
+        .map(|e| e.map_err(anyhow::Error::msg))
+        .collect::<Result<Vec<_>>>()?;
+    entries.sort_by(|a, b| a.file_name().cmp(b.file_name()));
+    for ent in entries {
+        let name = ent.file_name();
+        let name = name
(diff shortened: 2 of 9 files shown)


cgwalters:
OK this is now tested in concert with https://github.com/coreos/coreos-assembler/pull/3631

src/grub2/grub-static-greenboot.cfg
```diff
@@ -0,0 +1,19 @@
+# greenboot support, aka boot counter and boot success reporting
+insmod increment
+# Check if boot_counter exists and boot_success=0 to activate this behaviour.
+if [ -n "${boot_counter}" -a "${boot_success}" = "0" ]; then
+  # if countdown has ended, choose to boot rollback deployment,
+  # i.e. default=1 on OSTree-based systems.
+  if  [ "${boot_counter}" = "0" -o "${boot_counter}" = "-1" ]; then
```
say-paul, line 7: why forcing 0 to be -1 when it can happen with `decrement boot_counter` automatically , I have seen similar things [here](https://github.com/osbuild/osbuild/blob/3cc733d7cd3d2e65d76e227d56ef8a47ce32f469/stages/org.osbuild.grub2#L299) , couldn't understand the logic behind this.

cgwalters: I have no idea either, I'm just copying it from the osbuild code.

But actually the more I think about this, the more it feels like the right move actually is to keep these grub drop-ins in `ignition.rpm` and `greenboot.rpm` respectively in say `/usr/lib/bootupd/grub2-static` or so.  WDYT?

src/grub2/grub-static.cfg
```diff
@@ -59,9 +59,18 @@ function load_video {
...
-if [ -f $prefix/platform.cfg ]; then
-  source $prefix/platform.cfg
+if [ -f $prefix/ignition.cfg ]; then
+  source $prefix/ignition.cfg
+fi
+if [ -f $prefix/greenboot.cfg ]; then
+  source $prefix/greenboot.cfg
+fi
+# And two arbitrary other files
+if [ -f $prefix/platform01.cfg ]; then
```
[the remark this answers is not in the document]
cgwalters: To support additional drop-ins that other projects besides ignition/greenboot may want.

That said I think it'll be cleaner to make it fully generic per below...doing so

src/grub2/grub-static-greenboot.cfg
say-paul: having it in rpm makes sense, though I am not sure how to set that up in [grub.cfg](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/ch-working_with_the_grub_2_boot_loader#doc-wrapper) , which image builder writes it directly in the [config](https://github.com/osbuild/osbuild/blob/3cc733d7cd3d2e65d76e227d56ef8a47ce32f469/stages/org.osbuild.grub2#L591C18-L591C22)

cgwalters:
xref https://github.com/coreos/ignition/pull/1728

travier:
Not tested but LGTM.

I would recommend forcing a number-prefixed format for the `XY-foo.cfg` drop-in so that we have a strong ordering guarantee for the final config.

Makefile
```diff
@@ -46,4 +46,5 @@ install: install-units
 	ln -s ../bootupd.socket "${DESTDIR}$(PREFIX)/lib/systemd/system/multi-user.target.wants"
 
 install-grub-static:
-	install -D -t ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static src/grub2/*.cfg
\ No newline at end of file
+	install -D -t ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static src/grub2/*.cfg
```
travier, line 49:
```suggestion
	install -m 644 D -t ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static src/grub2/*.cfg
```

cgwalters:
> I would recommend forcing a number-prefixed format for the XY-foo.cfg drop-ins so that we have a strong ordering guarantee for the final config.

Good catch; I've added sorting.  A bit less certain about actually forcing a numeric prefix; just saying they're sorted seems OK for now.

src/grubconfigs.rs
```diff
@@ -0,0 +1,110 @@
...
+        let name = ent.file_name();
+        let name = name
+            .to_str()
+            .ok_or_else(|| anyhow!("Invalid UTF-8: {name:?}"))?;
+        if !name.ends_with(".cfg") {
+            log::debug!("Ignoring {name}");
+            continue;
+        }
+        // SAFETY: Cannot fail
+        writeln!(config, "source $prefix/{name}").unwrap();
```
say-paul, line 54: does this imply it cannot fail thats why no protection against unwrap() or is it should not fail at all, that might require some handling around Err().

cgwalters: Yes, we're just writing to a `String` which can never fail (except for OOM where we abort).

But since here it's equally ergonomic to use `?` I just changed to do that.

src/grub2/grub-static-greenboot.cfg
say-paul: Seems fair.
May not be a appropriate question, still I am going to ask to clear my doubts: Though I dont see many user doing this , but  using rpm installer  will require additional  `bootupctl backend install` to be run again to update the bootloader configs?

say-paul (approved):
Apart from the general question,  LGTM.

jmarrero (approved):
lgtm

cgwalters:
xref https://github.com/fedora-iot/greenboot/pull/129

src/grub2/grub-static-greenboot.cfg
cgwalters: We don't directly support updating the configs yet.

(In practice of course, it will trivially work to just `cp` them manually)

PR #546, opened 2023-10-16
efi: Rework mount order

cgwalters:
This ensures that we'll use `/boot/efi` first, which is what is more commonly expected in our current ecosystem.


cgwalters:
Rolled this in to https://github.com/coreos/bootupd/pull/543

PR #549, opened 2023-10-24
packaging: Move static configs into main package

cgwalters:
No reason to have them be separate, that's what I ended up doing in Fedora.

(We should probably delete this spec file...)


jmarrero (approved):
lgtm

PR #555, opened 2023-10-31
Add `backend install --auto`

cgwalters:
This drains the code from
https://github.com/containers/bootc/pull/155/commits/20aa19ce6da58078999368e4f6dd948b2e3d990b

Because it will also be what Anaconda wants to do by default.


jmarrero (approved):
lgtm

Issue #550, opened 2023-10-24
add s390x support


cgwalters:
Right now we're just missing s390x, which adds messy architecture conditionals to our code.

The way zipl works is somewhat UKI-like as far as I understand things in that it needs to be run for each kernel update, and it merges together the kernel/initramfs.  The ostree/zipl glue is basically just running `zipl` after an update to the BLS configs.

So I think the bootupd support for s390x is basically just "detect and no-op" effectively - almost.

Except for one detail: on the first install to a disk, there's a [magic command](https://github.com/rhinstaller/anaconda/blob/1480d230152b86489754286436bdc24e84a01403/pyanaconda/modules/storage/bootloader/zipl.py#L153) that needs to be run called "chreipl".

Also xref https://github.com/coreos/coreos-installer/blob/c4ed04156ae21029e84acc3cdf51493dc1fa8b8b/src/s390x/README.md?plain=1#L11

PR #556, opened 2023-11-02
No-op on s390x

cgwalters:
Closes: https://github.com/coreos/bootupd/issues/550

I believe we don't have anything to do here; zipl always merges kernel/initramfs/bootloader and hence s390x updates are owned by ostree effectively.

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -61,7 +61,7 @@ pub(crate) fn install(
         all_components.values().collect()
     };
 
-    if target_components.is_empty() {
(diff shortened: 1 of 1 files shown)


cgwalters:
cc @nikita-dubrovskii

jmarrero (approved):
lgtm

nikita-dubrovskii:
Yes, `ostree` with its `zipl` wraper handles `BLS` either automatically, or with extra logic in SecEx case

PR #571, opened 2023-11-20
efi: Gracefully no-op if there's no `BootCurrent`

cgwalters:
This happens in a default virt-manager run for example.


jmarrero (approved):
/lgtm

PR #583, opened 2023-12-11
grubconfigs: move the GRUB2DIR creation earlier

dustymabe:
We add the GRUB2DIR creation in ece9120 but there is an earlier use of that directory if there are dropins so let's move the creation of the directory earlier before the dropins are processed.

I hit this because I'm testing out things on FCOS. Building a qemu and metal image works just fine, because in that case we are doing the MBR install using `grub2-install` and I guess that tool creates the directory. In the metal4k case, though, there is no BIOS install and thus we get here and the GRUB2DIR doesn't exist and bootupd fails.

### DIFF
--- a/src/grubconfigs.rs
+++ b/src/grubconfigs.rs
@@ -38,6 +38,10 @@ pub(crate) fn install(target_root: &openat::Dir, efi: bool, write_uuid: bool) ->
         root_dev != boot_dev
     };
 
+    if !bootdir.exists(GRUB2DIR)? {
+        bootdir.create_dir(GRUB2DIR, 0o700)?;
+    }
+
     let mut config = std::fs::read_to_string(Path::new(CONFIGDIR).join("grub-static-pre.cfg"))?;
 
     let dropindir = openat::Dir::open(&Path::new(CONFIGDIR).join(DROPINDIR))?;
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
I'm ok with this as is but shouldn't we delete the later creation then?

dustymabe:
hmm. really odd.. my local `git diff HEAD^` shows me:


```patch
$ git diff HEAD^
diff --git a/src/grubconfigs.rs b/src/grubconfigs.rs
index 8b4b0c9..dbf4869 100644
--- a/src/grubconfigs.rs
+++ b/src/grubconfigs.rs
@@ -38,6 +38,10 @@ pub(crate) fn install(target_root: &openat::Dir, efi: bool, write_uuid: bool) ->
         root_dev != boot_dev
     };
 
+    if !bootdir.exists(GRUB2DIR)? {
+        bootdir.create_dir(GRUB2DIR, 0o700)?;
+    }
+
     let mut config = std::fs::read_to_string(Path::new(CONFIGDIR).join("grub-static-pre.cfg"))?;
 
     let dropindir = openat::Dir::open(&Path::new(CONFIGDIR).join(DROPINDIR))?;
@@ -68,10 +72,6 @@ pub(crate) fn install(target_root: &openat::Dir, efi: bool, write_uuid: bool) ->
         config.push_str(post.as_str());
     }
 
-    if !bootdir.exists(GRUB2DIR)? {
-        bootdir.create_dir(GRUB2DIR, 0o700)?;
-    }
-
     bootdir
         .write_file_contents(format!("{GRUB2DIR}/grub.cfg"), 0o644, config.as_bytes())
         .context("Copying grub-static.cfg")?;
```

dustymabe:
oh I see. the second half of the change didn't git added to the commit when I did `git add --patch`.. Fixed now!

cgwalters (approved):

PR #585, opened 2023-12-16
coreos: update aleph version handling

dustymabe:
In https://github.com/osbuild/osbuild/pull/1475 we are updating
aleph version to have more information and some of the fields
are changing slightly. Notably here `imgid` is no longer going to
be populated and `build` -> `version` now. Let's make these tweaks
and also just drop any fields from the definition here that we
aren't using to tolerate changes better.

### DIFF
--- a/src/coreos.rs
+++ b/src/coreos.rs
@@ -8,9 +8,8 @@
 
 use anyhow::{Context, Result};
 use chrono::prelude::*;
-use openat_ext::OpenatDirExt;
 use serde::{Deserialize, Serialize};
-use std::fs::{canonicalize, symlink_metadata};
+use std::fs::File;
 use std::path::Path;
 
 #[derive(Serialize, Deserialize, Clone, Debug, Hash, Ord, PartialOrd, PartialEq, Eq)]
@@ -28,41 +27,70 @@ pub(crate) struct AlephWithTimestamp {
 }
 
 /// Path to the file, see above
-const ALEPH_PATH: &str = "/sysroot/.coreos-aleph-version.json";
+const ALEPH_PATH: &str = "sysroot/.coreos-aleph-version.json";
 
-pub(crate) fn get_aleph_version() -> Result<Option<AlephWithTimestamp>> {
-    let sysroot = openat::Dir::open("/")?;
-    let mut path = ALEPH_PATH;
-    if !Path::new(ALEPH_PATH).exists() {
+pub(crate) fn get_aleph_version(root: &Path) -> Result<Option<AlephWithTimestamp>> {
+    let path = &root.join(ALEPH_PATH);
+    if !path.exists() {
         return Ok(None);
     }
-    let target;
-    let is_link = symlink_metadata(path)
-        .with_context(|| format!("reading metadata for {}", path))?
-        .file_type()
-        .is_symlink();
-    if is_link {
-        target =
-            canonicalize(path).with_context(|| format!("getting absolute path to {}", path))?;
-        path = target.to_str().unwrap()
-    }
-    if let Some(statusf) = sysroot.open_file_optional(path)? {
-        let meta = statusf.metadata()?;
-        let bufr = std::io::BufReader::new(statusf);
-        let aleph: Aleph = serde_json::from_reader(bufr)?;
-        Ok(Some(AlephWithTimestamp {
-            aleph,
-            ts: meta.created()?.into(),
-        }))
-    } else {
-        Ok(None)
-    }
+    let statusf = File::open(path).with_context(|| format!("Opening {path:?}"))?;
+    let meta = statusf.metadata()?;
+    let bufr = std::io::BufReader::new(statusf);
+    let aleph: Aleph = serde_json::from_reader(bufr)?;
+    Ok(Some(AlephWithTimestamp {
+        aleph,
+        ts: meta.created()?.into(),
+    }))
 }
 
 #[cfg(test)]
 mod test {
     use super::*;
     use anyhow::Result;
+
+    const V1_ALEPH_DATA: &str = r##"
+    {
+        "version": "32.20201002.dev.2",
+        "ref": "fedora/x86_64/coreos/testing-devel",
+        "ostree-commit": "b2ea6159d6274e1bbbb49aa0ef093eda5d53a75c8a793dbe184f760ed64dc862"
+    }"##;
+
+    #[test]
+    fn test_parse_from_root_empty() -> Result<()> {
+        // Verify we're a no-op in an empty root
+        let root: &tempfile::TempDir = &tempfile::tempdir()?;
+        let root = root.path();
+        assert!(get_aleph_version(root).unwrap().is_none());
+        Ok(())
+    }
+
+    #[test]
+    fn test_parse_from_root() -> Result<()> {
+        let root: &tempfile::TempDir = &tempfile::tempdir()?;
+        let root = root.path();
+        let sysroot = &root.join("sysroot");
+        std::fs::create_dir(sysroot).context("Creating sysroot")?;
+        std::fs::write(root.join(ALEPH_PATH), V1_ALEPH_DATA).context("Writing aleph")?;
+        let aleph = get_aleph_version(root).unwrap().unwrap();
+        assert_eq!(aleph.aleph.version, "32.20201002.dev.2");
+        Ok(())
+    }
+
+    #[test]
+    fn test_parse_from_root_linked() -> Result<()> {
+        let root: &tempfile::TempDir = &tempfile::tempdir()?;
+        let root = root.path();
+        let sysroot = &root.join("sysroot");
+        std::fs::create_dir(sysroot).context("Creating sysroot")?;
+        let target_name = ".new-ostree-aleph.json";
+        let target = &sysroot.join(target_name);
+        std::fs::write(root.join(target), V1_ALEPH_DATA).context("Writing aleph")?;
+        std::os::unix::fs::symlink(target_name, root.join(ALEPH_PATH)).context("Symlinking")?;
+        let aleph = get_aleph_version(root).unwrap().unwrap();
+        assert_eq!(aleph.aleph.version, "32.20201002.dev.2");
+        Ok(())
+    }
 
     #[test]
     fn test_parse_old_aleph() -> Result<()> {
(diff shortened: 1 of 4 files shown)


dustymabe:
I'm sure someone that knows rust better than I do can show me a much more elegant way to implement the 2nd commit here. What I have right now is kind of ugly IMO.

cgwalters:
OK so, the only (non-informational) use of the aleph data is for adoption on very old coreos systems.  Think about a case of updating from e.g. RHCOS 4.3 (RHEL 8.x) all the way to a modern version.  We still need to parse the *old* aleph file for that; we can't just hard switch.

Probably simplest is to change every field to `Option<>`.

Or...honestly (and this relates to a larger topic of bootupd versioning) we could just extract the btime of the aleph file...or failing that, hmm we could probably rely on the btime of `/boot` or `/sysroot/ostree`.  Basically our use of the aleph file here is just to have a "pretty" version but...who cares, we can just use timestamp-as-version.

cgwalters (changes requested):
Per above

dustymabe:
> OK so, the only (non-informational) use of the aleph data is for adoption on very old coreos systems. Think about a case of updating from e.g. RHCOS 4.3 (RHEL 8.x) all the way to a modern version. We still need to parse the old aleph file for that; we can't just hard switch.

I agree that we can't hard switch. I thought the changes I was doing here were backwards compatible so thank you for pointing out that they aren't! I thought by making it:

```
pub(crate) struct Aleph {      
    #[serde(alias = "build")]  
    pub(crate) version: String,
}                              
```

Would mean that if `build:` was in the file then it would pick up that as version, but if version was in there (the new files) then it would use that. Am I missing something?

 
> Probably simplest is to change every field to Option<>.

I removed all the other values from the `struct Aleph` because we don't really use them here in `bootupd` that I can tell. Do you think that will cause errors? I tested running `bootupctl status` on current FCOS, 4.15 RHCOS, and also OSBuild built FCOS via deploy-via-container and ostree deployed and it didn't seem to complain that there are fields it didn't know about.


> 
> Or...honestly (and this relates to a larger topic of bootupd versioning) we could just extract the btime of the aleph file...or failing that, hmm we could probably rely on the btime of /boot or /sysroot/ostree. Basically our use of the aleph file here is just to have a "pretty" version but...who cares, we can just use timestamp-as-version.

I kind of like having the version information, but I guess we could cut it out. Either way I'm mostly trying to not break things when we switch to OSBuild.

dustymabe:
I realized we were not failing if the file didn't exist before and my new code was failing if it didn't exist. Updated to not fail if no file exists.

I feel like the second commit is still really ugly and there is probably a much more elegant way to say "read this file, or the file this symlink points to if it is a symlink".

cgwalters:
> Would mean that if build: was in the file then it would pick up that as version, but if version was in there (the new files) then it would use that. Am I missing something?

Ah sorry...yes, I think that will work, however what would increase my confidence here a lot is to leave the existing unit test unchanged to prove we continue to parse the old format, and add a new unit test instead.

cgwalters:
> Let's canonicalize and find the actual file before dropping into
sysroot.open_file_optional() because the underlying code for that uses
O_NOFOLLOW and will give an ELOOP (too many symbolic links) error.

I think that's a deficiency of the `openat` crate, the `cap-std` library (which we want to port to) shouldn't behave that way.  I'm good with just a TODO here like:
`// TODO: When porting to cap-std we should be able to just automatically follow the link here`

src/coreos.rs
```diff
@@ -33,7 +32,21 @@ const ALEPH_PATH: &str = "/sysroot/.coreos-aleph-version.json";
 
 pub(crate) fn get_aleph_version() -> Result<Option<AlephWithTimestamp>> {
     let sysroot = openat::Dir::open("/")?;
```
cgwalters, line 34: Oh actually since we always operate on `/` here we can just use the `std` APIs too.  I can push a change to do that if you like.

dustymabe:
> > Would mean that if build: was in the file then it would pick up that as version, but if version was in there (the new files) then it would use that. Am I missing something?
> 
> Ah sorry...yes, I think that will work, however what would increase my confidence here a lot is to leave the existing unit test unchanged to prove we continue to parse the old format, and add a new unit test instead.

Good idea! Done!

src/coreos.rs
cgwalters: Done

cgwalters (approved):

dustymabe:
Changes LGTM.

I'm trying to figure out the CI failure now.

cgwalters:
Ah yes, "no components are adoptable" means we messed up something with the aleph detection for "this is a coreos system".

cgwalters:
Ah I see the error is
> [2023-12-18T15:42:12.526Z] # error: duplicate field `version` at line 3 column 11


And that's because the old aleph data has both `build`  *and* `imgid`:

```
{
	"build": "44.81.202004260825-0",
	"ref": "2062bce64e4932160feb58ce4976a885172d3f1017dc01f09177504bd55e035b",
	"ostree-commit": "2062bce64e4932160feb58ce4976a885172d3f1017dc01f09177504bd55e035b",
	"imgid": "rhcos-44.81.202004260825-0-qemu.x86_64.qcow2"
}
```

cgwalters:
For my and others' future reference, a handy guestfish interaction session to get this data from an old qcow2:

```
$ guestfish --ro -a rhcos-4.4.3-x86_64-qemu.x86_64.qcow2
><fs> run
><fs> debug sh "blockdev --getsize /dev/sda4"
32503775
><fs> debug sh "echo 0 32471007 linear /dev/sda4 32768 | dmsetup create noluks-root"
><fs> mount /dev/mapper/noluks-root /
><fs> cat /.coreos-aleph-version.json 
{
	"build": "44.81.202004260825-0",
	"ref": "2062bce64e4932160feb58ce4976a885172d3f1017dc01f09177504bd55e035b",
	"ostree-commit": "2062bce64e4932160feb58ce4976a885172d3f1017dc01f09177504bd55e035b",
	"imgid": "rhcos-44.81.202004260825-0-qemu.x86_64.qcow2"
}

><fs> 

```

dustymabe:
> Ah I see the error is
> 
> > [2023-12-18T15:42:12.526Z] # error: duplicate field `version` at line 3 column 11
> 
> And that's because the old aleph data has both `build` _and_ `imgid`:
> 
> ```
> {
> 	"build": "44.81.202004260825-0",
> 	"ref": "2062bce64e4932160feb58ce4976a885172d3f1017dc01f09177504bd55e035b",
> 	"ostree-commit": "2062bce64e4932160feb58ce4976a885172d3f1017dc01f09177504bd55e035b",
> 	"imgid": "rhcos-44.81.202004260825-0-qemu.x86_64.qcow2"
> }
> ```

How is that different from our current `test_parse_old_aleph()` test?

dustymabe:
IIUC the error is happening when `bootupctl status` is run:

https://github.com/coreos/bootupd/blob/055de7d72ec6cd49a060a4ed01f97d56a9bdc669/tests/e2e-update/e2e-update-in-vm.sh#L54

dustymabe:
interestingly enough I just ran the test locally and it passed:

```
sudo env COSA_DIR=/srv ./tests/e2e-update/e2e-update.sh
...
...
+ echo 'ok bootupd e2e'
ok bootupd e2e
```

??

dustymabe:
ok but somehow my local test was using the old code:

```
Starting /run/bootupd-source/tests/e2e-update//e2e-update-in-vm.sh
+ setenforce 0
+ runv rpm-ostree rebase /run/cosadir/tmp/repo:596371cd2387cd70ef593570e72359eb263d1f2c307ac83d0a3d6a807afc7db6
+ set -x
+ rpm-ostree rebase /run/cosadir/tmp/repo:596371cd2387cd70ef593570e72359eb263d1f2c307ac83d0a3d6a807afc7db6
53 metadata, 29 content objects imported; 125.6 MB content written
Staging deployment...done
Upgraded:
  grub2-common 1:2.06-89.fc38 -> 1:2.06-102.fc38
  grub2-efi-x64 1:2.06-89.fc38 -> 1:2.06-102.fc38
  grub2-pc 1:2.06-89.fc38 -> 1:2.06-102.fc38
  grub2-pc-modules 1:2.06-89.fc38 -> 1:2.06-102.fc38
  grub2-tools 1:2.06-89.fc38 -> 1:2.06-102.fc38
  grub2-tools-minimal 1:2.06-89.fc38 -> 1:2.06-102.fc38
Added:
  test-bootupd-payload-1.0-1.x86_64
Changes queued for next boot. Run "systemctl reboot" to start a reboot
+ runv touch /etc/e2e-update-in-vm.sh.upgraded
+ set -x
+ touch /etc/e2e-update-in-vm.sh.upgraded
+ runv systemd-run -- systemctl reboot
+ set -x
+ systemd-run -- systemctl reboot
Running as unit: run-ra21eb479578c4b15bd175e6fdfc57430.service
+ touch /run/rebooting
+ sleep infinity
Starting /run/bootupd-source/tests/e2e-update//e2e-update-in-vm.sh
Validated: EFI
ok validate
{
	"build": "39.20231218.dev.0",
	"version": "39.20231218.dev.0",
	"ref": "fedora/x86_64/coreos/testing-devel",
	"ostree-commit": "1c1ba117d54a1b4e3276beb9fea11c56d6266ed88771771f3435ac5826262468",
	"imgid": "fedora-coreos-39.20231218.dev.0-qemu.x86_64.qcow2"
}
Component EFI
  Installed: grub2-efi-x64-1:2.06-89.fc38.x86_64,shim-x64-15.6-2.x86_64
  Update: Available: grub2-efi-x64-1:2.06-102.fc38.x86_64,shim-x64-15.6-2.x86_64,test-bootupd-payload-1.0-1.x86_64
No components are adoptable.
CoreOS aleph image ID: fedora-coreos-39.20231218.dev.0-qemu.x86_64.qcow2
Boot method: EFI
ok update avail
Updated EFI: grub2-efi-x64-1:2.06-102.fc38.x86_64,shim-x64-15.6-2.x86_64,test-bootupd-payload-1.0-1.x86_64
ok update not avail
Completing TAP test with:
1..3
Powering off.

```

dustymabe:
oh actually I think I might know the issue here:

https://github.com/coreos/coreos-assembler/commit/c2d37f4005e0e7c98d24c5f561c149ab7fd4dfce

I was trying to be helpful with that but I think it caused a problem.

cgwalters:
Ah.  Yes, that commit created an aleph format which is neither the "v1" nor "v2"; probably should just revert.

dustymabe:
> oh actually I think I might know the issue here:
> 
> [coreos/coreos-assembler@c2d37f4](https://github.com/coreos/coreos-assembler/commit/c2d37f4005e0e7c98d24c5f561c149ab7fd4dfce)
> 
> I was trying to be helpful with that but I think it caused a problem.

An easy solution to the problem is to revert that commit and scratch the builds that we started today since we haven't released them yet.

PR #586, opened 2023-12-16
systemd: Drop unit files

EyeCantCU:
This removes bootupd's systemd service and socket and removes their usage


cgwalters (changes requested):
Hi, thanks for starting work on this!

This problem is a bit harder than this, because the systemd services are used today.  To fix this we need to unwind the "client-server" architecture that we have right now.

So this is missing this bit from the issue description:

>  where we want to run things under systemd, we do it dynamically via systemd-run e.g.

EyeCantCU:
> Hi, thanks for starting work on this!
> 
> This problem is a bit harder than this, because the systemd services are used today. To fix this we need to unwind the "client-server" architecture that we have right now.
> 
> So this is missing this bit from the issue description:
> 
> > where we want to run things under systemd, we do it dynamically via systemd-run e.g.

Thank you for the explanation! Ah, that makes sense. I had thought there was more to it than purging the service and socket, but wasn't entirely sure. I can certainly look more into what that would look like.

PR #608, opened 2024-01-31
ci: Add a workflow which does an install via bootc

cgwalters:
This helps effectively test e2e things like `--with-static-configs`.


jmarrero (approved):
lgtm

PR #609, opened 2024-01-31
model: Add tracking of our own static config versions

cgwalters:
Currently our static configs don't directly support updates. (They really should)

We also have a use case around simply introspecting the state that static configs were enabled (xref https://github.com/ostreedev/ostree/pull/3150) where we want to have ostree not run grub2-mkconfig in this case.

In preparation for both of these things, add tracking of our *own* version of the static configs into the metadata.

In theory of course, static configs could include other components (ignition, greenboot, etc.) and we should track those too.  For now this at least gets us basic metadata.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,3 +46,5 @@ jobs:
             -v /var/lib/containers:/var/lib/containers \
             localhost/bootupd:latest bootc install to-filesystem --skip-fetch-check \
             --disable-selinux --replace=alongside /target
+          # Verify we injected static configs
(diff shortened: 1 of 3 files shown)


jlebon:
This broke FCOS stable builds because we currently incorrectly use the bootupd from cosa instead of the one from the target system to `bootupctl backend install`. v0.2.18 will write this field, even if it's `null`, but v0.2.17 will hard error trying to deserialize it back.

(Obviously the fix here is on the cosa side, but just noting for posterity and for something to link to.)

PR #619, opened 2024-02-19
grub2: source in a console.cfg file if exists

dustymabe:
This will allow users or distro builders place console settings here that will get picked up on boot. This was discussed as part of https://github.com/coreos/fedora-coreos-tracker/issues/1671


dustymabe:
Creating as draft because I still need to test this in concert with other changes

jlebon (approved):

jmarrero (approved):
lgtm

Issue #630, opened 2024-03-15
handle efi vendordir better


cgwalters:
See https://github.com/rhinstaller/anaconda/pull/5508#issuecomment-2000443423 and previous

Basically during an installation process we should really fix things so that we find the EFI vendor for the shim/grub we're going to install (from `/usr/lib/bootupd/updates/EFI`), verify its presence in the real ESP, and then target that.

jkonecny12:
I see this is closed. Does that mean that the bootupd installed by Anaconda will work on dual boot or more adjustments are required?

HuijingHei:
> Does that mean that the bootupd installed by Anaconda will work on dual boot or more adjustments are required?

Not quite sure, this issue is only to make sure to find the `vendor` directory (and shim path) and will not break creating the EFI boot manager entry. Maybe @cgwalters has more context about your question.

travier:
https://github.com/coreos/bootupd/pull/633 helped but is not complete. There are cases where could have multiple shim binaries in the EFI partition and we only want to touch our own, not others.

HuijingHei:
> #633 helped but is not complete. There are cases where could have multiple shim binaries in the EFI partition and we only want to touch our own, not others.
> 
> The logic should be:
> 
> * Look at what we have in the image
> * Find the corresponding in the EFI partition
> * Update only that

For example, firstly find the `<vendor>/shimx64.efi` file under `/usr/lib/bootupd/updates/EFI/`, then find it under EFI partition, if they are the same, then get the shim path; if not, will fail. Is this correct?

travier:
I've edited my comments to try to clarify it

HuijingHei:
Thanks @travier for the clarification!

PR #633, opened 2024-03-26
grubconfig: handle efi vendordir better

HuijingHei:
Fix https://github.com/coreos/bootupd/issues/630

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -25,7 +25,7 @@ pub(crate) const ESP_MOUNTS: &[&str] = &["boot/efi", "efi", "boot"];
 
 /// The binary to change EFI boot ordering
 const EFIBOOTMGR: &str = "efibootmgr";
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
This looks sane to me, thanks!

(Longer term we need to stop hardcoding shim...that's it's own issue though)

cgwalters:
Just one unused import to fix from CI

cgwalters:
Hmm, this may have actually broken things, bootc's CI runs with the -dev image that tracks git main started falling over, see

https://github.com/containers/bootc/actions/runs/8443222436/job/23126361204?pr=447

```
error: boot data installation failed: installing component EFI: Updating EFI firmware variables: Locating EFI vendordir: Failed to find EFI vendor dir
```

HuijingHei:
Sorry, `aarch64` does not have `shimx64.efi`, seems we should use `shim.efi` instead.

PR #634, opened 2024-03-27
aa64: Minor fix for efi vendordir

HuijingHei:
Minor fix for https://github.com/coreos/bootupd/pull/633

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -25,6 +25,10 @@ pub(crate) const ESP_MOUNTS: &[&str] = &["boot/efi", "efi", "boot"];
 
 /// The binary to change EFI boot ordering
 const EFIBOOTMGR: &str = "efibootmgr";
+#[cfg(target_arch = "aarch64")]
+pub(crate) const SHIM: &str = "shimaa64.efi";
+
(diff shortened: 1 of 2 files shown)


src/efi.rs
```diff
@@ -484,7 +488,7 @@ pub(crate) fn set_efi_current(device: &str, espdir: &openat::Dir, vendordir: &st
     if espdir.exists(&shim)? {
         anyhow::bail!("Failed to find {SHIM}");
     }
-    let loader = format!("\\EFI\\{}\\shimx64.efi", vendordir);
+    let loader = format!("\\EFI\\{}\\{SHIM}", vendordir);
```
HuijingHei, line 491: @cgwalters update this to sync with `SHIM`, or might have issue on aarch64 as it does not have `shimx64.efi` when called by `--update-firmware`.

cgwalters: Yes, this looks right.  I suspect the only reason it may have appeared to work on aarch64 is because we ended up using the fallback path.

cgwalters (approved):

HuijingHei:
Thanks @cgwalters for the review!

Issue #658, opened 2024-05-15
`bootupctl backend install --update-firmware` should match current Anaconda logic


travier:
See discussions in:
- https://bugzilla.redhat.com/show_bug.cgi?id=2268505
- https://github.com/rhinstaller/anaconda/pull/5508
- https://pagure.io/workstation-ostree-config/pull-request/453#comment-199664
- https://gitlab.com/fedora/ostree/sig/-/issues/1

HuijingHei:
What I understood is, according to https://github.com/rhinstaller/anaconda/pull/5508#issuecomment-1984219257

The current issue in bootupd:

- with `--update-firmware`, it runs `efibootmgr` with `--create` (which is the same as -c, it creates an entry and adds it to the boot order)
- without `--update-firmware`, it does not run efibootmgr at all.

There's no way to make it create an entry but not add it to the boot order.

**Change to:**

- keep with `--update-firmware` as it is
- without `--update-firmware`, run `efibootmgr -C` (`-C` creates an entry but does not add it to the boot order)

Or

- make `--update-firmware` as default to run `efibootmgr -c` to create an entry and add it to boot order
- if there is explicit `--update-firmware false`, run `efibootmgr -C` to create an entry but not add it to the boot order

travier:
We want to be able to use `bootupd backend install --update-firmware` in Anaconda to setup the boot order on EFI systems. This is https://github.com/rhinstaller/anaconda/pull/5508 and ideally we'll re-open and merge this PR once the issue is fixed here.

The issue is that when called with `--update-firmware`, bootupd currently removes the `BootCurrent` boot entry (https://github.com/coreos/bootupd/blob/19610be787cf2d027e3f3dccb556bba9c2e4b389/src/efi.rs#L459) and then adds a new `BootEntry` for the system being installed (https://github.com/coreos/bootupd/blob/19610be787cf2d027e3f3dccb556bba9c2e4b389/src/efi.rs#L497).

The current Anaconda behavior (and the one we want to reproduce in bootupd) is to only create a new boot entry and set it as the first one in the boot order (not deleting the `BootCurrent` boot entry).

Anaconda also creates the `BootEntry` with the name from the `EFI/fedora/BOOTX64.CSV` file in the ESP (Fedora for Fedora instead of `fedora` for bootupd).

[...]

PR #665, opened 2024-06-05
efi: change `--update-firmware` to match current Anaconda logic

HuijingHei:
Copy Timothée's comment:
We want to be able to use `bootupd backend install --update-firmware`
 in Anaconda to setup the boot order on EFI systems.

The issue is that when called with `--update-firmware`, bootupd
currently removes the `BootCurrent` boot entry, and then adds a
new BootEntry for the system being installed.

The current Anaconda behavior is to remove all boot entries that
match the product name, then create a new boot entry using the
product name and set it as the first one in the boot order.

To sync with Anaconda behavior, when called with `--update-firmware`,
bootupd will remove all boot entries that match the product name, 
using the same Anaconda definition 
`ANACONDA_PRODUCTNAME=$(sed -r -e 's/ *release.*//' /etc/system-release)`
from https://github.com/rhinstaller/anaconda/blob/f52880c8ac0f0b19d5e4b7887c80aedec1d4bfa7/data/liveinst/liveinst#L46, and also creates a new boot entry.

See https://github.com/coreos/bootupd/issues/658

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -12,6 +12,7 @@ use std::process::Command;
 use anyhow::{bail, Context, Result};
 use fn_error_context::context;
 use openat_ext::OpenatDirExt;
+use os_release::OsRelease;
 use walkdir::WalkDir;
 use widestring::U16CString;
 
@@ -137,8 +138,14 @@ impl Efi {
             log::debug!("Not booted via EFI, skipping firmware update");
             return Ok(());
         }
-        clear_efi_current()?;
-        set_efi_current(device, espdir, vendordir)
+        // Read /etc/os-release
+        let release: OsRelease = OsRelease::new()?;
+        let product_name: &str = &release.name;
+        log::debug!("Get product name: {product_name}");
+        assert!(product_name.len() > 0);
+        // clear all the boot entries that match the target name
+        clear_efi_target(product_name)?;
+        create_efi_boot_entry(device, espdir, vendordir, product_name)
     }
 }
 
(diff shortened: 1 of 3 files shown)


HuijingHei:
CI failed as the ubuntu does not have `/etc/system-release`, so can not get product name
```
---- efi::tests::test_get_product_name stdout ----
Error: Get product name

Caused by:
    0: Failed to read /etc/system-release
    1: No such file or directory (os error 2)
```

cgwalters:
> /etc/system-release

That's a deprecated path, the preferred file is `/usr/lib/os-release` (also available as `/etc/os-release`)

travier:
The initial idea was to match Anaconda behavior exactly but that's indeed not ideal.

Maybe we should use the `NAME` entry from `/etc/os-release`. I wonder if spaces are allowed as that would let us use the `NAME` entry as is.

src/efi.rs
```diff
@@ -546,3 +591,54 @@ fn find_file_recursive<P: AsRef<Path>>(dir: P, target_file: &str) -> Result<Vec<
...
     Ok(result)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_get_product_name() -> Result<()> {
+        let product_name = get_product_name()?;
```
cgwalters, line 601: We shouldn't read from the host system in a unit test.  The best thing to do here is what you did in `test_parse_boot_entries()` - the unit test is testing parsing, leaving the "load the data" part outside of the test.

That all said, I do think we should be reading from the `os-release` file anyways, and actually for that...hmm, there are [a few crates out there](https://lib.rs/search?q=os-release), not sure of maintenance status of any of them.  Maybe not worth pulling in something external.  We may even have something in our ecosystem.  Ah yep, see https://github.com/coreos/rpm-ostree/blob/fec15f5e24ce2c44f97f602152ac913134bd3173/rust/src/countme.rs#L67

src/efi.rs
```diff
@@ -455,46 +461,84 @@ fn validate_esp(dir: &openat::Dir) -> Result<()> {
...
+/// Parse boot entries from efibootmgr output
+fn parse_boot_entries(output: &str) -> Vec<BootEntry> {
+    let mut entries = Vec::new();
+
+    for line in output.lines() {
+        // Example line: Boot0002* Fedora	HD(xx)
+        if line.starts_with("Boot0") {
+            if let Some((parts, _)) = line.split_once('\t') {
+                if let Some((id, name)) = parts.split_once(' ') {
+                    let id = id[4..].trim_end_matches('*').to_string();
```
cgwalters, line 493: Minor but: The `[4..]` part could panic if the input was malformed.  WDYT about this version?

```
/// Parse boot entries from efibootmgr output
fn parse_boot_entries(output: &str) -> Vec<BootEntry> {
    let mut entries = Vec::new();

    for line in output.lines() {
        let line = if let Some(l) = line.strip_prefix("Boot") {
            l
        } else {
            continue;
        };
        let parts = if let Some((parts, _)) = line.split_once('\t') {
            parts
        } else {
            continue;
        };
        if let Some((id, name)) = parts.split_once(' ') {
            let id = id.trim_end_matches('*').to_string();
            let name = name.trim().to_string();
            entries.push(BootEntry { id, name });
        }
    }
    entries
}
```

I was testing this out with https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=638374b60de0116bdf9462c9a8099603

cgwalters (commented):
Overall this is looking pretty good, but we definitely could use some manual testing.  I will try to do that tomorrow.

How much testing did you do with this on virtual or ideally some physical hardware?

HuijingHei:
Thanks @cgwalters @travier very much for the suggestion, it sounds more reasonable to use `NAME` in `/etc/os-release`, it looks good for rhel (`NAME="Red Hat Enterprise Linux"`) and centos (`NAME="CentOS Stream"`), but for fedora, get `NAME="Fedora Linux"`, this will not align with Anaconda (using `Fedora`), should we continue with `"Fedora Linux"`?

HuijingHei:
> Overall this is looking pretty good, but we definitely could use some manual testing. I will try to do that tomorrow.
> 
> How much testing did you do with this on virtual or ideally some physical hardware?

I did testing using `bootc install to-filesystem`, as this will trigger creating new boot entry

The steps what I did:
```
# start fedora vm with uefi (remove cloud-init and set root passwd first)
virt-customize --root-password password:xxx --uninstall cloud-init -a fcloud.qcow2
virt-install --boot uefi --import --name rhel9 --ram 4096 --disk path=/home/fedora/data/fcloud.qcow2 --os-variant rhel9-unknown --network default --nographics

# add more efibootmgr list and check
efibootmgr -c --disk /dev/vda --part 2 --loader \\EFI\\fedora\\shimx64.efi --label fedora
efibootmgr -c --disk /dev/vda --part 2 --loader \\EFI\\fedora\\shimx64.efi --label FEDORA

# build bootupd
dnf install -y podman; dnf clean all
podman run -it --name fcos-build --net=host -v ./:/srv -w /srv --security-opt=label=disable --privileged quay.io/coreos-assembler/fcos-buildroot:testing-devel bash

git clone -b sync-anaconda https://github.com/HuijingHei/bootupd.git
cd bootupd && make
podman rmi quay.io/coreos-assembler/fcos-buildroot:testing-devel

# build new bootc container to include the bin bootupd file
cat <<EOF > Containerfile
FROM quay.io/centos-bootc/fedora-bootc:eln
RUN rm -f /usr/lib/systemd/system/bootupd.service && \
    rm -f /usr/lib/systemd/system/bootupd.socket && \
    rm -f /usr/libexec/bootupd && \
    rm -f /usr/bin/bootupctl
COPY bootupd/target/release/bootupd /usr/libexec/bootupd
RUN ln -f /usr/libexec/bootupd /usr/bin/bootupctl
EOF

podman build -t test .

# run and check the log
podman run --rm --privileged -v /var/lib/containers:/var/lib/containers -v /:/target --pid=host --security-opt label=disable localhost/test:latest env BOOTC_BOOTLOADER_DEBUG=1 bootc install to-filesystem --skip-fetch-check --replace=alongside /target
```

src/efi.rs
HuijingHei: Good suggestion, the `filter_map` looks more rust and pretty, thank you very much!

HuijingHei:
Hold on, after running bootc `podman run --rm --privileged -v /var/lib/containers:/var/lib/containers -v /:/target --pid=host --security-opt label=disable localhost/test:latest env BOOTC_BOOTLOADER_DEBUG=1 bootc install to-filesystem --skip-fetch-check --replace=alongside /target`, vm failed to start after reboot, need to do more research.

HuijingHei:
Start vm with secure boot disabled `--boot firmware=efi,firmware.feature0.enabled=no,firmware.feature0.name=secure-boot`, seems `/boot/loader/entries/ostree-1.conf` missing console parameter, add `ignition.platform.id=qemu console=tty0 console=ttyS0,115200n8` after bootc before reboot, then reboot, hang at 
```
[FAILED] Failed to start systemd-vconsole-s…up.service - Virtual Console Setup.
See 'systemctl status systemd-vconsole-setup.service' for details.

[    **] Job dev-disk-by\x2duuid-8424dfd0\x2…start running (8min 3s / no limit)
```
Find `/dev/vda` is only `5 GiB`, maybe it is not big enough, will enlarge the disk and try again.

travier:
> should we continue with `"Fedora Linux"`?

Yes, we'll slightly differ from what Anaconda does but that's fine.

> How much testing did you do with this on virtual or ideally some physical hardware?

It will indeed need a lot of manual testing on real hardware. We should likely only push this code to Fedora Rawhide/41 so that we get some time to test it there. I can also make test containers/ISOs.

cgwalters (approved):

cgwalters:
> Boot0003* CentOS Linux
> Boot0004* CentOS Stream

OK yeah though we can see the difference here in using the os-release data vs system-release.  I guess a consequence of this is that we'll "leak" the old bootentry in the use case of a bootc install on top of a system installed from Anaconda.  Which...will actually be a somewhat common case.  I'm not an expert in this but I think at least some uefi firmware will GC entries at some point that don't have backing files?

src/efi.rs
cgwalters: To be clear your previous code was great too!  This was a very minor nit.

cgwalters:
OK so logistically we have a bit of a trap because:

- https://github.com/rhinstaller/anaconda/pull/5508 still means this code won't be called until we patch anaconda.  It's kind of tempting to add some kind of environment variable or at least to force this mode *on*
- If we hit a regression from this, it could be useful in theory also to have a way to easily force *off* this new efibootmgr logic

So I think we could add something like `env BOOTUPD_FORCE_EFIBOOTMGR=on|off` or so?

What I'm a bit uncertain is how easy it is to get environment variables in to the anaconda process itself.

Well, honestly I think my vote is: Let's reopen https://github.com/rhinstaller/anaconda/pull/5508 and merge both it and this to rawhide, after at least some basic testing of both pre-merge.

cgwalters:
OK I've at least tested this doesn't break the virt installation case.  Let's go ahead and get this in so it shows up in https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/package/bootupd/ for easier integration testing.

PR #666, opened 2024-06-09
bootupd.spec: remove `%{_unitdir}`

HuijingHei:
Drop systemd service since https://github.com/coreos/bootupd/commit/261fb5e7f28b9bd2c6099fd2d484f158adffb221, so we do not need this. See error in ci log:
```
RPM build errors:
    File not found: /builddir/build/BUILD/bootupd-xxx-build/BUILDROOT/usr/lib/systemd/system/*
```

### DIFF
--- a/contrib/packaging/bootupd.spec
+++ b/contrib/packaging/bootupd.spec
@@ -29,8 +29,7 @@ BuildRequires: systemd-devel
 %license LICENSE
 %doc README.md
 %{_bindir}/bootupctl
-%{_libexecdir}/bootupd  
-%{_unitdir}/*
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

cgwalters:
Not sure what's up with the jenkins run, but I don't think it can be related.

HuijingHei:
> Not sure what's up with the jenkins run

The error is in https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/build/7557148/, select `Log Detective` -> `Report fedora-rawhide-x86_64` for example, after this PR merged, build is succeeded.

Issue #454, opened 2023-03-30
Update the ESP by creating a tmpdir and RENAME_EXCHANGE


martinezjavier:
Since Linux v6.0 the vfat filesystem has support for `renameat2(..., RENAME_EXCHANGE)`:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=da87e1725ae2

So `bootupd` could instead of applying the diff files one by one to the destination dir, it could create a temporary dir that is a copy of the existing ESP, apply the diff to that temp dir and finally do an atomic rename exchange of the two directories.

That way, the update mechanism will be safer since it would only require a single `renameat2()` system call.

martinezjavier:
`openat::Dir` already has a `local_rename` that seems to support this: https://docs.rs/openat/latest/openat/struct.Dir.html#method.local_rename

cgwalters:
The cost here is write amplification; because FAT doesn't have reflinks, every update to the ESP would require rewriting every file, and a transient doubling of disk space.

travier:
It looks like  [`local_exchange`](https://docs.rs/openat/latest/openat/struct.Dir.html#method.local_exchange) does what we want in Rust:
- https://docs.rs/openat/latest/src/openat/dir.rs.html#366-377
- https://docs.rs/openat/latest/src/openat/dir.rs.html#540-573

travier:
So the logic would be, in pseudo code:

- `cp -a fedora .fedora.tmp` (to make sure to keep all other files that we do not explicitly track in bootupd)
- Update the content of `.fedora.tmp` with the new binaries
- Exchange `.fedora.tmp` -> `fedora`
- Remove now "old" `.fedora.tmp`

HuijingHei:
> * `cp -a fedora .fedora.tmp`

Do you mean the whole EFI dir that includes both `BOOT` and `fedora`?

>   * We start with a copy to make sure to keep all other files that we do not explicitly track in bootupd
> * Update the content of `.fedora.tmp` with the new binaries
> * Exchange `.fedora.tmp` -> `fedora`
> * Remove now "old" `.fedora.tmp`

Do we need to rename `.fedora.tmp` to other name to keep it, then it will be easy to do https://github.com/coreos/bootupd/issues/440 ?

cgwalters:
One thing I wanted to note here that's pretty important is: https://github.com/coreos/bootupd/pull/669#discussion_r1659273299

[...]

Issue #440, opened 2023-03-17
Would it be possible to do A/B updates using `BootNext` from https://github.com/rhboot/efibootmgr?


travier:
Bootupd would write the new bootloader in a new place (B), set BootNext to a new boot entry with this bootloader. And then on reboot would check that the boot succeeded and would do the reverse for the next update (use A).

travier:
If this works, that would potentially let us enable boot loader updates by default (just like it is on dnf based Fedora) has the update itself would be safe.

travier:
Related to https://github.com/rhboot/shim/pull/502

martinezjavier:
#454 is something that could be complementary to this and make the update mechanism safer as well.

pzmarzly:
This was recently discussed at All Systems Go 2024 talk [no more bootloader: please use the kernel instead](https://www.youtube.com/watch?v=xexyAs5Zzxo) (given by @marta-lewandowska, while [the project](https://github.com/rhboot/nmbl-poc) is co-developed with @vathpela). The main concert raised by the audience was NVRAM durability - writing to BootNext on every update could wear down the flash storage, and thus brick the motherboard.

This is a valid concern, one that I've heard before elsewhere, only the last time it was about monitors potentially dying because of f.lux and Lunar.app. The Lunar.app developer [disagrees with this](https://lunar.fyi/faq#ddc-safe), but I guess collecting more data (by conducting experiments on common hardware) would be great.

But even if motherboard chips were to die after the 100K writes like some theorise, that'd still give us space for 50K updates (2 operations: we set BootNext, the firmware unsets BootNext), which is a lot, even if someone updates kernel many times a day (kernel developers). We should be careful not to perform more writes than we need - e.g. rather than creating/deleting the boot entries on every update, we only change BootNext. How? We could create `current` and `next` entries on first installation, and then just one-time activate the `next` entry by setting the BootNext on update:
```
0001 coreos-current (\EFI\boot\bootx64.efi)
0002 coreos-next (\EFI\boot\bootx64-next.efi)
0003 Windows Boot Manager (\EFI\Microsoft\Boot\Bootmgfw.efi)
BootOrder: 0001,0003
BootNext: 0002
```

Does that make sense?

gregory-lee-bartholomew:
It's too bad UEFI doesn't let you point at a directory instead of a specific executable. It'd be great if you could just drop UKIs into a directory on the ESP and then the firmware would iterate over them in version-sorted order until one succeeded.

PR #669, opened 2024-06-19
efi: update the ESP by creating a tmpdir and RENAME_EXCHANGE

HuijingHei:
See Timothée's comment https://github.com/coreos/bootupd/issues/454#issuecomment-2178227050
Reuse `TMP_PREFIX`, logic is like this:
- `cp -a fedora .btmp.fedora`
  - We start with a copy to make sure to keep all other files
that we do not explicitly track in bootupd
- Update the content of `.btmp.fedora` with the new binaries
- Exchange `.btmp.fedora` -> `fedora`
- Remove now "old" `.btmp.fedora`

If we have a file not in a directory in `EFI`, then we can copy
it to `.btmp.foo` and then act on it and finally rename it. No
need to copy the entire `EFI`.

Additional info for the logic:
- for removals, copy to temp dir, remove file in temp dir; if 
have a file not in a directory, remove it directly
- for changes/additions, copy to temp dir, write files in temp 
dir; if have a file not in a directory, copy as temp file
- Do local exchange/rename and remove temp dir/file

And use `insert()` instead of `push()` to match `starts_with()`
 when scanning temp files & dirs.

Fixes https://github.com/coreos/bootupd/issues/454

---

filetree: add failpoint when doing exchange
Inspired by https://github.com/coreos/bootupd/pull/669#issuecomment-2220760948

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -141,7 +141,7 @@ impl Efi {
             log::debug!("Not booted via EFI, skipping firmware update");
             return Ok(());
         }
-        let sysroot = Dir::open_ambient_dir(&Path::new("/"), cap_std::ambient_authority())?;
+        let sysroot = Dir::open_ambient_dir("/", cap_std::ambient_authority())?;
         let product_name = get_product_name(&sysroot)?;
         log::debug!("Get product name: {product_name}");
         assert!(product_name.len() > 0);
--- a/src/filetree.rs
+++ b/src/filetree.rs
@@ -7,17 +7,20 @@
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 use anyhow::{bail, Context, Result};
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+use camino::{Utf8Path, Utf8PathBuf};
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 use openat_ext::OpenatDirExt;
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 use openssl::hash::{Hasher, MessageDigest};
+use rustix::fd::BorrowedFd;
 use serde::{Deserialize, Serialize};
 #[allow(unused_imports)]
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fmt::Display;
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 use std::os::unix::io::AsRawFd;
-#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-use std::path::Path;
+use std::os::unix::process::CommandExt;
+use std::process::Command;
 
 /// The prefix we apply to our temporary files.
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
@@ -231,7 +234,7 @@ impl FileTree {
     }
 }
 
-// Recursively remove all files in the directory that start with our TMP_PREFIX
+// Recursively remove all files/dirs in the directory that start with our TMP_PREFIX
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 fn cleanup_tmp(dir: &openat::Dir) -> Result<()> {
     for entry in dir.list_dir(".")? {
@@ -245,8 +248,13 @@ fn cleanup_tmp(dir: &openat::Dir) -> Result<()> {
 
         match dir.get_file_type(&entry)? {
             openat::SimpleType::Dir => {
-                let child = dir.sub_dir(name)?;
-                cleanup_tmp(&child)?;
+                if name.starts_with(TMP_PREFIX) {
+                    dir.remove_all(name)?;
+                    continue;
+                } else {
+                    let child = dir.sub_dir(name)?;
+                    cleanup_tmp(&child)?;
+                }
             }
             openat::SimpleType::File => {
                 if name.starts_with(TMP_PREFIX) {
@@ -272,7 +280,6 @@ pub(crate) struct ApplyUpdateOptions {
 // Let's just fork off a helper process for now.
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
 pub(crate) fn syncfs(d: &openat::Dir) -> Result<()> {
-    use rustix::fd::BorrowedFd;
     use rustix::fs::{Mode, OFlags};
     let d = unsafe { BorrowedFd::borrow_raw(d.as_raw_fd()) };
     let oflags = OFlags::RDONLY | OFlags::CLOEXEC | OFlags::DIRECTORY;
@@ -280,12 +287,37 @@ pub(crate) fn syncfs(d: &openat::Dir) -> Result<()> {
     rustix::fs::syncfs(d).map_err(Into::into)
 }
 
+/// Copy from src to dst at root dir
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-fn tmpname_for_path<P: AsRef<Path>>(path: P) -> std::path::PathBuf {
-    let path = path.as_ref();
-    let mut buf = path.file_name().expect("filename").to_os_string();
-    buf.push(TMP_PREFIX);
-    path.with_file_name(buf)
+fn copy_dir(root: &openat::Dir, src: &str, dst: &str) -> Result<()> {
+    let rootfd = unsafe { BorrowedFd::borrow_raw(root.as_raw_fd()) };
+    let r = unsafe {
+        Command::new("cp")
+            .args(["-a"])
+            .arg(src)
+            .arg(dst)
+            .pre_exec(move || rustix::process::fchdir(rootfd).map_err(Into::into))
+            .status()?
+    };
+    if !r.success() {
+        anyhow::bail!("Failed to copy {src} to {dst}");
+    }
+    log::debug!("Copy {src} to {dst}");
+    Ok(())
+}
+
+/// Get first sub dir and tmp sub dir for the path
+/// "fedora/foo/bar" -> ("fedora", ".btmp.fedora")
+/// "foo" -> ("foo", ".btmp.foo")
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+fn get_first_dir(path: &Utf8Path) -> Result<(&Utf8Path, String)> {
+    let first = path
+        .iter()
+        .next()
+        .ok_or_else(|| anyhow::anyhow!("Invalid path: {path}"))?;
+    let mut tmp = first.to_owned();
+    tmp.insert_str(0, TMP_PREFIX);
+    Ok((first.into(), tmp))
 }
 
 /// Given two directories, apply a diff generated from srcdir to destdir
@@ -302,41 +334,83 @@ pub(crate) fn apply_diff(
     let opts = opts.unwrap_or(&default_opts);
     cleanup_tmp(destdir).context("cleaning up temporary files")?;
 
-    // Write new and changed files
-    for pathstr in diff.additions.iter().chain(diff.changes.iter()) {
-        let path = Path::new(pathstr);
-        if let Some(parent) = path.parent() {
-            destdir.ensure_dir_all(parent, DEFAULT_FILE_MODE)?;
+    let mut updates = HashMap::new();
+    // Handle removals in temp dir, or remove directly if file not in dir
+    if !opts.skip_removals {
+        for pathstr in diff.removals.iter() {
+            let path = Utf8Path::new(pathstr);
+            let (first_dir, first_dir_tmp) = get_first_dir(path)?;
+            let path_tmp;
+            if first_dir != path {
+                path_tmp = Utf8Path::new(&first_dir_tmp).join(path.strip_prefix(&first_dir)?);
+                // copy to temp dir and remember
+                if !destdir.exists(&first_dir_tmp)? {
+                    copy_dir(destdir, first_dir.as_str(), &first_dir_tmp)?;
+                    updates.insert(first_dir, first_dir_tmp);
+                }
+            } else {
+                path_tmp = path.to_path_buf();
+            }
+            destdir
+                .remove_file(path_tmp.as_std_path())
+                .with_context(|| format!("removing {:?}", path_tmp))?;
         }
-        let destp = tmpname_for_path(path);
-        srcdir
-            .copy_file_at(path, destdir, destp.as_path())
-            .with_context(|| format!("writing {}", &pathstr))?;
     }
-    // Ensure all of the new files are written persistently to disk
+    // Write changed or new files to temp dir or temp file
+    for pathstr in diff.changes.iter().chain(diff.additions.iter()) {
+        let path = Utf8Path::new(pathstr);
+        let (first_dir, first_dir_tmp) = get_first_dir(path)?;
+        let mut path_tmp = Utf8PathBuf::from(&first_dir_tmp);
+        if first_dir != path {
+            if !destdir.exists(&first_dir_tmp)? && destdir.exists(first_dir.as_std_path())? {
+                // copy to temp dir if not exists
+                copy_dir(destdir, first_dir.as_str(), &first_dir_tmp)?;
+            }
+            path_tmp = path_tmp.join(path.strip_prefix(&first_dir)?);
+            // ensure new additions dir exists
+            if let Some(parent) = path_tmp.parent() {
+                destdir.ensure_dir_all(parent.as_std_path(), DEFAULT_FILE_MODE)?;
+            }
+            // remove changed file before copying
+            destdir
+                .remove_file_optional(path_tmp.as_std_path())
+                .with_context(|| format!("removing {path_tmp} before copying"))?;
+        }
+        updates.insert(first_dir, first_dir_tmp);
+        srcdir
+            .copy_file_at(path.as_std_path(), destdir, path_tmp.as_std_path())
+            .with_context(|| format!("copying {:?} to {:?}", path, path_tmp))?;
+    }
+
+    // do local exchange or rename
+    for (dst, tmp) in updates.iter() {
+        let dst = dst.as_std_path();
+        log::trace!("doing local exchange for {} and {:?}", tmp, dst);
+        if destdir.exists(dst)? {
+            destdir
+                .local_exchange(tmp, dst)
+                .with_context(|| format!("exchange for {} and {:?}", tmp, dst))?;
+        } else {
+            destdir
+                .local_rename(tmp, dst)
+                .with_context(|| format!("rename for {} and {:?}", tmp, dst))?;
+        }
+    }
+    // Ensure all of the updates & changes are written persistently to disk
     if !opts.skip_sync {
         syncfs(destdir)?;
     }
-    // Now move them all into place (TODO track interruption)
-    for path in diff.additions.iter().chain(diff.changes.iter()) {
-        let pathtmp = tmpname_for_path(path);
-        destdir
-            .local_rename(&pathtmp, path)
-            .with_context(|| format!("renaming {path}"))?;
-    }
-    if !opts.skip_removals {
-        for path in diff.removals.iter() {
-            destdir
-                .remove_file_optional(path)
-                .with_context(|| format!("removing {path}"))?;
-        }
+
+    // finally remove the temp dir
+    for (_, tmp) in updates.iter() {
+        log::trace!("cleanup: {}", tmp);
+        destdir.remove_all(tmp).context("clean up temp")?;
     }
     // A second full filesystem sync to narrow any races rather than
     // waiting for writeback to kick in.
     if !opts.skip_sync {
         syncfs(destdir)?;
     }
-
     Ok(())
 }
 
@@ -345,6 +419,7 @@ mod tests {
     use super::*;
     use std::fs;
     use std::io::Write;
+    use std::path::Path;
 
     fn run_diff(a: &openat::Dir, b: &openat::Dir) -> Result<FileTreeDiff> {
         let ta = FileTree::new_from_dir(a)?;
@@ -506,6 +581,131 @@ mod tests {
             "newgrub data"
         );
         assert!(!a.join(relp).join("shim.x64").exists());
+        Ok(())
+    }
+    #[test]
+    fn test_get_first_dir() -> Result<()> {
+        // test path
+        let path = Utf8Path::new("foo/subdir/bar");
+        let (tp, tp_tmp) = get_first_dir(path)?;
+        assert_eq!(tp, Utf8Path::new("foo"));
+        assert_eq!(tp_tmp, ".btmp.foo");
+        // test file
+        let path = Utf8Path::new("testfile");
+        let (tp, tp_tmp) = get_first_dir(path)?;
+        assert_eq!(tp, Utf8Path::new("testfile"));
+        assert_eq!(tp_tmp, ".btmp.testfile");
+        Ok(())
+    }
+    #[test]
+    fn test_cleanup_tmp() -> Result<()> {
+        let tmpd = tempfile::tempdir()?;
+        let p = tmpd.path();
+        let pa = p.join("a/.btmp.a");
+        let pb = p.join(".btmp.b/b");
+        std::fs::create_dir_all(&pa)?;
+        std::fs::create_dir_all(&pb)?;
+        let dp = openat::Dir::open(p)?;
+        {
+            let mut buf = dp.write_file("a/foo", 0o644)?;
+            buf.write_all("foocontents".as_bytes())?;
+            let mut buf = dp.write_file("a/.btmp.foo", 0o644)?;
+            buf.write_all("foocontents".as_bytes())?;
+            let mut buf = dp.write_file(".btmp.b/foo", 0o644)?;
+            buf.write_all("foocontents".as_bytes())?;
+        }
+        assert!(dp.exists("a/.btmp.a")?);
+        assert!(dp.exists("a/foo")?);
+        assert!(dp.exists("a/.btmp.foo")?);
+        assert!(dp.exists("a/.btmp.a")?);
+        assert!(dp.exists(".btmp.b/b")?);
+        assert!(dp.exists(".btmp.b/foo")?);
+        cleanup_tmp(&dp)?;
+        assert!(!dp.exists("a/.btmp.a")?);
+        assert!(dp.exists("a/foo")?);
+        assert!(!dp.exists("a/.btmp.foo")?);
+        assert!(!dp.exists(".btmp.b")?);
+        Ok(())
+    }
+    #[test]
+    fn test_apply_with_file() -> Result<()> {
+        let tmpd = tempfile::tempdir()?;
+        let p = tmpd.path();
+        let pa = p.join("a");
+        let pb = p.join("b");
+        std::fs::create_dir(&pa)?;
+        std::fs::create_dir(&pb)?;
+        let a = openat::Dir::open(&pa)?;
+        let b = openat::Dir::open(&pb)?;
+        a.create_dir("foo", 0o755)?;
+        a.create_dir("bar", 0o755)?;
+        let foo = Path::new("foo/bar");
+        let bar = Path::new("bar/foo");
+        let testfile = "testfile";
+        {
+            let mut buf = a.write_file(foo, 0o644)?;
+            buf.write_all("foocontents".as_bytes())?;
+            let mut buf = a.write_file(bar, 0o644)?;
+            buf.write_all("barcontents".as_bytes())?;
+            let mut buf = a.write_file(testfile, 0o644)?;
+            buf.write_all("testfilecontents".as_bytes())?;
+        }
+
+        let diff = run_diff(&a, &b)?;
+        assert_eq!(diff.count(), 3);
+        b.create_dir("foo", 0o755)?;
+        {
+            let mut buf = b.write_file(foo, 0o644)?;
+            buf.write_all("foocontents".as_bytes())?;
+        }
+        let b_btime_foo = fs::metadata(pb.join(foo))?.created()?;
+
+        {
+            let diff = run_diff(&b, &a)?;
+            assert_eq!(diff.count(), 2);
+            apply_diff(&a, &b, &diff, None).context("test additional files")?;
+            assert_eq!(
+                String::from_utf8(std::fs::read(pb.join(testfile))?)?,
+                "testfilecontents"
+            );
+            assert_eq!(
+                String::from_utf8(std::fs::read(pb.join(bar))?)?,
+                "barcontents"
+            );
+            // creation time is not changed for unchanged file
+            let b_btime_foo_new = fs::metadata(pb.join(foo))?.created()?;
+            assert_eq!(b_btime_foo_new, b_btime_foo);
+        }
+        {
+            fs::write(pa.join(testfile), "newtestfile")?;
+            fs::write(pa.join(bar), "newbar")?;
+            let diff = run_diff(&b, &a)?;
+            assert_eq!(diff.count(), 2);
+            apply_diff(&a, &b, &diff, None).context("test changed files")?;
+            assert_eq!(
+                String::from_utf8(std::fs::read(pb.join(testfile))?)?,
+                "newtestfile"
+            );
+            assert_eq!(String::from_utf8(std::fs::read(pb.join(bar))?)?, "newbar");
+            // creation time is not changed for unchanged file
+            let b_btime_foo_new = fs::metadata(pb.join(foo))?.created()?;
+            assert_eq!(b_btime_foo_new, b_btime_foo);
+        }
+        {
+            a.remove_file(testfile)?;
+            a.remove_file(bar)?;
+            let diff = run_diff(&b, &a)?;
+            assert_eq!(diff.count(), 2);
+            apply_diff(&a, &b, &diff, None).context("test removed files")?;
+            assert_eq!(b.exists(testfile)?, false);
+            assert_eq!(b.exists(bar)?, false);
+            let diff = run_diff(&b, &a)?;
+            assert_eq!(diff.count(), 0);
+            // creation time is not changed for unchanged file
+            let b_btime_foo_new = fs::metadata(pb.join(foo))?.created()?;
+            assert_eq!(b_btime_foo_new, b_btime_foo);
+        }
+
         Ok(())
     }
 }
--- a/src/filetree.rs
+++ b/src/filetree.rs
@@ -395,6 +395,7 @@ pub(crate) fn apply_diff(
                 .local_rename(tmp, dst)
                 .with_context(|| format!("rename for {} and {:?}", tmp, dst))?;
         }
+        crate::try_fail_point!("update::exchange");
     }
     // Ensure all of the updates & changes are written persistently to disk
     if !opts.skip_sync {
@@ -705,7 +706,6 @@ mod tests {
             let b_btime_foo_new = fs::metadata(pb.join(foo))?.created()?;
             assert_eq!(b_btime_foo_new, b_btime_foo);
         }
-
         Ok(())
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,6 +129,7 @@ version = "0.2.20"
 dependencies = [
  "anyhow",
  "bincode",
+ "camino",
  "cap-std-ext",
  "chrono",
  "clap",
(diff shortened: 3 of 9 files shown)


src/efi.rs
```diff
@@ -72,6 +72,17 @@ impl Efi {
         Ok(esp)
     }
 
+    fn esp_path_tmp(&self) -> Result<PathBuf> {
+        self.ensure_mounted_esp(Path::new("/"))
+            .map(|v| v.join(".EFI.tmp"))
+    }
```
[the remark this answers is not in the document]
cgwalters: But...related to @travier's comment...it'd be good to try to scope this to the "lowest" directory we need to make the change.

For us that means e.g. we only affect `EFI/fedora` for example and not all of `EFI`.

cgwalters (commented):
Thanks for working on this!

src/efi.rs
```diff
@@ -580,6 +619,28 @@ fn find_file_recursive<P: AsRef<Path>>(dir: P, target_file: &str) -> Result<Vec<
     Ok(result)
 }
 
+fn copy_dir_all(src: &Path, dst: &Path) -> Result<()> {
```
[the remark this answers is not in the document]
HuijingHei: SGTM, thanks!

src/efi.rs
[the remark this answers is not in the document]
HuijingHei: The `bootupd-state.json` also includes `BOOT/BOOTX64.EFI` and `BOOT/fbx64.efi` under `EFI/`, both are from shim-x64, we might need to skip the update?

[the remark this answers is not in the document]
travier: I think it should be reasonable to prepare both the new `BOOT` and `fedora` dirs and then do `fedora` first, `BOOT` second.

I don't think there is a dependency "across" those folders for the bootloader files, i.e. grub.efi may depend on an update shim.efi from the same folder but not on an update fbx.efy from the BOOT folder.

I'm not sure however.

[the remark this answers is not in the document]
travier: There were some elements in https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/VRJ7CRQUZ7SNEDW4AIOHKKJ7TMPTOX5A/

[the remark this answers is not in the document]
travier: https://github.com/rhboot/shim/pull/502

src/efi.rs
```diff
@@ -580,6 +609,18 @@ fn find_file_recursive<P: AsRef<Path>>(dir: P, target_file: &str) -> Result<Vec<
     Ok(result)
 }
 
+fn copy_dir(src: &Path, dst: &Path) -> Result<()> {
+    let r = std::process::Command::new("cp")
```
travier, line 613: Let's find a more "Rust" way to do that.

travier: Looks like there isn't anything readily available in the standard library. I found https://stackoverflow.com/questions/26958489/how-to-copy-a-folder-recursively-in-rust as a starting point but it does not preserve access times, etc. like `cp -a` does.

travier: It might be reasonable to leave this as is as the code in `cp` is battle tested and very unlikely to fail.

src/efi.rs
```diff
@@ -348,12 +348,41 @@ impl Component for Efi {
             .context("opening update dir")?;
         let updatef = filetree::FileTree::new_from_dir(&updated).context("reading update dir")?;
         let diff = currentf.diff(&updatef)?;
-        self.ensure_mounted_esp(Path::new("/"))?;
```
travier, line 351: I think we still need this one?

travier:
```
[2024-06-20T08:07:06.180Z] # Previous BIOS: grub2-tools-1:2.06-89.fc38.x86_64
[2024-06-20T08:07:06.180Z] # Updated BIOS: grub2-tools-1:2.06-102.fc38.x86_64
[2024-06-20T08:07:06.180Z] # error: Failed to update EFI: No such file or directory (os error 2)
[2024-06-20T08:07:06.180Z] + fatal 'test failed'
[2024-06-20T08:07:06.180Z] + echo error: test failed
[2024-06-20T08:07:06.180Z] error: test failed
```
:think

src/efi.rs
HuijingHei: See Colin's suggestion https://github.com/coreos/bootupd/pull/669#discussion_r1646764899 to do `cp -a`, just curious how do you check it does not preserve access times?

src/efi.rs
```diff
@@ -348,12 +348,45 @@ impl Component for Efi {
...
+        } else {
+            bail!("Failed to get vendor dir");
+        };
+
+        let esp = self.esp_path()?;
+        let vendordir = esp.join(&vendor);
+        let tmp_vendordir = esp.join(format!(".{vendor}.tmp"));
+        // remove a previous directory if it exists in order to handle being interrupted in the middle.
+        if tmp_vendordir.exists() {
+            std::fs::remove_dir_all(&tmp_vendordir)?;
```
HuijingHei, line 366: Maybe it is because of removing dir without checking if it exists.

HuijingHei:
```
[2024-06-20T12:39:28.464Z] # Previous BIOS: grub2-tools-1:2.06-89.fc38.x86_64

[2024-06-20T12:39:28.464Z] # Updated BIOS: grub2-tools-1:2.06-102.fc38.x86_64

[2024-06-20T12:39:28.464Z] # Previous EFI: grub2-efi-x64-1:2.06-89.fc38.x86_64,shim-x64-15.8-3.x86_64

[2024-06-20T12:39:28.464Z] # Updated EFI: grub2-efi-x64-1:2.06-102.fc38.x86_64,shim-x64-15.8-3.x86_64,test-bootupd-payload-1.0-1.x86_64

[2024-06-20T12:39:28.464Z] # grep: /tmp/tmp.hin3bC02Ci/EFI/fedora/test-bootupd.efi: No such file or directory

[2024-06-20T12:39:28.464Z] # ls: cannot access '/tmp/tmp.hin3bC02Ci/EFI/fedora/test-bootupd.efi': No such file or directory
```

src/efi.rs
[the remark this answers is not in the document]
HuijingHei: > prepare both the new `BOOT` and `fedora` dirs and then do `fedora` first, `BOOT` second.

This is a little tricky that for the `apply_diff`, need to check if it is for `BOOT` or `fedora`, then how about using ".EFI.tmp"?

travier:
We've merged the CI fixes. Can you rebase this one? Thanks

HuijingHei:
> We've merged the CI fixes. Can you rebase this one? Thanks

Sure, updated, thanks!

HuijingHei:
Updating `shim-x64 15.8-2 -> 15.8-3`, it will also update `BOOT/fbx64.efi`

```
[root@cosa-devsh ~]# bootupctl update -vvv
Previous EFI: grub2-efi-x64-1:2.06-103.fc40.x86_64,shim-x64-15.8-2.x86_64
Updated EFI: grub2-efi-x64-1:2.06-123.fc40.x86_64,shim-x64-15.8-3.x86_64

[root@cosa-devsh ~]# mount /dev/vda2 /boot/efi/
[root@cosa-devsh ~]# ls /boot/efi/EFI/BOOT/ -l
total 1014
-rwxr-xr-x. 1 root root 949424 Jan  1  1980 BOOTX64.EFI
-rwxr-xr-x. 1 root root  87816 Jun 21 10:01 fbx64.efi
```

travier:
The problem with copying the entire `EFI` directory is that we don't know what's there, and there might a lot of other vendor directories (`windows`, `ubuntu`, etc.) that we don't really want to copy as that might take too much space that what we have in the EFI partition.

HuijingHei:
> The problem with copying the entire `EFI` directory is that we don't know what's there, and there might a lot of other vendor directories (`windows`, `ubuntu`, etc.) that we don't really want to copy as that might take more space than what we have in the EFI partition.

Makes sense, we might need to do some change to `apply_diff`, if we make temp for `BOOT` and `fedora`.

src/efi.rs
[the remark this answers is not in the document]
HuijingHei: See https://github.com/coreos/bootupd/pull/669#issuecomment-2182555634, will prepare both the new `BOOT` and `fedora` dirs and then do `fedora` first, `BOOT` second.

src/filetree.rs
```diff
@@ -280,12 +280,63 @@ pub(crate) fn syncfs(d: &openat::Dir) -> Result<()> {
...
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-fn tmpname_for_path<P: AsRef<Path>>(path: P) -> std::path::PathBuf {
+fn tmp_for_path<P: AsRef<Path>>(path: P) -> std::path::PathBuf {
     let path = path.as_ref();
-    let mut buf = path.file_name().expect("filename").to_os_string();
-    buf.push(TMP_PREFIX);
-    path.with_file_name(buf)
+    let mut parent = String::from("");
+    if let Some(p) = path.parent() {
+        parent = format!("{}.tmp", p.display());
```
HuijingHei, line 301: Use `fedora.tmp` instead of `.fedora.tmp`, in case if the parent is like `a/b`, then it will be `.a/b.tmp` which is not expected.

HuijingHei:
Update the change in `apply_diff()`, not sure if it is the right place, steps are:
- get unique updates dir in `diff`, save them in hashset
- scan the hashset, 
  - extract the `diff` based on the same parent
  - copy original dir to temp dir
  - remove original file in temp dir first, then write new and changed files to temp dir
  - do local exchange for `fedora.tmp` to `fedora`
  - remove `fedora.tmp`

src/filetree.rs
```diff
@@ -280,12 +280,63 @@ pub(crate) fn syncfs(d: &openat::Dir) -> Result<()> {
...
+    Ok(())
+}
+
+/// Convert path parent to temp, `fedora/foo` -> `fedora.tmp/foo`
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-fn tmpname_for_path<P: AsRef<Path>>(path: P) -> std::path::PathBuf {
+fn tmp_for_path<P: AsRef<Path>>(path: P) -> std::path::PathBuf {
     let path = path.as_ref();
-    let mut buf = path.file_name().expect("filename").to_os_string();
-    buf.push(TMP_PREFIX);
```
HuijingHei, line 287: Seems the file name is like`a.btmp.`, so to check should not use `start_with`, but maybe I am wrong.

HuijingHei:
> I'm thinking about the following logic for `apply_diff`:
> 
> * First look at removals (to minimize disk space usage)
>   
>   * Look at which "root/parent" folder the file to remove is
>   * Does a temporary copy of this folder exists?
>     
>     * Yes, remove the file in the copy
>     * No:
>       
>       * Copy the folder
>       * Remember that we made a copy of this folder
>       * Remove the file in the copy
> * Repeat the same logic for changes and additions
> * Sync changes to the disk once here

I think this means copy the changed files to prepared temp dir, this seems sane, but it will be failed if the change dir is included by another change, if changes are `EFI/fedora/subdir/newgrub.x64` and `EFI/fedora/grub.x64`, the temp dir will be `EFI/fedora/subdir.tmp` and `EFI/fedora.tmp`, then do exchange `EFI/fedora/subdir.tmp -> EFI/fedora/subdir`, the problem is `EFI/fedora.tmp/grub.x64` will not include the new change `EFI/fedora/subdir`, this is the unit test in `test_filetree2()` that I debug most of the time, so I do the exchange separately based on dir, any other workarounds for this?
It is not problem for the current bootupd(no this patch) as we only update the file, not related to the dir.
   
>   * At this point, we haven't yet touched the content actually used to boot
>   * If there is not enough space to write all the files, we should have failed before this point
> * Iterate over the folders we made a copy of and do RENAME_EXCHANGE
> * Sync changes between each rename (or once at the end?)
> * Remove all the temporary folders
> * Sync one last time (optional?)

travier:
> if changes are EFI/fedora/subdir/newgrub.x64 and EFI/fedora/grub.x64, the temp dir will be EFI/fedora/subdir.tmp and EFI/fedora.tmp

With `EFI/fedora/subdir/newgrub.x64` changed, we should copy `EFI/fedora` to `EFI/fedora.tmp`. That should cover all the subdirs and should granular enough.

HuijingHei:
> > if changes are EFI/fedora/subdir/newgrub.x64 and EFI/fedora/grub.x64, the temp dir will be EFI/fedora/subdir.tmp and EFI/fedora.tmp
> 
> With `EFI/fedora/subdir/newgrub.x64` changed, we should copy `EFI/fedora` to `EFI/fedora.tmp`. That should cover all the subdirs and should be granular enough.

Thanks, LGTM, then need to add condition to check this, if `EFI/fedora.tmp` exists, just update file in `EFI/fedora.tmp/subdir` instead of `EFI/fedora/subdir.tmp`. If sometimes `EFI/fedora/subdir.tmp` comes first, should remove it, does this make sense?

HuijingHei:
> > if changes are EFI/fedora/subdir/newgrub.x64 and EFI/fedora/grub.x64, the temp dir will be EFI/fedora/subdir.tmp and EFI/fedora.tmp
> 
> With `EFI/fedora/subdir/newgrub.x64` changed, we should copy `EFI/fedora` to `EFI/fedora.tmp`. That should cover all the subdirs and should be granular enough.
> 
> I.e. we need to find the first subdir in `EFI` and copy that.

Thanks, will try this. (It might be corner case the change is the same dir with EFI.)

travier:
I'm not sure I understand what you mean by "which comes first". I think the idea is more to think of it as the first subdir in the `EFI` folder instead of thinking about it as "parent" directory.

If I have the following update/addition/removal for example:
* `EFI/fedora/subdir/foo`
* `EFI/fedora/subdir/bar/toto`
* `EFI/fedora/baz`

Then the logic would be:
* For the file `EFI/fedora/subdir/foo` as part of the diff:
  * Look if `fedora.tmp` exists (the first subdir of `EFI`)
  * Copy `fedora` to `fedora.tmp` and remember it
  * Update/remove/add `EFI/fedora.tmp/subdir/foo`
* For the file `EFI/fedora/subdir/bar/toto` as part of the diff:
  * Look if `fedora.tmp` exists
  * I does, so nothing to copy
  * Update/remove/add`EFI/fedora/subdir/bar/toto`
* Repeat for `EFI/fedora/baz`
* Once done for all files in the diff, do the rename for the folders that we remembered we copied:
  * Rename exchange `fedora.tmp` -> `fedora`, etc.

HuijingHei:
> I'm not sure I understand what you mean by "which comes first"

Sorry for the confusion,
if the changes are `EFI/fedora/subdir/newgrub.x64, EFI/fedora/grub.x64`, 
copy to `EFI/fedora/subdir.tmp` first, then comes `EFI/fedora.tmp`, need to remove `EFI/fedora/subdir.tmp` first, just copy to `EFI/fedora.tmp`.

if the changes are `EFI/fedora/grub.x64, EFI/fedora/subdir/newgrub.x64`, 
copy to `EFI/fedora.tmp` first, then comes `EFI/fedora/subdir.tmp`, check parent `EFI/fedora.tmp` exists, just copy to `EFI/fedora.tmp/subdir`.

> For the file EFI/fedora/subdir/foo as part of the diff:
> Look if fedora.tmp exists (the first subdir of EFI)

If the change is like `EFI/foo`, the temp dir should be `EFI.tmp`.

cgwalters:
Thanks for working on this! It's a really important feature. I hope to get some time by the end of this week or early next week to look at this.

src/filetree.rs
```diff
@@ -280,12 +280,63 @@ pub(crate) fn syncfs(d: &openat::Dir) -> Result<()> {
...
+pub(crate) fn diff_with_prefix(diff: &FileTreeDiff, prefix: &str) -> Result<FileTreeDiff> {
+    let mut new_diff = FileTreeDiff {
+        additions: HashSet::new(),
+        removals: HashSet::new(),
+        changes: HashSet::new(),
+    };
+    for pathstr in diff.additions.iter() {
+        let path = Path::new(pathstr);
+        if let Some(parent) = path.parent() {
+            if parent.to_string_lossy().eq(prefix) {
```
cgwalters, line 318: Let's avoid `to_string_lossy()`. Instead in general we can use https://docs.rs/camino/latest/camino/

src/filetree.rs
```diff
@@ -302,40 +353,82 @@ pub(crate) fn apply_diff(
...
-            .with_context(|| format!("writing {}", &pathstr))?;
     }
-    // Ensure all of the new files are written persistently to disk
-    if !opts.skip_sync {
-        syncfs(destdir)?;
-    }
-    // Now move them all into place (TODO track interruption)
-    for path in diff.additions.iter().chain(diff.changes.iter()) {
-        let pathtmp = tmpname_for_path(path);
+    let efipath = destdir.recover_path()?;
```
cgwalters, line 369: This is OK for now; however, it would definitely be better to operate strictly in terms of relative paths where we can because it makes it a lot easier to avoid bugs.

This strongly relates to e.g. https://github.com/coreos/bootupd/pull/674 which aims to start using cap-std - which strictly denies use of absolute paths.

I wouldn't say that's a requirement here, but maybe we do consider porting here to cap-std first. Up for discussion.

src/filetree.rs
```diff
@@ -302,40 +353,82 @@ pub(crate) fn apply_diff(
...
         }
     }
-    // A second full filesystem sync to narrow any races rather than
-    // waiting for writeback to kick in.
-    if !opts.skip_sync {
-        syncfs(destdir)?;
-    }
 
     Ok(())
 }
```
cgwalters, line 434: I think what would really help build confidence here is a few new unit tests. A good way to implement this would be by e.g. checking the ctime (creation time) on files. If we did any extra copying we shouldn't have, then the ctime on the files will have changed when it shouldn't.

src/filetree.rs
```diff
@@ -280,12 +280,63 @@ pub(crate) fn syncfs(d: &openat::Dir) -> Result<()> {
     rustix::fs::syncfs(d).map_err(Into::into)
 }
 
+fn copy_dir(src: &Path, dst: &Path) -> Result<()> {
+    let r = std::process::Command::new("cp")
```
cgwalters, line 284: To make this operate in terms of relative paths, the standard approach I use looks like this:
https://github.com/coreos/bootupd/blob/6a4fc6f8758ff18432d5a43995b8f53ad79d4419/src/filesystem.rs#L33

I think using that we may be able to drop the usage of `recover_path`?

src/filetree.rs
```diff
@@ -280,12 +280,63 @@ pub(crate) fn syncfs(d: &openat::Dir) -> Result<()> {
...
+    let mut parent = String::from("");
+    if let Some(p) = path.parent() {
+        parent = format!("{}.tmp", p.display());
+    }
+    let buf = path.file_name().expect("filename").to_os_string();
+    Path::new(parent.as_str()).join(buf)
+}
+
+/// Get diff based on sub dir
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
```
cgwalters, line 308: Also, while this is not at all directly related to your change, I found all the copy-paste of this conditional (there's a ton! you're only adding a bit more) in this file to be pretty ugly. I did https://github.com/coreos/bootupd/pull/673 in prep for reworking things here.

Probably the simplest thing to do is just to build this on all architectures and slap an `#[allow(dead_code)]` on the whole thing. But that's a distinct change.

src/efi.rs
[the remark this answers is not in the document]
cgwalters: > I don't think there is a dependency "across" those folders for the bootloader files, i.e. grub.efi may depend on an update shim.efi from the same folder but not on an update fbx.efy from the BOOT folder.

Hmmmm...this kind of rains on the parade here though given that `shim` includes files in these two directories, and we can't update it atomically without updating *everything* in `EFI`, which also includes potentially other files we don't manage.

src/filetree.rs
```diff
@@ -302,40 +353,82 @@ pub(crate) fn apply_diff(
...
-    // Ensure all of the new files are written persistently to disk
-    if !opts.skip_sync {
-        syncfs(destdir)?;
-    }
-    // Now move them all into place (TODO track interruption)
-    for path in diff.additions.iter().chain(diff.changes.iter()) {
-        let pathtmp = tmpname_for_path(path);
+    let efipath = destdir.recover_path()?;
+    for path in &updates {
+        let new_diff = diff_with_prefix(diff, path)?;
```
HuijingHei, line 371: @cgwalters one thing want to confirm is that, this logic is using apply/exchange based on updates dir's parent, but according to @travier 's suggestion, should copy first dir as temp (like `BOOT/foo` -> `BOOT.tmp/foo`) and remember, then do remove/addition/change files in temp dir, finally scan temp and do exchange. WDYT?

src/efi.rs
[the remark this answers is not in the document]
HuijingHei: According to @travier 's suggestion, maybe can copy first sub dir as temp (like `BOOT/foo` -> `BOOT.tmp/foo`) and remember, then do remove/addition/change files in temp dir, finally scan temp and do exchange. I think it might be helpful when we do A/B updates. WDYT?

src/filetree.rs
HuijingHei: Sorry that I changed the logic, use the first sub dir, if change like `fedora/subdir/foo`, will copy to `feodra.tmp/`, this might also copy other files that should not update. Not sure if the current change is OK.

travier:
> Sorry for the confusion, if the changes are `EFI/fedora/subdir/newgrub.x64, EFI/fedora/grub.x64`, copy to `EFI/fedora/subdir.tmp` first, then comes `EFI/fedora.tmp`, need to remove `EFI/fedora/subdir.tmp` first, just copy to `EFI/fedora.tmp`.

If we have `EFI/fedora/subdir/newgrub.x64, EFI/fedora/grub.x64`, we only need to copy at the `fedora` level, not at any other sub-levels.

> > For the file EFI/fedora/subdir/foo as part of the diff:
> > Look if fedora.tmp exists (the first subdir of EFI)
> 
> If the change is like `EFI/foo`, the temp dir should be `EFI.tmp`.

If we have a file not in a directory in EFI, then we can copy it to `foo.tmp` and then act on it and finally rename it. No need to copy the entire `EFI`.

travier:
The goal is to make the operation atomic at the `vendor` (i.e. `fedora`, `BOOT`, etc.) directory level, not at the `EFI` or any other sub-directory level.

travier:
(I've not looked at the code again yet)

src/efi.rs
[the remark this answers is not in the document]
cgwalters: Yes, this just means that shim updates won't be transactional, which is unfortunate as shim is the main one we need to update.

It should be an improvement...but...I'm a bit uncertain about saying it's enough that we unconditionally turn on auto-updates by default?

It feels like shim/grub update infrequently enough that in practice we could just accept the hit today of the non-transactionality and recommend turning on auto-updates in various distros/OSes?

I'd be roughly in favor of doing that in FCOS and fedora bootc today just to start.

src/filetree.rs
```diff
@@ -272,20 +274,43 @@ pub(crate) struct ApplyUpdateOptions {
...
 
+/// Copy from src to dst at root dir
+fn copy_dir(root: &openat::Dir, src: &str, dst: &str) -> Result<()> {
+    let rootfd = unsafe { BorrowedFd::borrow_raw(root.as_raw_fd()) };
+    let r = unsafe {
+        Command::new("cp")
+            .args(["-a"])
+            .arg(src)
+            .arg(dst)
+            .pre_exec(move || rustix::process::fchdir(rootfd).map_err(Into::into))
```
HuijingHei, line 292: This can only copy from `src` to `dst` at same `rootfd`, it will not work if `src` or `dst` is not under the same `rootfd`, maybe better to use `cap-std` in https://github.com/coreos/bootupd/pull/674

src/efi.rs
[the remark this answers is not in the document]
HuijingHei: > It should be an improvement...but...I'm a bit uncertain about saying it's enough that we unconditionally turn on auto-updates by default?

Not sure, is it necessary to check the shim files integrity (like `sha512`) after sync to provide more confidence?

src/filetree.rs
```diff
@@ -302,34 +330,85 @@ pub(crate) fn apply_diff(
...
+
+    // do local exchange
+    for (src, tmp) in updates {
+        log::trace!("doing local exchange for {} and {}", tmp, src);
+        destdir
+            .local_exchange(&tmp, &src)
+            .with_context(|| format!("exchange for {:?} and {:?}", tmp, src))?;
+        // finally remove the temp dir
+        log::trace!("cleanup: {:?}", tmp);
+        destdir.remove_all(&tmp).context("clean up temp")?;
```
travier, line 411: Ideally we do that in a second for loop to minimize the time between when we start exchanging the folders and when we complete it. The removal can append at the very end.

src/filetree.rs
```diff
@@ -272,20 +275,45 @@ pub(crate) struct ApplyUpdateOptions {
...
+    log::debug!("Copy {src} to {dst}");
+    Ok(())
+}
+
+/// Get first sub dir and tmp sub dir for the path
+/// "fedora/foo/bar" -> ("fedora", "fedora.tmp")
+/// "foo" -> ("foo", "foo.tmp")
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+fn get_subdir(path: &Utf8Path) -> Result<(String, String)> {
+    let buf = if let Some(p) = path.parent().filter(|&v| !v.as_os_str().is_empty()) {
```
travier, line 309: Maybe https://docs.rs/camino/latest/camino/struct.Utf8Path.html#method.components would make it easier than iterating over the parents.

src/filetree.rs
```diff
@@ -302,34 +330,85 @@ pub(crate) fn apply_diff(
...
         syncfs(destdir)?;
     }
-    // Now move them all into place (TODO track interruption)
-    for path in diff.additions.iter().chain(diff.changes.iter()) {
-        let pathtmp = tmpname_for_path(path);
-        destdir
-            .local_rename(&pathtmp, path)
-            .with_context(|| format!("renaming {path}"))?;
+
+    // Write new files to temp dir if exists, else write directly in dest
```
travier, line 360: Theoretically, a binary could change behavior if another file is there or not.

So maybe let's move that after the changed files loop to make it safer. Or we add only to a temp dir copy in all cases.

src/filetree.rs
HuijingHei: Agree with you, no need to iter parent, just iter path and get iter.next() is enough.
If using components, need to import module camino::Utf8Component, but I think iter the path is enough. WDTH?

src/filetree.rs
HuijingHei: > So maybe let's move that after the changed files loop to make it safer. Or we add only to a temp dir copy in all cases.

Move additions after changed seems easier.
If move to temp dir, it might be different as there are no original copy and finally will use `rename()` instead of `exchange()`. We can change to temp if it is necessary. WDYT?

travier: Note that there will be no original copy only if the file is at the root of `EFI/`, which is rather unlikely as this is not the case in any of our distributions (but should still be handled nevertheless if we do this).

src/filetree.rs
```diff
@@ -302,34 +326,88 @@ pub(crate) fn apply_diff(
...
         }
-        let destp = tmpname_for_path(path);
-        srcdir
-            .copy_file_at(path, destdir, destp.as_path())
-            .with_context(|| format!("writing {}", &pathstr))?;
     }
     // Ensure all of the new files are written persistently to disk
     if !opts.skip_sync {
         syncfs(destdir)?;
     }
```
travier, line 355: This should happen right after we renamed the temp dirs

src/filetree.rs
```diff
@@ -302,34 +326,88 @@ pub(crate) fn apply_diff(
...
+    }
+
+    // do local exchange
+    for (src, tmp) in updates.iter() {
+        log::trace!("doing local exchange for {} and {}", tmp, src);
+        destdir
+            .local_exchange(tmp, src)
+            .with_context(|| format!("exchange for {} and {}", tmp, src))?;
+    }
+    // finally remove the temp dir
```
travier, line 407: Sync here, before removing the temp dirs

travier (commented):
Looks like we're mostly there. I wonder if we can make a test to verify the case where the update fails in the middle, right between the two rename exchange calls.

src/filetree.rs
HuijingHei: And use `insert()` instead of `push()` to match `starts_with()`

src/filetree.rs
HuijingHei: OK, change logic here: if additional file is at the root of `EFI/` or new subdir that destination does not have, for example `testfile`, `fedora/newsub/newfile`, then write to `.btmp.testfile` , `fedora/newsub/.btmp.newfile` (reuse `TMP_PREFIX`), finally do `rename()`, is this accepted?

travier: > testfile -> .btmp.testfile

Agree

> fedora/newsub/newfile -> fedora/newsub/.btmp.newfile

That creates another case, and we would have to clean those files on failure. I would prefer we stick to copying the entire dir in the non-top-level case: `fedora/newsub/newfile` -> fedora.tmp/newsub/newfile` + rename_exchange.

travier:
> make a test to verify the case where the update fails in the middle, right between the two rename exchange calls.

We need to make sure that we can properly resume an update if it fails in the middle, when half has already been exchanged and the other half still needs to be exchanged (`BOOT.tmp` -> `BOOT` exchange already done, `fedora.tmp` -> `fedora` exchange still pending). It would be incredibly unlikely to happen in practice but this is the point of this PR thus it would be great to have that.

travier:
That might need a special "fail" mode flag for testing however so might not be desireable.

src/filetree.rs
HuijingHei: > copying the entire dir in the non-top-level case: `fedora/newsub/newfile` -> `fedora.tmp/newsub/newfile` + rename_exchange.

Updated, thanks!

HuijingHei:
> > make a test to verify the case where the update fails in the middle, right between the two rename exchange calls.
> 
> We need to make sure that we can properly resume an update if it fails in the middle, when half has already been exchanged and the other half still needs to be exchanged (`BOOT.tmp` -> `BOOT` exchange already done, `fedora.tmp` -> `fedora` exchange still pending)

This is likely to happen, for example, the power can be outage at anytime during update.
Not sure about the resume, if running `bootupctl update` again, `apply_diff()` will firstly call `cleanup_tmp()` to remove all temp dirs/files, and then do the sync. I can manually do the testing for this firstly. Not sure about the auto test.

cgwalters:
> Edit: Not easy to do manually testing, as the update only uses about 1.5s.

We could do https://github.com/coreos/rpm-ostree/pull/4259/commits/aa8d7fb0ceaabfaf10252180e2ddee049d07aae3 in this repo too

cgwalters:
I think my high level feelings on this are basically:

- I liked the original idea generally
- It feels 40% less useful if we can't do transactional updates for shim (to rephrase: it's still useful, just less so)
- This is complex and subtle code and problem domain and while bootupd has tests (and thanks for adding new unit tests I just remain in an uncertain feeling due to the above combined

src/filetree.rs
cgwalters: I think for our use case here we're always operating on the ESP, so we can just use that as the root fd and that's fine.

src/filetree.rs
```diff
@@ -508,4 +595,127 @@ mod tests {
         assert!(!a.join(relp).join("shim.x64").exists());
         Ok(())
     }
+    #[test]
+    fn test_get_subdir() -> Result<()> {
+        // test path
+        let path = Utf8Path::new("foo/subdir/bar");
+        let (tp, tp_tmp) = get_subdir(path)?;
+        assert_eq!(tp.eq("foo"), tp_tmp.eq(".btmp.foo"));
```
cgwalters, line 603: This seems odd - we're asserting on the results of two comparisons?
Did you mean to do
```
assert_eq!(tp, "foo");
assert_eq(tp_tmp, ".btmp.foo");
```
?

src/filetree.rs
```diff
@@ -272,20 +280,41 @@ pub(crate) struct ApplyUpdateOptions {
...
+    }
+    log::debug!("Copy {src} to {dst}");
+    Ok(())
+}
+
+/// Get first sub dir and tmp sub dir for the path
+/// "fedora/foo/bar" -> ("fedora", ".btmp.fedora")
+/// "foo" -> ("foo", ".btmp.foo")
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+fn get_subdir(path: &Utf8Path) -> Result<(String, String)> {
```
cgwalters, line 313: This returns a `Result` but currently cannot fail

src/filetree.rs
```diff
@@ -272,20 +280,41 @@ pub(crate) struct ApplyUpdateOptions {
...
+    let buf = path.iter().next().unwrap().to_owned();
+    let mut buf_tmp = buf.clone();
+    buf_tmp.insert_str(0, TMP_PREFIX);
+    Ok((buf, buf_tmp))
```
cgwalters, line 317: I think this can look like this instead:

```
let first = path.iter().next().ok_or_else(|| anyhow!("Invalid path: {path}"))?;
let mut tmp = first.as_str().to_owned();
tmp.insert_str(0, TMP_PREFIX);
Ok((first, tmp))
```

The return value would then be `Result<(&Utf8Path, String)>` - notice we didn't need to make a copy of the first one, we can just return a borrowed value.

src/filetree.rs
HuijingHei: Yes, you are correct, thanks!

src/filetree.rs
HuijingHei: LGTM, thanks, updated.

src/filetree.rs
HuijingHei: Good catch, thanks!

HuijingHei:
> It feels 40% less useful if we can't do transactional updates for shim

Is it something like https://github.com/coreos/bootupd/issues/440?

HuijingHei:
The error in CI is wired:
```
enabling baseos-source repository
enabling appstream-source repository
enabling extras-common-source repository
CentOS Stream 9 - BaseOS - Source               496  B/s | 3.0 kB     00:06    
Errors during downloading metadata for repository 'baseos-source':
  - Downloading successful, but checksum doesn't match. Calculated: e13bd479141ed656a70be1867b0c7258625f469b13baaf70011319723361358415da1f8417f92cf37eaf5dd4bb83dddae4ae7e52fddbc4c6dfb0ac3af2c3233a(sha512)  Expected: c435af47f3197ce3267e5244050b6b1ef1ee668c91cee288d83c887471d5c239b1c82cfec43c283624ba25d40b6eb01e56768a54bd7ffa4b47a5e21fb62365b3(sha512) 
Error: Failed to download metadata for repo 'baseos-source': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
Error: building at STEP "RUN dnf -y install dnf-utils zstd && dnf builddep -y rust-bootupd": while running runtime: exit status 1
```

travier:
> * It feels 40% less useful if we can't do transactional updates for shim (to rephrase: it's still useful, just less so)

Well, it's transactional, but not "across" folders. I find it unlikely that shim and grub would end up needing to be updated at the same time for them to properly work. I've asked input from the bootloader team but did not get any so far.

So far, it's the best that we've got without touching the EFI variables.

HuijingHei:
> > * It feels 40% less useful if we can't do transactional updates for shim (to rephrase: it's still useful, just less so)
> 
> Well, it's transactional, but not "across" folders. I find it unlikely that shim and grub would end up needing to be updated at the same time for them to properly work.

> So far, it's the best that we've got without touching the EFI variables.

With this patch, it is just to update files in another way (via temp dir). Before enable auto-updates, I think it should be more safer if include https://github.com/coreos/bootupd/issues/440

cgwalters:
Looks like this needs a rebase now btw

cgwalters:
> Anything else that I should do before merged? CC @cgwalters @travier , thanks!

I should emphasize again: thanks for working on this!

My feeling on this though is still very mixed not because of any flaw in your code, but because the structure of shim means we will get splits: https://github.com/coreos/bootupd/pull/669#discussion_r1659273299

There's also the fact that (again no fault of yours) this is just an inherently subtle problem domain, and while you did add a lot of new unit tests, OTOH, if we mess this up we break people's bootloaders.

cgwalters (approved):
Nevertheless, while I haven't yet truly audited the code, it looks overall sane to me and the newly added tests help, so if you feel confident I am OK moving ahead. Adding my approval.

HuijingHei:
Thanks @cgwalters and @travier for the grate help and a lot of instructions.
I am OK to merge this now to trigger more testing from `bootc` and find regressions earlier, WDYT? @travier

travier:
Thanks for the work here Huijing. Let's get this into Rawhide and test it there on full systems.

HuijingHei:
Much appreciated for the help from @travier and @cgwalters !

champtar:
Note: vfat RENAME_EXCHANGE support is not present in EL 9.5, will be in 9.6
https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/commit/0dd84c35238c594a53ac0e67b9793aa48cce34bc

PR #671, opened 2024-06-20
github/workflows: Update to ubuntu-24.04 & podman

travier:
See: https://github.com/ostreedev/ostree/pull/3270


travier:
```
=> Installing the firefox snap
```

OK, we'll have to remove some stuff before we do that :/

cgwalters:
See https://github.com/ostreedev/ostree/pull/3270

PR #673, opened 2024-06-28
ci: Add cross-arch builds

cgwalters:
Because we have a lot of arch conditionals that would be easy to break.


HuijingHei (approved):

HuijingHei:
LGTM, seems there is something blocked CI, is there anyway to re-trigger the testing?

Issue #449, opened 2023-03-21
port to cap-std


cgwalters:
openat is deprecated and less secure, and pulls in less maintained deps.

PR #674, opened 2024-06-28
Add cap-std, use in grubconfigs

cgwalters:
Starting on https://github.com/coreos/bootupd/issues/449

`openat` is pretty dead upstream, and especially going forward as we start to do more nontrivial file things, it's really useful to have cap-std's additional verification that we aren't accidentally escaping the root.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Rolv-Apneseth:
https://github.com/coreos/bootupd/pull/1107

PR #678, opened 2024-07-01
README: Update now that it is no longer a daemon

travier:
See: https://github.com/coreos/bootupd/pull/663

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,28 +118,25 @@ But out of conservatism currently today for e.g. Fedora CoreOS, bootupd is disab
 by default.  On the other hand, if your OS update mechanism isn't transactional,
 then you may want to enable bootupd by default.
 
-- Why is bootupd a daemon?
+- Is bootupd a daemon?
 
-It's not, really.  The name was intended to be "bootloader-upDater"
-not "bootloader-updater-Daemon"; the choice of a "d" suffix is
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,28 +118,25 @@ But out of conservatism currently today for e.g. Fedora CoreOS, bootupd is disab
...
-  DBus or Cap'n Proto or varlink or something else).  Having
-  a socket (without a defined stable API) is preparatory work for that.
+- Is bootupd a daemon?
+
+It's no longer a daemon. The name was intended to be "bootloader-upDater" not
+"bootloader-updater-Daemon". The choice of a "d" suffix is in retrospect
+probably too confusing.
+
+bootupd used to be a "daemon-like" application implemented as a
+socket-activated systemd service which spawns a `bootupd.service`. However -
```
cgwalters, line 128: How about:
> bootupd used to have an internally-facing `bootupd.service` and `bootupd.socket` that acted as a locking mechanism; the service would *very quickly* auto exit. There's nothing long-running, so it was not really a daemon.

cgwalters (changes requested):
Thanks for doing this.

travier:
Thanks, updated.

cgwalters (approved):

Issue #682, opened 2024-07-10
New bootupd release


HuijingHei:
Should create a new release according to https://bugzilla.redhat.com/show_bug.cgi?id=2268505#c27 to test for https://github.com/coreos/bootupd/pull/665.

PR #683, opened 2024-07-10
Release 0.2.20

HuijingHei:
Fixes https://github.com/coreos/bootupd/issues/682

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "bootupd"
-version = "0.2.19"
+version = "0.2.20"
 dependencies = [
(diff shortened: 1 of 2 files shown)


travier (approved):

cgwalters:
Thanks for this!

cgwalters:
One thing we probably need to do is set up something that basically at least does a PR for a release each week or so that can be merged; other teams do this.

HuijingHei:
> One thing we probably need to do is set up something that basically at least does a PR for a release each week or so that can be merged; other teams do this.

I think you mean using packit, we can enable it for bootupd (https://issues.redhat.com/browse/COS-2724)

PR #685, opened 2024-07-16
efi: sync bootentry label with Anaconda

HuijingHei:
Read `/etc/system-release` if exists, else get `NAME` from `/etc/os-release`

See https://bugzilla.redhat.com/show_bug.cgi?id=2268505#c24

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -139,15 +139,26 @@ impl Efi {
             log::debug!("Not booted via EFI, skipping firmware update");
             return Ok(());
         }
-        // Read /etc/os-release
-        let release: OsRelease = OsRelease::new()?;
-        let product_name: &str = &release.name;
+        let product_name = get_product_name()?;
         log::debug!("Get product name: {product_name}");
         assert!(product_name.len() > 0);
         // clear all the boot entries that match the target name
-        clear_efi_target(product_name)?;
-        create_efi_boot_entry(device, espdir, vendordir, product_name)
+        clear_efi_target(&product_name)?;
+        create_efi_boot_entry(device, espdir, vendordir, &product_name)
     }
+}
+
+#[context("Get product name")]
+fn get_product_name() -> Result<String> {
+    let file_path = Path::new("/etc/system-release");
+    if file_path.exists() {
+        let content = std::fs::read_to_string(file_path)?;
+        let re = regex::Regex::new(r" *release.*").unwrap();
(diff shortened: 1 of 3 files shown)


src/efi.rs
```diff
@@ -659,4 +670,10 @@ Boot0003* test";
         );
         Ok(())
     }
+    #[test]
+    fn test_get_product_name() -> Result<()> {
+        let name = get_product_name()?;
```
cgwalters, line 675: I think we shouldn't read global state in a unit test. This will fail on e.g. Debian or other systems and we want to support those.

We can unit test parsing from a `&str` easily enough though.

(That said of course, we could use openat or better cap-std to load from a test root...see what we do in https://github.com/containers/bootc/blob/c866bba1e96e912ab2e36b9eec3985e93e7a0c9b/lib/src/generator.rs#L90 - but unfortunately `OsRelease` doesn't support that yet)

cgwalters (changes requested):
Thanks! Looks sane, just one minor nit.

src/efi.rs
```diff
@@ -139,17 +139,28 @@ impl Efi {
...
         // clear all the boot entries that match the target name
-        clear_efi_target(product_name)?;
-        create_efi_boot_entry(device, espdir, vendordir, product_name)
+        clear_efi_target(&product_name)?;
+        create_efi_boot_entry(device, espdir, vendordir, &product_name)
     }
 }
 
+#[context("Get product name")]
+fn get_product_name() -> Result<String> {
```
HuijingHei, line 152: Should this change to `fn get_product_name(root: &Dir)` to pass root dir as parameter?

src/efi.rs
cgwalters: > This will fail on e.g. Debian or other systems and we want to support those.

I was wrong, it shouldn't because the `system-release` is optional.

However, it does still make sense to avoid reading global state longer term.

cgwalters (approved):

src/efi.rs
cgwalters: Yeah, but not a blocker, just a nice to have IMO.

src/efi.rs
HuijingHei: Thanks! Yes, you are correct, will switch to `os-release` if not exist `system-release`.

src/efi.rs
HuijingHei: Thanks, will update in the later PR.

PR #686, opened 2024-07-17
efi: avoid reading global state for `get_product_name()`

HuijingHei:
Inspired by https://github.com/coreos/bootupd/pull/685#discussion_r1679742803

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -10,6 +10,8 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use anyhow::{bail, Context, Result};
+use cap_std::fs::Dir;
+use cap_std_ext::cap_std;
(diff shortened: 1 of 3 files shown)


src/efi.rs
```diff
@@ -139,7 +141,8 @@ impl Efi {
             log::debug!("Not booted via EFI, skipping firmware update");
             return Ok(());
         }
-        let product_name = get_product_name()?;
+        let sysroot = Dir::open_ambient_dir(&Path::new("/"), cap_std::ambient_authority())?;
```
cgwalters, line 144: Minor nit, you can just pass `"/"` here, because the API accepts `impl AsRef<Path>`.

cgwalters (approved):
Nice, thanks!

Issue #707, opened 2024-08-22
Graceful recover after a failure


travier:
From https://github.com/coreos/bootupd/issues/694#issuecomment-2256146956:

If for whatever reason the previous command fails, then I need to `systemctl reset-failed` before calling bootupd again.

```
root@fedora:~# systemd-run --unit bootupd --property=MountFlags=slave --service-type=oneshot --pipe bootupctl status
Running as unit: bootupd.service
root@fedora:~# systemd-run --unit bootupd --property=MountFlags=slave --service-type=oneshot --pty bootupctl status
Running as unit: bootupd.service; invocation ID: c111d0e1cebe4401bb89af4cfa5358f1
Press ^] three times within 1s to disconnect TTY.

root@fedora:~# systemd-run --unit bootupd --property=MountFlags=slave --service-type=oneshot --pty bootupctl status
Failed to start transient service unit: Unit bootupd.service was already loaded or has a fragment file.
root@fedora:~# systemctl status bootupd.service
× bootupd.service - /usr/bin/bootupctl status
     Loaded: loaded (/run/systemd/transient/bootupd.service; transient)
  Transient: yes
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: failed (Result: signal) since Mon 2024-07-29 16:46:17 CEST; 14s ago
 Invocation: c111d0e1cebe4401bb89af4cfa5358f1
    Process: 4624 ExecStart=/usr/bin/bootupctl status (code=killed, signal=HUP)
   Main PID: 4624 (code=killed, signal=HUP)
   Mem peak: 3M
        CPU: 32ms

Jul 29 16:46:17 fedora systemd[1]: Starting bootupd.service - /usr/bin/bootupctl status...
Jul 29 16:46:17 fedora systemd[1]: bootupd.service: Main process exited, code=killed, status=1/HUP
Jul 29 16:46:17 fedora systemd[1]: bootupd.service: Failed with result 'signal'.
Jul 29 16:46:17 fedora systemd[1]: Failed to start bootupd.service - /usr/bin/bootupctl status.
root@fedora:~# systemd-run --unit bootupd --property=MountFlags=slave --service-type=oneshot --pty bootupctl status
Failed to start transient service unit: Unit bootupd.service was already loaded or has a fragment file.
```

travier:
A quick & dirty way to avoid that is issue is to always call `systemctl reset-failed bootupd.service` before trying to run another command.

[...]

PR #715, opened 2024-09-02
bootupctl: Clear failure status from previous runs

travier:
If for whatever reason a bootupd command fails, it will leave the
systemd service unit in a failed state and systemd will then refuse to
run a unit under the same name with `systemd-run` again until the
failure is cleared.

Thus systematically call `systemctl reset-failed` before calling
`systemd-run` to clear any potential failures from previous calls.

Do not check the return code of the systemctl command on purpose as it
may fail if the unit does not exists yet, i.e. if no bootupctl command
has been run yet.

Also ignore stdout/stderr to avoid showing unexpected errors messages to
users.

See: https://github.com/coreos/bootupd/issues/707
See: https://github.com/coreos/bootupd/pull/663

### DIFF
--- a/src/cli/bootupctl.rs
+++ b/src/cli/bootupctl.rs
@@ -4,7 +4,7 @@ use clap::Parser;
 use log::LevelFilter;
 
 use std::os::unix::process::CommandExt;
-use std::process::Command;
+use std::process::{Command, Stdio};
 
 static SYSTEMD_ARGS_BOOTUPD: &[&str] = &[
     "--unit",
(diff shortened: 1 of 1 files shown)


travier:
This "explicitly" does not check the return code of the `systemctl command`as it may fail if the unit does not exists yet, i.e. if no bootupctl command has been run yet.

cgwalters:
> Maybe we should also completely ignore stdout/stderr for this command.

This also sounds like a good idea.

cgwalters (approved):

Issue #432, opened 2023-02-22
consider folding into github.com/containers/bootc


cgwalters:
This project was designed to be distribution independent; that's in theory.  In practice today it is tied to both ostree and rpm (though the rpm part is definitely written in an attempted-to-be pluggable way).

But I think it will make more sense to fold this into https://github.com/containers/bootc - we're not going to have non-ostree support anytime soon, and if we do it should work in the same way as bootc.  Also, I think we want `bootc bootloader update` e.g.

Also, the two projects are similarly named and it will be less confusing to just have bootc, not bootc and bootupd.

dbnicholson:
FWIW, I'm hoping to start using bootupd in Endless OS and I doubt we're going to move to OCI payloads any time in the near future. Not to say that you can't do this or there's not value in unifying there. We'd probably just fork the existing code and carry on with it in that case.

cgwalters:
> and I doubt we're going to move to OCI payloads any time in the near future. 

BTW, I'd love to chat about this with you at some point!  Or maybe you can just write up at least briefly here why not or what the blockers/problems you see are?

jmarrero:
I have no issue with this, the only hangup up I could see is how we do messaging about bootc being a doing one and one thing well (deployments).

I guess the bootloader is part of the deployment so.

`bootc manages your deployments including the bootloader, using the container as the source of truth`  maybe is enough.

cgwalters:
@dbnicholson what if we did it in a way where one can build bootupd out of bootc git, while keeping all of the existing binaries/etc?

dbnicholson:
Unfortunately, my project to use bootupd stalled (even though it appeared to work nicely), and I can't see us getting back it soon. I do think this is a nice little project that has nothing to do with containers or OCI storage, but if it's causing you friction to have it separate it won't affect us right now to subsume it. It would be nice to continue to be able to build `bootupd` as a standalone utility, but I wouldn't spend a lot of time on it.

cgwalters:
I put up a draft for this at https://github.com/containers/bootc/pull/874

Conan-Kudo:
FWIW, I had been considering looking into integrating support for this into mainline Fedora variants (and had discussed this with @travier about doing so sometime in the next year or so), but if it merges inside of bootc, I don't think I'd continue pursuing that line. It also wouldn't be useful for me to consider advocating for usage with Btrfs based systems, where this could be useful with too, as the bootc part is irrelevant for that.

[...]

PR #716, opened 2024-09-03
packaging: add a systemd unit to run every boot

jbtrystram:
This enables bootloader updates automatically on boot. Note that the service is intentionally not enabled by default, it should be up to the distribution to add a systemd preset if auto-update for the bootloader is desired.

Right now RAID setups are not supported but see [1] for an example in coreos.

[1] https://github.com/coreos/fedora-coreos-config/pull/3042

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,9 @@ install-grub-static:
 	install -m 644 -D -t ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static src/grub2/*.cfg
 	install -m 755 -d ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static/configs.d
 
+install-systemd-unit:
+	install -m 644 -D -t "${DESTDIR}$(PREFIX)/lib/systemd/system/" contrib/packaging/bootupctl-update.service
+
 bin-archive:
 	rm target/inst -rf
 	$(MAKE) install install-grub-static DESTDIR=$$(pwd)/target/inst
--- a/contrib/packaging/bootupctl-update.service
+++ b/contrib/packaging/bootupctl-update.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Update Bootloader on boot
+Documentation=https://github.com/coreos/bootupd
+ConditionPathExists=/dev/disk/by-label/EFI-SYSTEM
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/bootupctl update
+RemainAfterExit=yes
+MountFlags=slave
+
+[Install]
+WantedBy=multi-user.target
--- a/contrib/packaging/bootupd.spec
+++ b/contrib/packaging/bootupd.spec
@@ -21,6 +21,7 @@ BuildRequires: cargo
 BuildRequires: git
 BuildRequires: openssl-devel
 BuildRequires: systemd-devel
+BuildRequires: systemd-rpm-macros
 
 %description 
 %{summary}
@@ -31,6 +32,7 @@ BuildRequires: systemd-devel
 %{_bindir}/bootupctl
 %{_libexecdir}/bootupd
 %{_prefix}/lib/bootupd/grub2-static/
+%{_unitdir}/bootupctl-update.service
 
 %prep
 %autosetup -n %{crate}-%{version} -p1 -Sgit
@@ -50,6 +52,7 @@ cargo build --release
 %install
 %make_install INSTALL="install -p -c"
 make install-grub-static DESTDIR=%{?buildroot} INSTALL="%{__install} -p"
+make install-systemd-unit DESTDIR=%{?buildroot} INSTALL="%{__install} -p"
 
 %changelog
 * Tue Oct 18 2022 Colin Walters <walters@verbum.org> - 0.2.8-3


travier:
I'll do some manual testing.

jbtrystram:
I built a RPM and tested in a fcos build.
After enabling the service and rebooting : 

```
[core@cosa-devsh ~]$ systemctl status bootupctl-update.service 
● bootupctl-update.service - Update Bootloader on boot
     Loaded: loaded (/usr/lib/systemd/system/bootupctl-update.service; enabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: active (exited) since Tue 2024-09-03 15:15:56 UTC; 23s ago
       Docs: https://github.com/coreos/bootupd
    Process: 1099 ExecStart=/usr/bin/bootupctl update (code=exited, status=0/SUCCESS)
   Main PID: 1099 (code=exited, status=0/SUCCESS)
        CPU: 9ms

Sep 03 15:15:56 localhost systemd[1]: Starting bootupctl-update.service - Update Bootloader on boot...
Sep 03 15:15:56 localhost bootupctl[1099]: No update available for any component.
Sep 03 15:15:56 localhost systemd[1]: Finished bootupctl-update.service - Update Bootloader on boot.
```

contrib/packaging/bootupctl-update.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Update Bootloader on boot
+Documentation=https://github.com/coreos/bootupd
+ConditionPathExists=/dev/disk/by-label/EFI-SYSTEM
```
travier, line 4: We should add as well:
```
/dev/disk/by-partlabel/EFI\\x20System\\x20Partition
```

travier, line 4:
```suggestion
ConditionPathExists=/dev/disk/by-partlabel/EFI-SYSTEM
```

As that's what's used in https://github.com/coreos/bootupd/blob/main/src/efi.rs#L39

travier:
needs a `|` in the condition: https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Conditions%20and%20Asserts

travier:
With those changes, that works for me on Silverblue

contrib/packaging/bootupctl-update.service
cgwalters: There's already also `ConditionFirmware=uefi`

cgwalters: That said we need to support BIOS too...so probably the best would be e.g.

```
RequiresMountsFor=/boot
ConditionPathExists=/boot/bootupd-state.json
```

cgwalters:
I'm not opposed to this but the reason I didn't do this until now is basically I think what we really want is more like https://github.com/coreos/bootupd/issues/108 which means the thing doing the OS updates drives bootupd itself by default.

contrib/packaging/bootupctl-update.service
jbtrystram: `ConditionPathExists=/boot/bootupd-state.json` 
wouldn't that prevent it to run on atomic desktops as bootupd was not enabled until now ?

Makefile
```diff
@@ -41,6 +41,9 @@ install-grub-static:
 	install -m 644 -D -t ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static src/grub2/*.cfg
 	install -m 755 -d ${DESTDIR}$(PREFIX)/lib/bootupd/grub2-static/configs.d
 
+install-systemd-unit:
+	install -m 644 -D -t "${DESTDIR}$(PREFIX)/lib/systemd/system/" contrib/packaging/bootupctl-update.service
```
cgwalters, line 45: An overall challenge here is that we'll have come full circle since https://github.com/coreos/bootupd/pull/663 ...and then e.g. people maintaining SELinux policy are still going to come by and say "oh hey there's a systemd service, we need to write a policy for it...".

With this combined I would probably say that instead of this we should implement bootupd support into rpm-ostree and bootc.

(This also relates to https://github.com/coreos/bootupd/issues/432 )

cgwalters (changes requested):
Marking as requested-changes per https://github.com/coreos/bootupd/pull/716#discussion_r1742642664 but not a *hard* blocker for me, if you or someone wants to do this in the interim I'm OK, but it does basically clash and I think longer term it'll be more understandable for users; we'd have to maintain this newly introduced service unit here ~forever.

travier:
I would really like to get this into Atomic Desktops for F41 as we desperately need it there. So I can carry it there in the mean time until we get a cleaner solution here.

Agree that https://github.com/coreos/bootupd/issues/108 feels like an improvement but I don't think it should hold us from doing this one in the mean time?

cgwalters:
> I would really like to get this into Atomic Desktops for F41 as we desperately need it there. 

Isn't what is desperately needed more something like automatic blocking of OS updates when the bootloader state is too old? I think we could turn around an implementation of that in a few days.

cgwalters:
One thing this is also basically sweeping under the rug of an on-by-default systemd unit is the rationale stated in https://github.com/coreos/bootupd?tab=readme-ov-file#questions-and-answers from the very start of this project. Yes the problem stinks, and [the rename_exchage](https://github.com/coreos/bootupd/pull/669) work *probably* makes it better, but still...the semantics here are going to be that during the middle of OS bootup we do this (potentially) unsafe operation with zero warning about that unsafety.

cgwalters:
> fwupd already has a mechanism to hold updates if the system is not on AC or doesn't have enough battery so that could be a good place to look at as well as that is already integrated in Plasma Discover & GNOME Software.

I like this idea, it makes sense to me to encourage grouping of these two things, even if technically they are distinct.

cgwalters:
Anyways to your larger point...fair. I am OK to merge but let's at least get https://github.com/coreos/bootupd/pull/716/files#r1742591590 changed?

contrib/packaging/bootupctl-update.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
travier, line 1: Let's name this unit `bootloader-update.service` as its what will be user visible and users ideally should not have to know about bootupd.

contrib/packaging/bootupctl-update.service
```diff
@@ -0,0 +1,13 @@
...
+Description=Update Bootloader on boot
```
travier, line 2:
```suggestion
Description=Update bootloader on boot
```

travier:
> Anyways to your larger point...fair. I am OK to merge but let's at least get [#716 (files)](https://github.com/coreos/bootupd/pull/716/files#r1742591590) changed?

I think we should keep this check as broad as possible. We can not require `ConditionPathExists=/boot/bootupd-state.json` as we explicitly want to also adopt and update systems that are not yet using bootupd.
I'm installing a BIOS system again to check the partitions.

travier:
I can't find a way to select BIOS systems in https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#ConditionFirmware=.

travier:
OK, it looks like this requires different trade-offs and conditions check depending on which system is installed right now so it might be better downstream. I've made: https://pagure.io/workstation-ostree-config/pull-request/571

cgwalters:
I'm OK with that too, though the tradeoff is it's just going to get copy-pasted again by IoT...

jlebon:
>> Isn't what is desperately needed more something like automatic blocking of OS updates when the bootloader state is too old? I think we could turn around an implementation of that in a few days.
>
> If we did that, we would have to surface that somehow to the user in the interface.

Hmm, I think if we did that in rpm-ostree in a way that literally prevented `Deploy()` from working, that should surface, no? (As an rpm-ostree error?) Obviously not pretty, but wouldn't require any changes to g-s/KDE.

jlebon:
Basically, something like:

```
# rpm-ostreed.conf
[Daemon]
RequireUpdatedBootloader=true
```

When true (default is false), rpm-ostree does `bootupctl status --json`, checks for `updatable` if it's `at-latest-version`, and if not, error out with something like:

> error: RequireUpdatedBootloader is enabled and bootloader is out of date. Please run `bootupctl update`

?

jlebon:
> Hmm, I think if we did that in rpm-ostree in a way that literally prevented `Deploy()` from working, that should surface, no? (As an rpm-ostree error?)

Chatting with @travier, it doesn't seem to be the case. See e.g. https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2000.

Obviously, there's nothing that prevents us from adding it now, and then improving it during the f41 cycle.

travier:
Looks like we never implemented [adoption for BIOS](https://github.com/coreos/bootupd/issues/717) systems so we will have to make this EFI only 😞

contrib/packaging/bootupctl-update.service
HuijingHei: I am OK with `ConditionFirmware=uefi`, can remove it until we add `adopt-and-update` for BIOS later.
One minor suggestion, should we add `ConditionKernelCommandLine=ostree` to skip this for live boots like https://github.com/coreos/rpm-ostree/pull/4944 ?

jbtrystram: i think the overall agreement was to close this PR and carry this in the distribution

jbtrystram:
I feel the general agreement here is that this would be better carried in the distribution side. I'll close this

HuijingHei:
One thing I am concern is if booting with `BIOS`, is it necessary to also update `UEFI` (on x86_64)? Can we update the current boot firmware, like `install`, eg https://github.com/coreos/bootupd/commit/5ee0483a64d5dcadc8c94b49951f2df6146dbb50

But first thing, will add adoption for BIOS.

Issue #717, opened 2024-09-05
Adoption failure on Atomic Desktops using BIOS


travier:
Trying to adopt / update an Atomic Desktop using BIOS boot fails:

```
$ bootupctl adopt-and-update -vvvv
[TRACE bootupd] executing cli
Running as unit: bootupd.service
[TRACE bootupd] executing cli
[TRACE bootupd::bootupd] No saved state
[TRACE bootupd::bootupd] Remaining known components: 2
[TRACE bootupd::bootupd] Not adoptable: BIOS
[DEBUG bootupd::efi] Unmounting
error: Failed to find ESP device
```

This is on a Fedora Silverblue 41 system, installed from F40, with the latest SELinux policy fixes & bootupd code:

```
root@fedora:~# rpm-ostree status
State: idle
Deployments:
● fedora:fedora/41/x86_64/silverblue
                  Version: 41.20240903.n.0 (2024-09-03T08:13:22Z)
               BaseCommit: 9c5a385934b5948c1f4bb23a441c7bcd85ed250455d75fbb06a3318c35de1b77
             GPGSignature: Valid signature by 466CF2D8B60BC3057AA9453ED0622462E99D6AD1
           LocalOverrides: bootupd 0.2.20-2.fc41 -> 202409030125.0.2.21.4.g0f3de09-1.fc41
                           selinux-policy selinux-policy-targeted 41.14-1.fc41 -> 41.15-1.20240904174231942019.pr2344.8.g9131c0b7b.fc41

  fedora:fedora/41/x86_64/silverblue
                  Version: 41.20240903.n.0 (2024-09-03T08:13:22Z)
                   Commit: 9c5a385934b5948c1f4bb23a441c7bcd85ed250455d75fbb06a3318c35de1b77
             GPGSignature: Valid signature by 466CF2D8B60BC3057AA9453ED0622462E99D6AD1

  fedora:fedora/40/x86_64/silverblue
                  Version: 40.20240904.0 (2024-09-04T02:12:02Z)
                   Commit: d516f5b47a48e18e2317f9f038656363384b0b880bf9d48cd9686991bb3a8a2d
             GPGSignature: Valid signature by 115DF9AEF857853EE8445D0A0727707EA15B79CC
                   Pinned: yes
```

travier:
[...]

PR #723, opened 2024-09-12
bios.rs: add `adopt-and-update`

HuijingHei:
Fixes https://github.com/coreos/bootupd/issues/717

### DIFF
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -122,16 +122,22 @@ impl Component for Bios {
     }
 
     fn query_adopt(&self) -> Result<Option<Adoptable>> {
-        Ok(None)
+        crate::component::query_adopt_state()
     }
 
-    #[allow(unused_variables)]
-    fn adopt_update(
-        &self,
-        sysroot: &openat::Dir,
-        update: &ContentMetadata,
-    ) -> Result<InstalledContent> {
-        todo!();
+    fn adopt_update(&self, _: &openat::Dir, update: &ContentMetadata) -> Result<InstalledContent> {
+        let Some(meta) = self.query_adopt()? else {
+            anyhow::bail!("Failed to find adoptable system")
+        };
+
+        let device = self.get_device()?;
+        let device = device.trim();
+        self.run_grub_install("/", device)?;
+        Ok(InstalledContent {
+            meta: update.clone(),
+            filetree: None,
+            adopted_from: Some(meta.version),
+        })
     }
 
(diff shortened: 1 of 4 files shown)


HuijingHei:
Hold on, need to deduplicate the code first.

src/bios.rs
```diff
@@ -122,16 +122,24 @@ impl Component for Bios {
...
-        sysroot: &openat::Dir,
-        update: &ContentMetadata,
-    ) -> Result<InstalledContent> {
-        todo!();
+    fn adopt_update(&self, _: &openat::Dir, update: &ContentMetadata) -> Result<InstalledContent> {
+        let meta = if let Some(meta) = self.query_adopt()? {
+            meta
+        } else {
+            anyhow::bail!("Failed to find adoptable system");
+        };
```
cgwalters, line 133: Totally fine as is, we have a lot of this in the code but in newer Rust one can do instead:

`let Some(meta) = self.query_adopt()? else { anyhow::bail!("Failed to find adoptable system") };`
which is shorter. See e.g. https://github.com/containers/bootc/pull/650/files

src/component.rs
```diff
@@ -144,6 +144,38 @@ pub(crate) fn get_component_update(
     }
 }
 
+#[context("query adopt impl")]
```
cgwalters, line 147: The intention for `context` is to provide something friendly for an external user. Maybe `Querying adoptable state`?

src/component.rs
```diff
@@ -144,6 +144,38 @@ pub(crate) fn get_component_update(
...
+pub(crate) fn query_adopt_impl() -> Result<Option<Adoptable>> {
+    // This would be extended with support for other operating systems later
+    if let Some(coreos_aleph) = crate::coreos::get_aleph_version(Path::new("/"))? {
+        let meta = ContentMetadata {
+            timestamp: coreos_aleph.ts,
+            version: coreos_aleph.aleph.version,
+        };
+        log::trace!("BIOS adoptable: {:?}", &meta);
```
cgwalters, line 155: I think we can just remove "BIOS" here

cgwalters (approved):
Overall looks sane to me, just a few nits.
Thanks for working on this!

src/bios.rs
HuijingHei: Will update this in later patch, thanks!

cgwalters (approved):

PR #729, opened 2024-09-20
Skip `UEFI` or `BIOS` on none hybrid partition layout

HuijingHei:
- When booting with `BIOS` and no `ESP` device detected, skip `UEFI` checking
- When booting with `UEFI` and no `bios_boot` partition found, skip `BIOS` checking

Fixes https://github.com/coreos/bootupd/issues/717

### DIFF
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -8,9 +8,22 @@ use crate::packagesystem;
 use anyhow::{bail, Result};
 
 use crate::util;
+use serde::{Deserialize, Serialize};
 
 // grub2-install file path
 pub(crate) const GRUB_BIN: &str = "usr/sbin/grub2-install";
+
+#[derive(Serialize, Deserialize, Debug)]
+struct BlockDevice {
+    path: String,
+    pttype: String,
+    parttypename: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct Devices {
+    blockdevices: Vec<BlockDevice>,
+}
 
 #[derive(Default)]
 pub(crate) struct Bios {}
@@ -22,12 +35,11 @@ impl Bios {
         #[cfg(target_arch = "x86_64")]
         {
             // find /boot partition
-            let boot_dir = Path::new("/").join("boot");
             cmd = Command::new("findmnt");
             cmd.arg("--noheadings")
                 .arg("--output")
                 .arg("SOURCE")
-                .arg(boot_dir);
+                .arg("/boot");
             let partition = util::cmd_output(&mut cmd)?;
 
             // lsblk to find parent device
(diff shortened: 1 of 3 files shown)


HuijingHei:
The CI will not cover related testing, do some testing manually on none hybrid partition layout:


on BIOS

```
root@fedora:/home/fedora# bootupctl status
Component BIOS
  Installed: grub2-tools-1:2.12-4.fc41.x86_64
  Update: At latest version
No components are adoptable.
Boot method: BIOS

root@fedora:/home/fedora# bootupctl validate
Running as unit: bootupd.service; invocation ID: 8ec29c38bb534882b7f509ea10be9cf9
Skipped: BIOS

root@fedora:/home/fedora# rm -f /boot/bootupd-state.json 
root@fedora:/home/fedora# bootupctl update
Adopted and updated: BIOS: grub2-tools-1:2.12-4.fc41.x86_64
```

on UEFI

```
root@fedora:/home/fedora# bootupctl status
Component EFI
  Installed: grub2-efi-ia32-1:2.12-4.fc41.x86_64,grub2-efi-x64-1:2.12-4.fc41.x86_64,shim-ia32-15.8-3.x86_64,shim-x64-15.8-3.x86_64
  Update: At latest version
No components are adoptable.
Boot method: EFI

root@fedora:/home/fedora# bootupctl validate
Running as unit: bootupd.service; invocation ID: a98fa7e453064291a16594dc83a15c9b
Validated: EFI

root@fedora:/home/fedora# rm -f /boot/bootupd-state.json
root@fedora:/home/fedora# bootupctl update
Adopted and updated: EFI: grub2-efi-ia32-1:2.12-4.fc41.x86_64,grub2-efi-x64-1:2.12-4.fc41.x86_64,shim-ia32-15.8-3.x86_64,shim-x64-15.8-3.x86_64
```

src/bios.rs
```diff
@@ -81,6 +93,38 @@ impl Bios {
         }
         Ok(())
     }
+
+    // check bios_boot partition on gpt type disk
+    fn get_bios_boot_partition(&self) -> Result<Option<String>> {
+        let target = self.get_device()?;
+        // lsblk to list children with bios_boot
+        let output = Command::new("lsblk")
```
cgwalters, line 101: Side note if we fold into bootc, we already have the equivalent of this code there https://github.com/containers/bootc/blob/a9cfdc83cc747ac97a6f396735e8d6cb57c68e30/lib/src/blockdev.rs#L281

cgwalters (approved):
This looks sane to me...I assume you tested it on the target env?

src/bios.rs
HuijingHei: Good insight, hmm, bootupd can not use that code, need to find workaround for this.

cgwalters: I'm thinking recently again about https://github.com/coreos/bootupd/issues/432 which would fix this.

Anyways, it was just an aside.

In the short term maybe we can try to literally copy in `blockdev.rs` from bootc or so.

HuijingHei: Maybe can try with `/dev/disk/by-partlabel/BIOS-BOOT`, but this does not exist on silverblue (start using BIOS):
```
$ sudo blkid
/dev/vda1: PARTUUID="928fc854-1795-4856-b9d7-7818dff9bfbc"
$ ls /dev/disk/by-partlabel/
ls: cannot access '/dev/disk/by-partlabel/': No such file or directory
```

HuijingHei: > I'm thinking recently again about https://github.com/coreos/bootupd/issues/432 which would fix this.

No objection, maybe can add the check when call bootupd.

> In the short term maybe we can try to literally copy in blockdev.rs from bootc or so.

Will try this, thanks for the pointer.

HuijingHei:
> This looks sane to me...I assume you tested it on the target env?

Yes, see https://github.com/coreos/bootupd/pull/729#issuecomment-2363008895, do some testings on silverblue using BIOS or using UEFI, both work. Will update this according to your suggestion and do testing again.

cgwalters:
> Will update this according to your suggestion

I think this can be a followup, no need to block on it.

HuijingHei:
> > Will update this according to your suggestion
> 
> I think this can be a followup, no need to block on it.

SGTM, thanks for the review and pointer!

PR #730, opened 2024-09-20
Convert to `let-else` syntax

HuijingHei:
Convert to [let else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) where possible.

Inspired by https://github.com/coreos/bootupd/pull/723#discussion_r1757480310


cgwalters (approved):
Great, thanks!

PR #734, opened 2024-09-25
Release 0.2.23

HuijingHei:
See https://github.com/coreos/bootupd/issues/733


HuijingHei:
The selinux issue is tracked by https://github.com/fedora-selinux/selinux-policy/issues/2362

Issue #739, opened 2024-10-01
Bootupd: thread 'main' panicked at src/bios.rs:116:62


TheGreatDeadOne:
**Describe the bug**
After updating bootupd, this bug appeared when trying to check the status of bootupd.

**To Reproduce**
Please describe the steps needed to reproduce the bug:
1. Update the system
2. Reboot
3. view bootupd status (or update)

**Screenshots**

![Captura de tela de 2024-10-01 04-30-57](https://github.com/user-attachments/assets/9c6132db-e52b-40df-9f3a-3c795bba1d20)

**OS version:**
```
State: idle
BootedDeployment:
● fedora:fedora/rawhide/x86_64/silverblue
                  Version: Rawhide.20240930.n.0 (2024-09-30T06:07:39Z)
                   Commit: d44afe119fb862868666a1829781e2a1888497d4c0189186e290289cc35c369d
             GPGSignature: Valid signature by B0F4950458F69E1150C6C5EDC8AC4916105EF944
```

**Additional context**
I ran it with RUST_BACKTRACE=full, but I didn't get any more information.

/boot/bootupd-state.json content:

```
{
  "installed": {
    "EFI": {
      "meta": {
        "timestamp": "2024-08-08T12:14:11Z",
        "version": "grub2-efi-ia32-1:2.12-4.fc41.x86_64,grub2-efi-x64-1:2.12-4.fc41.x86_64,shim-ia32-15.8-3.x86_64,shim-x64-15.8-3.x86_64"
      },
      "filetree": {
        "children": {
          "BOOT/BOOTIA32.EFI": {
            "size": 747681,
            "sha512": "sha512:6d4396f289400516b883733f0fd3621e7ec4d70afd02e988651f37db81298775da69c04b998d87a4760b2fa4b96130c70eb0875fad1cf290c52ea606ae40d12e"
          },
          "BOOT/BOOTX64.EFI": {
            "size": 949424,
            "sha512": "sha512:cc23d8c3cb2dcf749075268b77eb796fb430182cbbc04171ded14d43e32b4a5cdeeb1a08666ee0e288bd37d63f657a9af5e7f2012dd70694d11212d705c60b42"
          },
          "BOOT/fbia32.efi": {
            "size": 70360,
            "sha512": "sha512:c3e94b8d6f07ec9c88b66d48e5b887d29288dd1c9914bf8226af7656e35a023c07a6dbf8a2875994485c348ca358591e02a8eb19574299db3ffb6486697a1568"
[...]

PR #740, opened 2024-10-02
bios: Handle empty pttype from lsblk output

travier:
zram, sr0 (CD/DVD) and LUKS devices generally don't use partitions, thus don't have a partition table and thus don't have a partition table type so this field may be null.

Fixes: https://github.com/coreos/bootupd/issues/739

### DIFF
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -16,7 +16,7 @@ pub(crate) const GRUB_BIN: &str = "usr/sbin/grub2-install";
 #[derive(Serialize, Deserialize, Debug)]
 struct BlockDevice {
     path: String,
-    pttype: String,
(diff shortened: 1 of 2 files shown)


travier:
(Not tested yet)

cgwalters:
I pushed a change to fix compilation - would obviously be good to test in the failing scenario for real, but we could also add a unit test fixture JSON.

travier:
Thanks. I was setting up VMs to test this up but got distracted by something else. Will hopefully test tomorrow.

travier:
I've added a small unit test and manually verified that this works on Fedora Silverblue Rawhide BIOS & UEFI.

jmarrero (approved):
lgtm

PR #759, opened 2024-10-29
Disable tests that require creation time on musl

JustSoup312:
Creation time will only be supported on the musl target once https://github.com/rust-lang/rust/pull/125692 lands.

This MR disables the tests that require creation time for musl.


cgwalters (approved):
Hmm, we could probably use mtime here too in practice.

travier (approved):

Issue #762, opened 2024-11-04
Adopt fails with error "removing ".btmp.fedora/mmia32.efi": No such file or directory (os error 2)"


jorgeml:
On a X86-84 system running Fedora Silverblue 41:

```
$ sudo bootupctl status
Running as unit: bootupd.service
No components installed.
Detected: EFI: unknown
Boot method: EFI

$ sudo bootupctl adopt-and-update
Running as unit: bootupd.service
error: Failed adopt and update: applying filesystem changes: removing ".btmp.fedora/mmia32.efi": No such file or directory (os error 2)

$ sudo bootupctl --version
bootupctl 0.2.24

```

HuijingHei:
Thanks for the reports.

Does this mean you have upgraded the OS from older version to Fedora Silverblue 41 latest?

Which version are you using for Fedora Silverblue 41? As I remembered, the latest version includes [bootloader-update.service](https://pagure.io/workstation-ostree-config/c/2181f67a2af8b1d34629151fe8b09ae834ebf7c4?branch=f41), which should automatically update bootloader during booting after upgrade, so you might do not need to update manually.

jorgeml:
Thanks @HuijingHei ! Yes, this was an upgraded system. I have been running Silverblue for a few releases.

I updated to the latest Silverblue compose (41.20241105.0 (2024-11-05T00:40:00Z)) and the error has changed:

```
$ sudo bootupctl adopt-and-update
Running as unit: bootupd.service
error: Failed adopt and update: applying filesystem changes: removing ".btmp.BOOT/fbia32.efi": No such file or directory (os error 2)

$ sudo bootupctl validate
Running as unit: bootupd.service
No components installed.
```

HuijingHei:
I can not reproduce when I upgrade to 41.20241105.0, maybe there is something different with your env, could you help to output your `lsblk` result?

```
fedora@fedora:~$ journalctl -b -u bootloader-update.service --no-page
Nov 06 01:32:38 fedora systemd[1]: Starting bootloader-update.service - Update bootloader on boot...
Nov 06 01:32:38 fedora bootupctl[920]: Adopted and updated: EFI: grub2-efi-ia32-1:2.12-10.fc41.x86_64,grub2-efi-x64-1:2.12-10.fc41.x86_64,shim-ia32-15.8-3.x86_64,shim-x64-15.8-3.x86_64
[...]

PR #765, opened 2024-11-07
filetree: skip the error if removing unexisting files

HuijingHei:
Fixes https://github.com/coreos/bootupd/issues/762

### DIFF
--- a/src/filetree.rs
+++ b/src/filetree.rs
@@ -348,7 +348,7 @@ pub(crate) fn apply_diff(
                 path_tmp = path.to_path_buf();
             }
             destdir
-                .remove_file(path_tmp.as_std_path())
+                .remove_file_optional(path_tmp.as_std_path())
                 .with_context(|| format!("removing {:?}", path_tmp))?;
         }
     }
@@ -691,12 +691,19 @@ mod tests {
             assert_eq!(b_btime_foo_new, b_btime_foo);
         }
         {
-            a.remove_file(testfile)?;
+            b.remove_file(testfile)?;
+            let ta = FileTree::new_from_dir(&a)?;
+            let diff = ta.relative_diff_to(&b)?;
+            assert_eq!(diff.removals.len(), 1);
+            apply_diff(&a, &b, &diff, None).context("test removed files with relative_diff")?;
+            assert_eq!(b.exists(testfile)?, false);
+        }
+        {
             a.remove_file(bar)?;
             let diff = run_diff(&b, &a)?;
             assert_eq!(diff.count(), 2);
             apply_diff(&a, &b, &diff, None).context("test removed files")?;
-            assert_eq!(b.exists(testfile)?, false);
+            assert_eq!(b.exists(testfile)?, true);
             assert_eq!(b.exists(bar)?, false);
             let diff = run_diff(&b, &a)?;
             assert_eq!(diff.count(), 0);
--- a/tests/e2e-update/e2e-update-in-vm.sh
+++ b/tests/e2e-update/e2e-update-in-vm.sh
@@ -89,7 +89,19 @@ rm -f /boot/bootupd-state.json
 bootupctl adopt-and-update | tee out.txt
 assert_file_has_content out.txt "Adopted and updated: BIOS: .*"
 assert_file_has_content out.txt "Adopted and updated: EFI: .*"
+bootupctl validate
 ok adopt-and-update
+
+# Verify the adoption does not fail when install files if they are missing on the disk.
+# see https://github.com/coreos/bootupd/issues/762
+rm -f /boot/bootupd-state.json
+[ -f "${tmpefimount}/EFI/fedora/test-bootupd.efi" ] && rm -f ${tmpefimount}/EFI/fedora/test-bootupd.efi
+bootupctl adopt-and-update | tee out.txt
+assert_file_has_content out.txt "Adopted and updated: BIOS: .*"
+assert_file_has_content out.txt "Adopted and updated: EFI: .*"
+if bootupctl validate 2>err.txt; then
+  fatal "unexpectedly passed validation"
+fi
 
 tap_finish
 touch /run/testtmp/success


HuijingHei:
The `bootupctl validate` failed is because there are some files ( from `shim-ia32` and `grub2-efi-ia32` packages) removed  in `/boot/efi/EFI` compared to `/usr/lib/bootupd/updates/EFI/`, refer to https://github.com/coreos/bootupd/blob/main/src/efi.rs#L413-L420. 

Do we still need the `ia32` files on x64 platform? Check on fedora-coreos we do not have the related `shim-ia32` and `grub2-efi-ia32` packages:
```
$ podman run --rm -it quay.io/fedora/fedora-coreos:testing-devel bash
bash-5.2# rpm -qa | grep grub
grub2-common-2.12-10.fc41.noarch
grub2-pc-modules-2.12-10.fc41.noarch
grub2-tools-minimal-2.12-10.fc41.x86_64
grub2-tools-2.12-10.fc41.x86_64
grub2-efi-x64-2.12-10.fc41.x86_64
grub2-pc-2.12-10.fc41.x86_64
bash-5.2# rpm -qa | grep shim
shim-x64-15.8-3.x86_64

bash-5.2# cat /etc/os-release 
PRETTY_NAME="Fedora CoreOS 41.20241106.20.0"

bash-5.2# ls -al /usr/lib/bootupd/updates/EFI/ -R
/usr/lib/bootupd/updates/EFI/:
total 0
drwxr-xr-x. 1 root root  20 Nov  7 08:21 .
drwxr-xr-x. 1 root root  40 Nov  7 08:21 ..
drwxr-xr-x. 1 root root  40 Nov  7 08:21 BOOT
drwxr-xr-x. 1 root root 100 Nov  7 08:21 fedora

/usr/lib/bootupd/updates/EFI/BOOT:
total 1016
drwxr-xr-x. 1 root root     40 Nov  7 08:21 .
drwxr-xr-x. 1 root root     20 Nov  7 08:21 ..
-rwx------. 4 root root 949424 Jan  1  1970 BOOTX64.EFI
-rwx------. 2 root root  87816 Jan  1  1970 fbx64.efi

/usr/lib/bootupd/updates/EFI/fedora:
total 6664
drwxr-xr-x. 1 root root     100 Nov  7 08:21 .
drwxr-xr-x. 1 root root      20 Nov  7 08:21 ..
-rwx------. 2 root root     110 Jan  1  1970 BOOTX64.CSV
-rwx------. 2 root root 4066624 Jan  1  1970 grubx64.efi
-rwx------. 2 root root  848080 Jan  1  1970 mmx64.efi
-rwx------. 4 root root  949424 Jan  1  1970 shim.efi
-rwx------. 4 root root  949424 Jan  1  1970 shimx64.efi
```

HuijingHei:
Check that we have `shim-ia32-15.8-3.x86_64` and `grub2-efi-ia32-2.12-10.fc41.x86_64` on fedora-silverblue image:
```
fedora@hhei-fedora:~/data$ podman run --rm -it quay.io/fedora/fedora-silverblue:41 bash
bash-5.2# rpm -qa | grep shim
shim-x64-15.8-3.x86_64
shim-ia32-15.8-3.x86_64
bash-5.2# rpm -qa | grep grub2-efi
grub2-efi-ia32-2.12-10.fc41.x86_64
grub2-efi-x64-2.12-10.fc41.x86_64

bash-5.2# ls -al /usr/lib/bootupd/updates/EFI/ -R
/usr/lib/bootupd/updates/EFI/BOOT:
total 1820
drwxr-xr-x. 1 root root     84 Nov  7 08:04 .
drwxr-xr-x. 1 root root     20 Nov  7 08:04 ..
-rwx------. 3 root root 747681 Jan  1  1970 BOOTIA32.EFI
-rwx------. 4 root root 949424 Jan  1  1970 BOOTX64.EFI
-rwx------. 2 root root  70360 Jan  1  1970 fbia32.efi
-rwx------. 2 root root  87816 Jan  1  1970 fbx64.efi

/usr/lib/bootupd/updates/EFI/fedora:
total 11012
drwxr-xr-x. 1 root root     192 Nov  7 08:04 .
drwxr-xr-x. 1 root root      20 Nov  7 08:04 ..
-rwx------. 2 root root     112 Jan  1  1970 BOOTIA32.CSV
-rwx------. 2 root root     110 Jan  1  1970 BOOTX64.CSV
-rwx------. 2 root root 3022144 Jan  1  1970 grubia32.efi
-rwx------. 2 root root 4066624 Jan  1  1970 grubx64.efi
-rwx------. 2 root root  673992 Jan  1  1970 mmia32.efi
-rwx------. 2 root root  848080 Jan  1  1970 mmx64.efi
-rwx------. 4 root root  949424 Jan  1  1970 shim.efi
-rwx------. 3 root root  747681 Jan  1  1970 shimia32.efi
-rwx------. 4 root root  949424 Jan  1  1970 shimx64.efi
```

travier:
It looks like the adoption phase does not install files if they are missing on the disk.

Initially I was thinking that we should not install them but then it means that we can not add new files or rename existing ones. So instead, if bootupd does not find a file on the partition during adoption, it should install it.

travier:
So the test for this case should:
- Setup an EFI folder where some files are missing from the expected list in the bootupd manifest.
- Adopt the system without failing
- Verify that validation does fail on the missing files as we do not install them

Or we could decide to always make sure all files are installed at the end of all update operations.

travier:
In short, I think the code change is right, the test is not.

HuijingHei:
> not install files if they are missing on the disk.

The main reason is that because of `relative_diff_to()`, can not install files if they are missing on the disk, so the unit test is to test that, but agree that should add the integration test for your [comment](https://github.com/coreos/bootupd/pull/765#issuecomment-2464430648).

Failed without the code change:
```
# cargo test filetree::tests::test_apply_with_file
running 1 test
test filetree::tests::test_apply_with_file ... FAILED

failures:

---- filetree::tests::test_apply_with_file stdout ----
Error: test removed files with relative_diff

Caused by:
    0: removing "testfile"
    1: No such file or directory (os error 2)
```

Pass with the change:
```
# cargo test filetree::tests::test_apply_with_file
running 1 test
test filetree::tests::test_apply_with_file ... ok
```

HuijingHei:
> * Verify that validation does fail on the missing files as we do not install them

Do you mean that in this case the validate error is accepted?
```
# bootupctl validate

Removed: BOOT/fbia32.efi
Removed: fedora/BOOTIA32.CSV
Removed: fedora/shimia32.efi
Removed: BOOT/BOOTIA32.EFI
Removed: fedora/grubia32.efi
Removed: fedora/mmia32.efi
error: Caught validation errors
```

HuijingHei:
Thanks @travier for the pointer!
Ready to review, and add test according to Timothée's [comment](https://github.com/coreos/bootupd/pull/765#issuecomment-2464430648).

HuijingHei:
Confused about the CI failed, not sure why the `fuse-libs` is included.

```
Will download: 1 package (98.9?kB)

warning: Found 1 packages to download in cache-only mode

Downloading from 'fedora'...done

error: Installing packages: Downloading from 'fedora': Cannot download Packages/f/fuse-libs-2.9.9-22.fc41.x86_64.rpm: All mirrors were tried; Last error: Curl error (6): Couldn't resolve host name for https://dl.fedoraproject.org/pub/fedora/linux/releases/41/Everything/x86_64/os/Packages/f/fuse-libs-2.9.9-22.fc41.x86_64.rpm [Could not resolve host: dl.fedoraproject.org]
```

travier:
> > * Verify that validation does fail on the missing files as we do not install them
> 
> Do you mean that in this case the validate error is accepted?
> 
> ```
> # bootupctl validate
> 
> Removed: BOOT/fbia32.efi
> Removed: fedora/BOOTIA32.CSV
> Removed: fedora/shimia32.efi
> Removed: BOOT/BOOTIA32.EFI
> Removed: fedora/grubia32.efi
> Removed: fedora/mmia32.efi
> error: Caught validation errors
> ```

Yes, right now, if you adopt a system with some files missing, it will fail validate and that's "expected" until we fix the update to also install the missing files.

HuijingHei:
> Yes, right now, if you adopt a system with some files missing, it will fail validate and that's "expected" until we fix the update to also install the missing files.

The validate failed after adopt is because we use [relative_diff_to()](https://github.com/coreos/bootupd/blob/main/src/efi.rs#L278), which will ignore any files or directories that are not part of the original tree, this means we will not install the files from updates dir if they are missing in `/boot/efi/EFI`, this is by design, right? Do you mean we should use [diff()](https://github.com/coreos/bootupd/blob/main/src/efi.rs#L345)?

Maybe we can just ignore the validate error after adopt in this case, if run `update` the next time, then it will use [diff()](https://github.com/coreos/bootupd/blob/main/src/efi.rs#L345) and will install the missing files.

HuijingHei:
Thanks @travier for the rebase.

travier:
> The validate failed after adopt is because we use [relative_diff_to()](https://github.com/coreos/bootupd/blob/main/src/efi.rs#L278), which will ignore any files or directories that are not part of the original tree, this means we will not install the files from updates dir if they are missing in `/boot/efi/EFI`, this is by design, right? Do you mean we should use [diff()](https://github.com/coreos/bootupd/blob/main/src/efi.rs#L345)?
> 
> Maybe we can just ignore the validate error after adopt in this case, if run `update` the next time, then it will use [diff()](https://github.com/coreos/bootupd/blob/main/src/efi.rs#L345) and will install the missing files.

Hum, maybe we should indeed turn those errors into a warning then.

Maybe it looks like the adoption phase does not properly record that those files are missing or maybe it's because I'm calling `update` instead of `adopt-and-update` in the bootloader-update systems unit on Atomic Desktops?

I can reproduce the issue on my system when I remove `/boot/bootupd-state.json` and adopt again.

It not does look like an update re-added the files for me so this would need more testing.

HuijingHei:
> It not does look like an update re-added the files for me so this would need more testing.

Will do more testing for this.

travier:
I've tested that fixes the issue for me on Silverblue. Let's make a release to get this down to Fedora 41.

PR #773, opened 2024-11-14
tests/e2e-upgrade: update grub2 version

HuijingHei:


HuijingHei:
During e2e-upgrade tests, what we do is building image A with the older grub2 version, also build image B with newer grub2 version, starting A image and upgrade to B. When building image with older grub2 version (actually `grub2-tools-minimal`) which has the dependency of `fuse-libs-2.9.9-22.fc41.x86_64 (fedora)`, and finally got error that might be the Fedora infra issue, but would be better if can workaround this.
```
error: Installing packages: Downloading from 'fedora': Cannot download Packages/f/fuse-libs-2.9.9-22.fc41.x86_64.rpm: All mirrors were tried; Last error: Curl error (6): Couldn't resolve host name for https://dl.fedoraproject.org/pub/fedora/linux/releases/41/Everything/x86_64/os/Packages/f/fuse-libs-2.9.9-22.fc41.x86_64.rpm [Could not resolve host: dl.fedoraproject.org]
```
Check `grub2-tools-minimal` on f40 that still needs the fuse-libs dependency, but f41 does not need it, so update the override grub2 version to f41.
```
# cat /etc/os-release
NAME="Fedora Linux"
VERSION="40 (Container Image)"
# dnf repoquery --whatrequires fuse-libs | grep grub2 
grub2-tools-minimal-1:2.06-119.fc40.x86_64
grub2-tools-minimal-1:2.06-123.fc40.x86_64

# cat /etc/os-release
NAME="Fedora Linux"
VERSION="41 (Container Image)"
# dnf repoquery --whatrequires fuse-libs | grep grub2
```

Not sure if this can fix the upgrade failed CI, waiting for the result.

HuijingHei:
Check https://src.fedoraproject.org/rpms/grub2/c/863039eee208bb902d037d93b81c92625a1028a3?branch=f41 is using `fuse3`, try [grub2-2.12-4.fc41](https://koji.fedoraproject.org/koji/buildinfo?buildID=2523645)

travier (approved):

Issue #789, opened 2024-12-03
Add GRUB static config migration as sub command


travier:
We should convert the script to Rust and include it in bootupd so that it can be used by both IoT & Atomic Desktops directly.

See: https://pagure.io/workstation-ostree-config/pull-request/591
See: https://gitlab.com/fedora/ostree/sig/-/issues/35

champtar:
> We should convert the script to Rust and include it in bootupd so that it can be used by both IoT & Atomic Desktops directly.

curious, what / where is the original script ?

travier:
I initial worked on it in https://hackmd.io/B8lMCzLFQjGgr5jhl_Iw-w and posted it to https://gitlab.com/fedora/ostree/sig/-/issues/35.

Issue #578, opened 2023-12-06
Support for adopt --with-static-configs


cgwalters:
It'd be nice to be able to convert Fedora ostree+Anaconda users to `bootupd --with-static-configs`.

champtar:
Until this is implemented, one can just reinstall the bootloader (tested on EL 9.5)
```
cat > /usr/lib/systemd/system/bootloader-adopt.service <<'EOF'
[Unit]
Description=bootloader adopt
ConditionPathExists=!/boot/bootupd-state.json
Before=bootloader-update.service

[Service]
Type=oneshot
ExecStart=/usr/libexec/bootloader-adopt
RemainAfterExit=yes
# Keep this stuff in sync with SYSTEMD_ARGS_BOOTUPD in general
PrivateNetwork=yes
ProtectHome=yes
KillMode=mixed
MountFlags=slave
# inhibit locks are ignored by reboot/poweroff (need to use systemctl)
# so delay being killed as much as possible
KillSignal=SIGCONT

[Install]
WantedBy=multi-user.target
EOF
```
```
cat > /usr/libexec/bootloader-adopt <<'EOF'
#!/usr/bin/bash
set -euo pipefail

# We reinstall the bootloader as we also want to "adopt" grub.cfg
# https://github.com/coreos/bootupd/issues/578

PART=$(findmnt --noheadings --nofsroot --output SOURCE /boot)
echo "/boot partition = ${PART}"
DISK=$(lsblk --paths --noheadings --output PKNAME "${PART}")
echo "/boot disk = ${DISK}"
echo "Reinstalling bootloader using bootupd"
bootupctl backend install --auto --write-uuid --update-firmware --device "${DISK}" /
sync -f /boot
if [ -e /boot/efi ]; then
    sync -f /boot/efi
fi
EOF
```

PR #790, opened 2024-12-03
Import GRUB static migration code

travier:
migrate-static-grub-config: Add GRUB static migration subcommand

Add a hidden subcommand that migrates existing systems using a dynamic
GRUB config to a static one.

This command is expected to be run after a successful bootloader update.
One way to do that is to add it as a droppin unit config for the
`bootloader-update.service` unit included in this repo:

```
$ cat /usr/lib/systemd/system/bootloader-update.service.d/migrate-static-grub-config.conf
[Service]
ExecStart=/usr/bin/bootupctl migrate-static-grub-config
```

This will be used on Atomic Desktops & IoT systems to migrate systems to
a static GRUB config before enabling composefs as GRUB curently does not
interact well with it [1].

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2308594

See: https://gitlab.com/fedora/ostree/sig/-/issues/35
See: https://pagure.io/workstation-ostree-config/pull-request/591
See: https://fedoraproject.org/wiki/Changes/ComposefsAtomicDesktops
Fixes: https://github.com/coreos/bootupd/issues/789

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -9,10 +9,16 @@ use crate::model::{ComponentStatus, ComponentUpdatable, ContentMetadata, SavedSt
 use crate::util;
 use anyhow::{anyhow, Context, Result};
 use clap::crate_version;
+use fn_error_context::context;
+use libc::mode_t;
+use libc::{S_IRGRP, S_IROTH, S_IRUSR, S_IWUSR};
+use openat_ext::OpenatDirExt;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
-use std::path::Path;
+use std::fs::{self, File};
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::path::{Path, PathBuf};
 
 pub(crate) enum ConfigMode {
     None,
@@ -486,6 +492,150 @@ pub(crate) fn client_run_validate() -> Result<()> {
     if caught_validation_error {
         anyhow::bail!("Caught validation errors");
     }
+    Ok(())
+}
+
+#[context("Migrating to a static GRUB config")]
+pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
+    // Did we already complete the migration?
+    let mut cmd = std::process::Command::new("ostree");
+    let result = cmd
+        .args([
+            "config",
+            "--repo=/sysroot/ostree/repo",
+            "get",
+            "sysroot.bootloader",
+        ])
+        .output()
+        .context("Querying ostree sysroot.bootloader")?;
+    if !result.status.success() {
+        // ostree will exit with a non zero return code if the key does not exists
+        println!("ostree repo 'sysroot.bootloader' config option is not set yet");
+    } else {
+        let res = String::from_utf8(result.stdout)
+            .with_context(|| "decoding as UTF-8 output of ostree command")?;
+        let bootloader = res.trim_end();
+        if bootloader == "none" {
+            println!("Already using a static GRUB config");
+            return Ok(());
+        }
+        println!(
+            "ostree repo 'sysroot.bootloader' config option is currently set to: '{}'",
+            bootloader
+        );
+    }
+
+    // Remount /boot read write just for this unit (we are called in a slave mount namespace by systemd)
+    ensure_writable_boot()?;
+
+    let grub_config_dir = PathBuf::from("/boot/grub2");
+    let dirfd = openat::Dir::open(&grub_config_dir).context("Opening /boot/grub2")?;
+
+    // We mark the bootloader as BLS capable to disable the ostree-grub2 logic.
+    // We can do that as we know that we are run after the bootloader has been
+    // updated and all recent GRUB2 versions support reading BLS configs.
+    // Ignore errors as this is not critical. This is a safety net if a user
+    // manually overwrites the (soon) static GRUB config by calling `grub2-mkconfig`.
+    // We need this until we can rely on ostree-grub2 being removed from the image.
+    println!("Marking bootloader as BLS capable...");
+    _ = File::create("/boot/grub2/.grub2-blscfg-supported");
+
+    // Migrate /boot/grub2/grub.cfg to a static GRUB config if it is a symlink
+    let grub_config_filename = PathBuf::from("/boot/grub2/grub.cfg");
+    match dirfd.read_link("grub.cfg") {
+        Err(_) => {
+            println!(
+                "'{}' is not a symlink, nothing to migrate",
+                grub_config_filename.display()
+            );
+        }
+        Ok(path) => {
+            println!("Migrating to a static GRUB config...");
+
+            // Resolve symlink location
+            let mut current_config = grub_config_dir.clone();
+            current_config.push(path);
+
+            // Backup the current GRUB config which is hopefully working right now
+            let backup_config = PathBuf::from("/boot/grub2/grub.cfg.backup");
+            println!(
+                "Creating a backup of the current GRUB config '{}' in '{}'...",
+                current_config.display(),
+                backup_config.display()
+            );
+            fs::copy(&current_config, &backup_config).context("Failed to backup GRUB config")?;
+
+            // Read the current config, strip the ostree generated GRUB entries and
+            // write the result to a temporary file
+            println!("Stripping ostree generated entries from GRUB config...");
+            let current_config_file =
+                File::open(current_config).context("Could not open current GRUB config")?;
+            let stripped_config = String::from("grub.cfg.stripped");
+            // mode = -rw-r--r-- (644)
+            let mut writer = BufWriter::new(
+                dirfd
+                    .write_file(
+                        &stripped_config,
+                        (S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH) as mode_t,
+                    )
+                    .context("Failed to open temporary GRUB config")?,
+            );
+            let mut skip = false;
+            for line in BufReader::new(current_config_file).lines() {
+                let line = line.context("Failed to read line from GRUB config")?;
+                if line == "### END /etc/grub.d/15_ostree ###" {
+                    skip = false;
+                }
+                if skip {
+                    continue;
+                }
+                if line == "### BEGIN /etc/grub.d/15_ostree ###" {
+                    skip = true;
+                }
+                writer
+                    .write_all(&line.as_bytes())
+                    .context("Failed to write stripped GRUB config")?;
+                writer
+                    .write_all(b"\n")
+                    .context("Failed to write stripped GRUB config")?;
+            }
+            writer
+                .flush()
+                .context("Failed to write stripped GRUB config")?;
+
+            // Sync changes to the filesystem (ignore failures)
+            let _ = dirfd.syncfs();
+
+            // Atomically exchange the configs
+            dirfd
+                .local_exchange(&stripped_config, "grub.cfg")
+                .context("Failed to exchange symlink with current GRUB config")?;
+
+            // Sync changes to the filesystem (ignore failures)
+            let _ = dirfd.syncfs();
+
+            println!("GRUB config symlink successfully replaced with the current config");
+
+            // Remove the now unused symlink (optional cleanup, ignore any failures)
+            _ = dirfd.remove_file(&stripped_config);
+        }
+    };
+
+    println!("Setting 'sysroot.bootloader' to 'none' in ostree repo config...");
+    let status = std::process::Command::new("ostree")
+        .args([
+            "config",
+            "--repo=/sysroot/ostree/repo",
+            "set",
+            "sysroot.bootloader",
+            "none",
+        ])
+        .status()?;
+    if !status.success() {
+        anyhow::bail!("Failed to set 'sysroot.bootloader' to 'none' in ostree repo config");
+    }
+
+    println!("Static GRUB config migration completed successfully");
     Ok(())
 }
 
--- a/src/cli/bootupctl.rs
+++ b/src/cli/bootupctl.rs
@@ -59,6 +59,12 @@ pub enum CtlVerb {
     AdoptAndUpdate,
     #[clap(name = "validate", about = "Validate system state")]
     Validate,
+    #[clap(
+        name = "migrate-static-grub-config",
+        hide = true,
+        about = "Migrate a system to a static GRUB config"
+    )]
+    MigrateStaticGrubConfig,
 }
 
 #[derive(Debug, Parser)]
@@ -96,6 +102,7 @@ impl CtlCommand {
             CtlVerb::Backend(CtlBackend::Install(opts)) => {
                 super::bootupd::DCommand::run_install(opts)
             }
+            CtlVerb::MigrateStaticGrubConfig => Self::run_migrate_static_grub_config(),
         }
     }
 
@@ -135,6 +142,12 @@ impl CtlCommand {
     fn run_validate() -> Result<()> {
         ensure_running_in_systemd()?;
         bootupd::client_run_validate()
+    }
+
+    /// Runner for `migrate-static-grub-config` verb.
+    fn run_migrate_static_grub_config() -> Result<()> {
+        ensure_running_in_systemd()?;
+        bootupd::client_run_migrate_static_grub_config()
     }
 }
 


src/bootupd.rs
```diff
@@ -489,6 +496,88 @@ pub(crate) fn client_run_validate() -> Result<()> {
     Ok(())
 }
 
+pub(crate) fn client_run_migrate() -> Result<()> {
+    // Used to condition execution of this unit at the systemd level
+    let stamp_file="/var/lib/.fedora_atomic_desktops_static_grub";
```
cgwalters, line 501: Missing a `cargo fmt`.

Also... now that this is part of bootupd calling the stamp file name "fedora_atomic_desktops_" seems odd.

Bikeshedding things a bit more...bootupd already has its own little database where we could store this state.

I'm also fine just keeping it as a stamp file, but how about e.g. `.bootupd-static-migration-complete`? Also since this is about data in `/boot` I think we should probably keep the stamp file there?

src/bootupd.rs
```diff
@@ -489,6 +496,88 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+    //     echo "Regenerating a pure BLS GRUB config..."
+    //     grub2-mkconfig -o /boot/grub2/grub.cfg
+    // fi
+    if !destdir.exists(&first_dir_tmp)? {
+        copy_dir(destdir, first_dir.as_str(), &first_dir_tmp)?;
+        updates.insert(first_dir, first_dir_tmp);
+    }
+
+    // # Remount /sysroot read write just for this unit (we are called in a slave mount namespace by systemd)
+    ensure_writable_sysroot();
```
cgwalters, line 564: Hm, is this necessary? We're just forking off ostree which knows how to do this.

cgwalters (commented):
Thanks for this!

src/bootupd.rs
travier: If we want to have this represented in bootup state directly then we would have to add a new mode to distinguish between the bootupd managed static grub configs and the ones that we imported from a system that are still user managed because they may contain arbitrary changes, os-prober systems, etc.

src/bootupd.rs
travier: Indeed

travier:
I immediately ran into https://github.com/fedora-selinux/selinux-policy/pull/2444 while testing this.

src/bootupd.rs
cgwalters: Hmm...I more meant that we add a new key to the JSON file

HuijingHei:
Overall LGTM, I can help to do some testing if you have some instructions.

travier:
How I'm (manually) testing this change:
- Install Fedora Silverblue 40 in two VMs, one using BIOS & another UEFI (see https://gitlab.com/fedora/ostree/scripts for scripts that automate that), do the same for Fedora Silverblue 41
- Update the Fedora 40 VMs to Fedora 41
- Optional: Make a snapshot of the VM
- Build bootupd and copy it to the VM:
```
$ cargo build --release
$ scp target/release/bootupd silverblue-41-uefi-bootupd:
[silverblue]$ sudo rpm-ostree usroverlay
[silverblue]$ sudo cp bootupd /usr/bin/bootupctl && sudo cp bootupd /usr/libexec/bootupd
```
- Run the migration:
```
$ sudo bootupctl migrate
```
- Validate the state of the system:

  - Things to verify, on a system installed from Fedora 41:

```
root@fedora:~# sudo ls -alh /boot/grub2
total 16K
drwx------. 2 root root 4.0K Oct 25 13:29 .
drwxr-xr-x. 7 root root 4.0K Nov 29 14:37 ..
-rw-r--r--. 1 root root   53 Oct 25 13:29 bootuuid.cfg
-rw-r--r--. 1 root root 2.6K Oct 25 13:29 grub.cfg

root@fedora:~# grep ostree /boot/grub2/grub.cfg
<empty>

root@fedora:~# grep blscfg /boot/grub2/grub.cfg
blscfg

root@fedora:~# sudo cat /sysroot/ostree/repo/config
[core]
repo_version=1
mode=bare

[sysroot]
readonly=true
bootloader=none
```

  - Things to verify, on a system installed from an older release:

```
root@fedora:~# ls -alh /boot/grub2
total 64K
drwx------. 5 root root 4.0K Dec  2 20:30 .
drwxr-xr-x. 6 root root 4.0K Dec  2 20:13 ..
-rw-r--r--. 1 root root   64 Oct 18 17:50 device.map
drwx------. 2 root root 4.0K Jan  1  1970 fonts
-rw-r--r--. 1 root root    0 Dec  2 20:30 .grub2-blscfg-supported
-rw-------. 1 root root 7.0K Dec  2 20:30 grub.cfg
-rw-------. 1 root root 8.8K Nov 29 10:15 grub.cfg.backup
-rw-------. 1 root root 1.0K Dec  2 15:22 grubenv
drwxr-xr-x. 2 root root  20K Dec  2 20:13 i386-pc
drwxr-xr-x. 2 root root 4.0K Dec  2 20:13 locale

root@fedora:~# ls -alh /boot/grub2/.grub2-blscfg-supported 
-rw-r--r--. 1 root root 0 Dec  2 20:30 /boot/grub2/.grub2-blscfg-supported

root@fedora:~# grep ostree /boot/grub2/grub.cfg
  set kernelopts="root=UUID=f59453a5-036d-4b9a-b1a2-4428bf5eaecf ro rootflags=subvol=root/ostree/deploy/fedora/deploy/a7eefc0751cf4688e6daf0d998d24a435ae4d44ab193d8dce7e01f66d366fa08.0 rhgb quiet "
### BEGIN /etc/grub.d/15_ostree ###
### END /etc/grub.d/15_ostree ###

root@fedora:~# grep blscfg /boot/grub2/grub.cfg
# The blscfg command parses the BootLoaderSpec files stored in /boot/loader/entries and
insmod blscfg
blscfg

root@fedora:~# cat /sysroot/ostree/repo/config
[core]
repo_version=1
mode=bare

[sysroot]
readonly=true
bootloader=none
```

travier:
So this works as far as I've tested but we have no tests for it yet.

travier:
I've added the systemd unit to this PR and rebased it on top of https://github.com/coreos/bootupd/pull/803 to avoid conflicts for the systemd unit setup in the Makefile & specfile.

I've split the clippy lint fixes in https://github.com/coreos/bootupd/pull/804.

contrib/packaging/bootupd-static-grub-migration.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=bootupd static GRUB config migration
+Documentation=https://github.com/coreos/bootupd
+ConditionPathExists=!/boot/.bootupd-static-migration-complete
+RequiresMountsFor=/sysroot /boot
+# Only run after a successful bootloader update
```
cgwalters, line 6: Are there any systems which would enable the static migration but *not* the update service?

If not, why wouldn't we just fold this functionality into the `bootloader-update.service` via

```
ExecStart=bootupctl migrate
ExecStart=bootupctl update
```

Or I guess just have `bootupctl update --migrate-if-needed` or so.

src/bootupd.rs
```diff
@@ -489,6 +490,149 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+    let mut ostree_cmd = std::process::Command::new("ostree");
+    let result = ostree_cmd
+        .args([
+            "config",
+            "--repo=/sysroot/ostree/repo",
+            "get",
+            "sysroot.bootloader",
+        ])
+        .output()
+        .with_context(|| "failed to call ostree command. Not performing migration")?;
```
cgwalters, line 507: contexts should be a single sentence fragment; this will appear strangely when rendered.
I'd just use:
`.context("Querying ostree sysroot.bootloader")` or so

src/bootupd.rs
```diff
@@ -489,6 +490,149 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+    // Remount /boot read write just for this unit (we are called in a slave mount namespace by systemd)
+    ensure_writable_boot()?;
+
+    let grub_config_dir = PathBuf::from("/boot/grub2");
+    let Ok(dirfd) = openat::Dir::open(&grub_config_dir).with_context(|| "a") else {
+        anyhow::bail!(
+            "Could not open {}. Not performing migration.",
+            grub_config_dir.display()
+        );
+    };
```
cgwalters, line 536: Here and elsewhere I think the translation from "careful error checking shell" to Rust doesn't make sense because we can just use Rust's builtin `?`. Like this

```
let dirfd = Dir::open(&grub_config_dir).context("Opening /boot/grub2")?;
```

The "Not performing migration" bit is implicit in the whole operation.

src/bootupd.rs
```diff
@@ -489,6 +490,149 @@ pub(crate) fn client_run_validate() -> Result<()> {
     Ok(())
 }
 
+pub(crate) fn client_run_migrate() -> Result<()> {
```
cgwalters, line 493: I'd add `#[context("Migrating to static grub config")]` here

src/bootupd.rs
```diff
@@ -489,6 +490,149 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+            current_config.push(path);
+
+            // Backup the current GRUB config which is hopefully working right now
+            let backup_config = PathBuf::from("/boot/grub2/grub.cfg.backup");
+            println!(
+                "Creating a backup of the current GRUB config '{}' in '{}'...",
+                current_config.display(),
+                backup_config.display()
+            );
+            fs::copy(&current_config, &backup_config).map_err(|e| {
```
cgwalters, line 561: Same here, this I think is clearer as just
`fs::copy(...).context("Failed to back up GRUB config")?;`

src/bootupd.rs
cgwalters: (Which would then avoid the duplicate "Not performing migration" in all the error contexts)

contrib/packaging/bootupd-static-grub-migration.service
travier: We indeed want all the systems that do the migration to update their bootloader before.

We don't need to do the migration on systems which already have a static GRUB configs and we can check the state for that, so maybe this should indeed be folded into a single command and moved from a stamp file to a field in the state JSON.

If we merge both units then it also means that we have to careful when rolling this out as we won't be easily able to enable one or the other independently. But that's also kind of OK as I'm looking at doing that in F41 as well before the F42 release.

HuijingHei:
Ask a silly question, is there any reason should do the migration only when `/boot/grub2/grub.cfg` is symlink?

Check on silverblue40 (using efi), it is symlink to `/boot/loader/grub.cfg`, but on silverblue41, it is a normal file.

travier:
> Ask a silly question, is there any reason should do the migration only when `/boot/grub2/grub.cfg` is symlink?

As far I know, this is the marker that lets us tell a system with a dynamic config from one with a static one.

> Check on silverblue40 (using efi), it is symlink to `/boot/loader/grub.cfg`, but on silverblue41, it is a normal file.

On Silverblue 40, GRUB is setup using a dynamic config like package mode Fedora and on Silverblue 41, it is setup by bootupd with a static config.

src/cli/bootupctl.rs
```diff
@@ -58,6 +58,8 @@ pub enum CtlVerb {
     AdoptAndUpdate,
     #[clap(name = "validate", about = "Validate system state")]
     Validate,
+    #[clap(name = "migrate", about = "Migrate a system to static a GRUB config")]
```
champtar, line 61: Migrate is a bit generic, should this be migrate-grub ?
```suggestion
    #[clap(name = "migrate-grub", about = "Migrate a system to a static GRUB config")]
```

contrib/packaging/bootupd-static-grub-migration.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=bootupd static GRUB config migration
+Documentation=https://github.com/coreos/bootupd
+ConditionPathExists=!/boot/.bootupd-static-migration-complete
```
champtar, line 4: If we don't want to leave another stamp file around we can use
```suggestion
ConditionPathIsSymbolicLink=/etc/grub2/grub.cfg
```

src/bootupd.rs
```diff
@@ -489,6 +491,127 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+        ])
+        .status()?;
+    if !status.success() {
+        anyhow::bail!("Failed to set 'sysroot.bootloader' to 'none' in ostree repo config");
+    }
+
+    // Migration complete, let's write the stamp file
+    File::create(stamp_file)?;
+
+    println!("Static GRUB config migration completed successfully!");
```
champtar, line 611: In the end this script doesn't switch to static config (`"static-configs":null`), it just ensure we are properly switched to a BLS compatible config (really old grub or if we had ostree-grub2)

Also `grub2-mkconfig` will actually still use `grub2-probe` (`os-prober`) and fail with composefs enabled,
so we can't update + migrate + enable composefs in a single update (my use case is offline appliances, possibly skipping updates).

Shouldn't we switch to static config generated by bootupd (same as during install) ?

src/cli/bootupctl.rs
travier: Sure, I'll do that.

contrib/packaging/bootupd-static-grub-migration.service
travier: It's `/boot/grub2/grub.cfg` that is a symlink but we can not do that as this would skip all the logic after the symlink change that is still needed.

src/bootupd.rs
```diff
@@ -489,6 +491,127 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+            fs::copy(&current_config, &current_config_copy)
+                .context("Failed to copy the current GRUB config")?;
+
+            // Atomically exchange the configs
+            dirfd
+                .local_exchange("grub.cfg.current", "grub.cfg")
+                .context("Failed to exchange symlink with current GRUB config")?;
+
+            // Remove the now unused symlink (optional cleanup, ignore any failures)
+            _ = dirfd.remove_file("grub.cfg.current");
```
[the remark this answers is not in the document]
travier: > Should we also delete the target of the symlink (`/boot/loader/grub.cfg`) ?
> It'll be delete on the next ostree deploy, but it's cleaner IMO.

We could, but it's harmless and as you said it will be "removed" (not generated) for the next deployment. Keeping it makes things easier for debugging for now.

> Or we could just `mv $(readlink -ne /boot/grub2/grub.cfg) /boot/grub2/grub.cfg`

AFAIK that would not be atomic.

> I would also add a `sync` call after the `mv`

We should indeed do that. I'll add this.

src/bootupd.rs
travier: > In the end this script doesn't switch to static config (`"static-configs":null`), it just ensure we are properly switched to a BLS compatible config (really old grub or if we had ostree-grub2)

It does switch the system to a static GRUB config that is no longer updated on deployment changes. What it does not do is use the static config from bootupd, to preseve changes that users may have done in their existing configs. Thus we can not record those in the static-configs entry in the state JSON as those are not the same configs.

> Also `grub2-mkconfig` will actually still use `grub2-probe` (`os-prober`) and fail with composefs enabled, so we can't update + migrate + enable composefs in a single update (my use case is offline appliances, possibly skipping updates).

Indeed, this is why I'm looking at pushing that to Fedora 41 as well, before the F42 release. I'll have to investigate that more.

> Shouldn't we switch to static config generated by bootupd (same as during install) ?

We can not do that as that would remove local user config customisations.

champtar: > > Shouldn't we switch to static config generated by bootupd (same as during install) ?

> We can not do that as that would remove local user config customisations.

With your Fedora hat it makes sense to be cautious, but for the appliance use case I would really like to be able to just remove grub-tools completely and have bootupd adopt grub.cfg as well, so there is no drift between an upgraded system and a fresh install

I'm tempted to just run
```
rm /boot/bootupd-state.json
bootupctl backend install --auto --write-uuid /
```

src/bootupd.rs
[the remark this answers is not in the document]
champtar: > AFAIK that would not be atomic.

both files are under a single fs /boot so I don't see why it would not be atomic

travier:
> > Also `grub2-mkconfig` will actually still use `grub2-probe` (`os-prober`) and fail with composefs enabled, so we can't update + migrate + enable composefs in a single update (my use case is offline appliances, possibly skipping updates).
> 
> Indeed, this is why I'm looking at pushing that to Fedora 41 as well, before the F42 release. I'll have to investigate that more.

Thinking about this more, this means that we can not take the approach as written in this current PR as the grub configuration generation would fail as part of the generation on the first boot on F42 with composefs enabled.

We have multiple options:
- Do a plain copy/paste of the config instead and strip the ostree boot entries "manually" by removing all the content between the `#### ...ostree..` GRUB marker in the config. This would let us keep composefs enabled directly on the first boot in F42 but would mean that any migration failure will likely block updates.
- Use the `ostree.prepare-root.composefs=0` kernel argument instead and a migration process similar to the one we did with the [sysroot RO one](https://pagure.io/workstation-ostree-config/blob/f39/f/postprocess.sh) to only enable composefs once the migration has been safely completed. On the first boot in F42, we would do the migration and update the kernel argument only when ready. This would mean that we can not remove the ostree-grub2 package in F42 and we would have to wait for F43 to statically enable composefs.

src/bootupd.rs
champtar: Actually I want https://github.com/coreos/bootupd/issues/578

src/bootupd.rs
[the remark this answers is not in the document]
travier: I though I had read somewhere that only renames of files in the same directory would be atomic but looks like I misread of misunderstood something as I can not find that in: https://manpages.debian.org/testing/manpages-dev/renameat2.2.en.html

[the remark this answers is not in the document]
champtar: In the ERRORS section:

[EXDEV](https://manpages.debian.org/testing/manpages-dev/renameat2.2.en.html#EXDEV)
    oldpath and newpath are not on the same mounted filesystem. (Linux permits a filesystem to be mounted at multiple points, but rename() does not work across different mount points, even if the same filesystem is mounted on both.)

travier:
I reworked this PR to do both the static GRUB config migration and enable composefs via the karg.

Instead of calling `grub2-mkconfig`, I now manually strip all the content between the ` ### BEGIN /etc/grub.d/15_ostree ###` and `### END /etc/grub.d/15_ostree ###` lines. This means that this should be safe to run on a system where composefs is already enabled.

I manually tested this change using the same instructions as https://github.com/coreos/bootupd/pull/790#issuecomment-2539709504.

The only difference is the new name of the hidden argument:

```
$ sudo bootupctl migrate-static-grub-config-composefs
ostree repo 'sysroot.bootloader' config option not set yet
Marking bootloader as BLS capable...
Migrating to a static GRUB config...
Creating a backup of the current GRUB config '/boot/grub2/../loader/grub.cfg' in '/boot/grub2/grub.cfg.backup'...
Stripping ostree generated entries from GRUB config...
GRUB config symlink successfully replaced with the current config
Setting up 'sysroot.bootloader' to 'none' in ostree repo config...
Static GRUB config migration completed successfully
Editing kernel command line arguments to enable composefs...
Composefs migration completed successfully
```

travier:
Hum, unfortunately I think I've found another case where that will not work:
- F41 installed system
  - never updated
  - thus older ostree, not generating composefs by default yet
- Rebase to F42
  - composefs is not enabled in the image thus not generated
  - bootupd enables composefs
  - reboot
  - fails to boot as composefs blob does not exists for the deployment

So this means that we need to enable day 1 composefs in the image for F42 to make sure that ostree from F41 generates the composefs blob and hope that the migration works :/

src/bootupd.rs
```diff
@@ -489,6 +496,173 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+            for line in BufReader::new(current_config_file).lines().map_while(Result::ok) {
+                if line == "### END /etc/grub.d/15_ostree ###" {
+                    skip = false;
+                }
+                if skip {
+                    continue;
+                }
+                if line == "### BEGIN /etc/grub.d/15_ostree ###" {
+                    skip = true;
+                }
```
champtar, line 589: No need to keep "### BEGIN /etc/grub.d/15_ostree ###" in the new file
```suggestion
                if line == "### BEGIN /etc/grub.d/15_ostree ###" {
                    skip = true;
                }
                if line == "### END /etc/grub.d/15_ostree ###" {
                    skip = false;
                }
                if skip {
                    continue;
                }
```

src/bootupd.rs
```diff
@@ -489,6 +496,173 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+    // Update all BLS configs to add a kernel argument to tell ostree to use composefs
+    // client_migrate_static_grub_config which is called before us made sure
+    // that /boot was mounted and writable
+    // ostree admin kargs edit-in-place --append-if-missing=ostree.prepare-root.composefs=1
+    println!("Editing kernel command line arguments to enable composefs...");
+    let mut cmd = Command::new("ostree");
+    cmd.arg("admin")
+        .arg("kargs")
+        .arg("edit-in-place")
+        .arg("--append-if-missing=ostree.prepare-root.composefs=1");
```
champtar, line 653: This will break people that want to use 'signed' or 'verity' no ?

Migrating to grub static config is just one of the prerequisite to enable composefs (tmpfs-on-tmp, having the composefs blob actually generated, patching fstab to avoid an error (for now done by bootc generator), and fixed kdump), so I don't think enabling composefs should be done by bootupd to avoid tight coupling

champtar (changes requested):

src/bootupd.rs
cgwalters: > This will break people that want to use 'signed' or 'verity' no ?

At a quick glance at the code, I think it will disable signatures for the `signed` case, but leave the verity case unchanged.

> so I don't think enabling composefs should be done by bootupd to avoid tight coupling

But yes, I agree with you.

cgwalters (changes requested):
Marking requested-changes per discussion

src/bootupd.rs
```diff
@@ -489,6 +496,173 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+
+    // Update all BLS configs to add a kernel argument to tell ostree to use composefs
+    // client_migrate_static_grub_config which is called before us made sure
+    // that /boot was mounted and writable
+    // ostree admin kargs edit-in-place --append-if-missing=ostree.prepare-root.composefs=1
+    println!("Editing kernel command line arguments to enable composefs...");
+    let mut cmd = Command::new("ostree");
+    cmd.arg("admin")
+        .arg("kargs")
+        .arg("edit-in-place")
```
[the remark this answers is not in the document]
champtar: Agreed this can break rollback
To help with downgrade for Fedora (don't know if it's a thing), you could ship right away
```
cat >> /usr/lib/ostree/prepare-root.conf << 'EOF'
[composefs]
enabled = no
EOF
```

else you can use my script to delete .ostree.cfs after the fact https://github.com/ostreedev/ostree/pull/3366#issuecomment-2599513140

src/bootupd.rs
travier: This will leave the END line only, which is inconsistent. The logic is simpler to understand if I keep both lines.

travier:
>> This will break people that want to use 'signed' or 'verity' no ?
> 
> At a quick glance at the code, I think it will disable signatures for the `signed` case, but leave the verity case unchanged.

If we want to avoid that then this means that we can not rely on the kernel argument for the transition.

If I understand correctly https://github.com/ostreedev/ostree/pull/3353, it means that before this patch, the composefs blob was only generated if the destination system had it explicitly enabled.

This means that a freshly installed Fedora 41, not updated, which includes such an older ostree version, will only generate the composefs metadata if it's enabled in the Fedora 42 image.

I don't know how it would react to a config in the F42 image set to `maybe` (error or ignore? will need to test).

If older `ostree` ignores `maybe` and does nothing, then we need to make sure that the system can not update again after the rebase to F42 until the static GRUB config migration has been completed successfully. Otherwise the next deployment will have the composefs blob generated and then next boot composefs enabled and then updates will fail if we are not using a static GRUB config yet. That would mean blocking all rpm-ostree operations on the migration succeeding.

If older `ostree` errors out on `maybe`, then we are forced to enable compose (`yes`) in F42 and similarly we will have to block any rpm-ostree operation on the static migration having successfully completed as the system will be booted with composefs directly.

Overall, the safest path seems to be setting composefs to enabled in F42 and directly blocking rpm-ostree operations on the migration being successful, via an `ExecStartPre=bootupctl migrate` from a droppin config file. This seems the less likely to fail in a weird way path.

src/bootupd.rs
travier: I have not done that yet but I removed the stamp file and the logic now relies only on the ostree bootloader repo config option.

src/bootupd.rs
```diff
@@ -489,6 +496,173 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+        .arg("--output")
+        .arg("SOURCE")
+        .arg("/");
+    let rootfs_source = util::cmd_output(&mut cmd).context("Failed to query root mount point")?;
+    if rootfs_source == "composefs" {
+        println!("Already using composefs for /");
+        return Ok(());
+    }
+
+    // Update all BLS configs to add a kernel argument to tell ostree to use composefs
```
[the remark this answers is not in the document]
travier: I removed the composefs part.

travier:
Alright, thanks for the reviews, I've updated that again to remove the composefs part and keep this only about the static GRUB config migration following my investigations detailed above.

src/bootupd.rs
```diff
@@ -489,6 +495,150 @@ pub(crate) fn client_run_validate() -> Result<()> {
...
+            writer
+                .flush()
+                .context("Failed to write stripped GRUB config")?;
+
+            // Sync changes to the filesystem (ignore failures)
+            let _ = dirfd.syncfs();
+
+            // Atomically exchange the configs
+            dirfd
+                .local_exchange(&stripped_config, "grub.cfg")
```
champtar, line 611: why not [local_rename](https://docs.rs/openat/latest/openat/struct.Dir.html#method.local_rename) ?

champtar (approved):
One last comment but not a blocker

src/bootupd.rs
travier: `local_exchange` is guaranteed to be atomic which should be safer here

champtar: local_rename uses renameat() and is also atomic (https://man7.org/linux/man-pages/man2/rename.2.html). The big advantage of local_exchange is that it can be used with folders, but it's not needed here.

travier: From the linked man page:
```
       RENAME_EXCHANGE
              Atomically exchange oldpath and newpath.  Both pathnames
              must exist but may be of different types (e.g., one could
              be a non-empty directory and the other a symbolic link).
```
This is what we want here as the "current" config is a symlink, the new one a file.

champtar:
```
If newpath already exists, it will be atomically replaced, so that
there is no point at which another process attempting to access
newpath will find it missing.
```

testing with `mv`
```
touch a
ln -s a b
strace -f -yy -s1000 mv a b
...
renameat2(AT_FDCWD</home/echampetier/git/appliance-deployer/framework/appliance-base>, "a", AT_FDCWD</home/echampetier/git/appliance-deployer/framework/appliance-base>, "b", RENAME_NOREPLACE) = -1 EEXIST (Le fichier existe)
openat(AT_FDCWD</home/echampetier/git/appliance-deployer/framework/appliance-base>, "b", O_RDONLY|O_PATH|O_DIRECTORY) = -1 ENOTDIR (N'est pas un dossier)
newfstatat(AT_FDCWD</home/echampetier/git/appliance-deployer/framework/appliance-base>, "a", {st_mode=S_IFREG|0644, st_size=0, ...}, AT_SYMLINK_NOFOLLOW) = 0
newfstatat(AT_FDCWD</home/echampetier/git/appliance-deployer/framework/appliance-base>, "b", {st_mode=S_IFLNK|0777, st_size=1, ...}, AT_SYMLINK_NOFOLLOW) = 0
renameat(AT_FDCWD</home/echampetier/git/appliance-deployer/framework/appliance-base>, "a", AT_FDCWD</home/echampetier/git/appliance-deployer/framework/appliance-base>, "b") = 0
...
```

renameat works just fine to overwrite a symlink, and my understanding of the doc it's always atomic

travier:
I've tested all the scenarios I could think of with that code and things appears to be working as expected. I would appreciate if we could get this in before the F42 beta freeze.

cgwalters (approved):

Issue #788, opened 2024-12-03
status: support running in a container


cgwalters:
For the use case of a disk image generation tool (e.g. bootc-image-builder), some payloads may support multiple bootloader types, and the partitioning layout could be conditional on those types - whether or not to handle BIOS with a MBR on x86_64 is the biggest example.

Let's make `bootupctl status` runnable in a container - it should print the available payloads. This would allow tooling to query for what's installed in a generic way - and then those tools could e.g. only conditionally create a MBR partition if BIOS is available.

PR #793, opened 2024-12-05
status: support running in a container

HuijingHei:
The output in container is like:
```
bash-5.2# bootupctl status
Available: BIOS EFI
```

Fixes: https://github.com/coreos/bootupd/issues/788


cgwalters (approved):

cgwalters:
Nice!

PR #799, opened 2024-12-13
efi.rs: ensure writable ESP mount

ruihe774:
Just like what we do with `/boot`.


travier:
Could you clarify why we need that? Thanks

ruihe774:
> Could you clarify why we need that? Thanks

For the scenario that ESP is mounted read-only.

We have the handling for `/boot` and IMO it is supposed to also have it with `/boot/efi`. The meaning of mounting `/boot` read-only is to prevent accidentally modifications to it and I think it is also true for `/boot/efi`.

PR #803, opened 2024-12-18
packaging: Add systemd unit to trigger updates on boot

travier:
packaging: Add systemd unit to trigger updates on boot

Add a systemd service unit to trigger an adoption and update on every
boot.

Note that the service is intentionally not enabled by default as it
should be up to the distribution to add a systemd preset if auto-update
for the bootloader is desired.

This unit does not come with any specific restrictions (i.e. EFI or BIOS
only). For an assesment of the safety of updates as performed by
bootupd, see https://github.com/coreos/bootupd/issues/454.

Distributiuons should also apply the restrictions (i.e. EFI or BIOS only
for example) as needed as unit files overrides.

Notably, Fedora CoreOS can not yet enable automatic updates until we get
support for the multiple EFI partitions for RAID setups.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1468
See: https://github.com/coreos/fedora-coreos-config/pull/3042

Initial version from: https://github.com/coreos/bootupd/pull/716
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2332868

---

contrib/packaging/bootupd.spec: Fix EOL at EOF


travier:
Untested yet (but it's a copy paste & update from the previous PR).

cgwalters (approved):
One hesitation I have is that in the medium term I think we want bootupd integrated more with the OS update tool by default.

But yes, this is clearly better than having it external.

PR #811, opened 2025-01-08
Delay being killed on reboot/poweroff

champtar:
Setting KillSignal=SIGCONT is a hack, but it gives us DefaultTimeoutStopSec before being killed (45s on Fedora, 90s on EL9).
The proper systemd way would be to use inhibit locks but they are ignored by reboot/poweroff commands, ie you need the systemctl reboot/poweroff variants, and just using systemd-inhibit makes all the logs starts with systemd-inhibit instead of bootupctl.

### DIFF
--- a/contrib/packaging/bootloader-update.service
+++ b/contrib/packaging/bootloader-update.service
@@ -6,7 +6,12 @@ Documentation=https://github.com/coreos/bootupd
 Type=oneshot
 ExecStart=/usr/bin/bootupctl update
 RemainAfterExit=yes
+PrivateNetwork=yes
+ProtectHome=yes
 MountFlags=slave
+# inhibit locks are ignored by reboot/poweroff (need to use systemctl)
+# so just delay being killed as much as possible
+KillSignal=SIGCONT
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


cgwalters (changes requested):
Thanks for the patch! I'd prefer to have us explicitly ignore SIGTERM in our own code.

champtar:
> Thanks for the patch! I'd prefer to have us explicitly ignore SIGTERM in our own code.

If we want to fully integrate with systemd, we can use sd_notify() and send `EXTEND_TIMEOUT_USEC=` (https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#EXTEND_TIMEOUT_USEC=%E2%80%A6)

If we just ignore `SIGTERM`, we must set `KillMode=mixed` so that subprocess do not also get `SIGTERM` (https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillMode=)

I'm hello world level in rust so if someone wants to implement this please go ahead

cgwalters:
Moving this to https://github.com/coreos/bootupd/pull/812

Thanks again for submitting a patch though!

PR #812, opened 2025-01-08
Catch and ignore SIGTERM during update operations

cgwalters:
Since our updates are non-transactional in general, we should at least be robust against some concurrent invocation of e.g. `reboot`.

Replaces: https://github.com/coreos/bootupd/pull/811

### DIFF
--- a/contrib/packaging/bootloader-update.service
+++ b/contrib/packaging/bootloader-update.service
@@ -6,6 +6,10 @@ Documentation=https://github.com/coreos/bootupd
 Type=oneshot
 ExecStart=/usr/bin/bootupctl update
 RemainAfterExit=yes
+# Keep this stuff in sync with SYSTEMD_ARGS_BOOTUPD in general
+PrivateNetwork=yes
+ProtectHome=yes
+KillMode=mixed
 MountFlags=slave
 
 [Install]
--- a/src/cli/bootupctl.rs
+++ b/src/cli/bootupctl.rs
@@ -6,20 +6,17 @@ use log::LevelFilter;
 use std::os::unix::process::CommandExt;
 use std::process::{Command, Stdio};
 
-static SYSTEMD_ARGS_BOOTUPD: &[&str] = &[
-    "--unit",
-    "bootupd",
-    "--property",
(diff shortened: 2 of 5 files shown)


cgwalters:
Copying the comment from #811 

> If we just ignore SIGTERM, we must set KillMode=mixed so that subprocess do not also get SIGTERM (https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillMode=)

This is an interesting point; I was going to say we don't fork subprocesses but in fact we do.

src/cli/bootupctl.rs
```diff
@@ -13,6 +13,10 @@ static SYSTEMD_ARGS_BOOTUPD: &[&str] = &[
     "PrivateNetwork=yes",
     "--property",
     "ProtectHome=yes",
+    // While only our main process during update catches SIGTERM, we don't
+    // want systemd to send it to other processes.
+    "--property",
+    "KillMode=mixed",
```
champtar, line 19: Also needed in `contrib/packaging/bootloader-update.service` as if we already run under systemd we will not use systemd-run

cgwalters: Thanks, fixed.

contrib/packaging/bootloader-update.service
```diff
@@ -6,7 +6,10 @@ Documentation=https://github.com/coreos/bootupd
 Type=oneshot
 ExecStart=/usr/bin/bootupctl update
 RemainAfterExit=yes
+# Keep this stuff in sync with SYSTEMD_ARGS_BOOTUPD in general
+ProtectHome=yes
```
champtar, line 10: you skipped `PrivateNetwork=yes`

src/cli/bootupctl.rs
```diff
@@ -11,8 +11,13 @@ static SYSTEMD_ARGS_BOOTUPD: &[&str] = &[
     "bootupd",
     "--property",
     "PrivateNetwork=yes",
+    // Keep this stuff in sync with bootloader-update.service
```
champtar, line 14:
```suggestion
    // Keep this stuff in sync with bootloader-update.service
    "--property",
    "PrivateNetwork=yes",
```

champtar (approved):

HuijingHei (approved):
Overall LGTM

Issue #132, opened 2021-01-06
Support updating multiple EFIs in mirrored setups


jlebon:
Follow-up to https://github.com/coreos/bootupd/pull/127 (specifically https://github.com/coreos/bootupd/pull/127#discussion_r546995248). We need to update the EFI in all the devices matching the appropriate GUID, not just the one which wins the `/dev/disk/by-partlabel/EFI-SYSTEM` race.

bgilbert:
One small correction: in the mirrored boot disk case, no partitions are labeled `EFI-SYSTEM`.

jlebon:
Note that just dumbly doing a type GUID match will not work for things like multipath. We solved this in `rdcore` by looking for ESPs only on the same devices which back the boot partition: https://github.com/coreos/coreos-installer/blob/78e2a7a4bf8d9ddb90b5e4fb5729a2550999ad11/src/blockdev.rs#L880-L945

Would probably make sense for bootupd to use the same heuristic (or if it finds a better heuristic to use, update coreos-installer to match).

travier:
Notes for implementing this for Fedora CoreOS:
- https://github.com/coreos/fedora-coreos-tracker/issues/1485#issuecomment-1533350911
- https://github.com/coreos/fedora-coreos-tracker/issues/1752#issuecomment-2194950462

But unfortunately we might have to add more heuristics to support Anaconda installations.

travier:
As far as I could see when tried various RAID setups in Anaconda (https://docs.fedoraproject.org/en-US/fedora/f36/install-guide/install/Installing_Using_Anaconda/#sect-installation-gui-manual-partitioning-swraid), they only create a single EFI system partition.

So Fedora CoreOS is the only case with one per disk.

See also: https://gitlab.com/fedora/ostree/sig/-/issues/1#note_2086744541

HuijingHei:
Test failed with multiple EFIs in mirrored setups (using latest bootupd in copr build):

```
$ cosa run --add-disk 10G -B ../ign/mirror.bu --qemu-firmware uefi -m 4096
[root@cosa-devsh ~]# mount -o remount,rw /boot
[root@cosa-devsh ~]# mv /boot/bootupd-state.json .
[root@cosa-devsh ~]# bootupctl adopt-and-update
Running as unit: bootupd.service
lsblk: : not a block device
error: Failed to run lsblk
```

[...]

PR #813, opened 2025-01-09
Support updating multiple EFIs in mirrored setups (RAID1)

HuijingHei:
Add blockdev.rs to parse block device

---
bios.rs: update some functions to use blockdev
- Remove `get_device()`
- Update `get_bios_boot_partition()` to use blockdev
---

The EFI System Partition is not mounted after booted, on systems
configured with boot device mirroring, there are independent EFI
partitions on each constituent disk, need to mount each disk and
updates. But skip updating BIOS in this case.

Xref to https://github.com/coreos/bootupd/issues/132

### DIFF
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -6,7 +6,6 @@ use crate::blockdev;
 use crate::component::*;
 use crate::model::*;
 use crate::packagesystem;
-use crate::util;
 
 use anyhow::{bail, Result};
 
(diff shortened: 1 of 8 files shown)


travier:
AFAIK, this is only for EFI setups and not BIOS ones. I don't know if we mirror the MBR as well for BIOS RAID setups in FCOS.

HuijingHei:
> AFAIK, this is only for EFI setups and not BIOS ones. I don't know if we mirror the MBR as well for BIOS RAID setups in FCOS.

I think you are right, let's skip BIOS first, can add it later if needed.

travier:
Acording to Hristo on the fcos channel, we do mirror the MBR for RAID mirror setups with BIOS. We can do that later as we mostly care about the UEFI case for now.

cgwalters:
I only gave this a superficial skim, it seems sane (nice work!! and thank you for working on this!) but it's a big change and merits a deeper review. I will try to get some time.

In the intermediate time, can you split off the first two prep commits to a separate PR?

HuijingHei:
> can you split off the first two prep commits to a separate PR?

Sure, the PR is https://github.com/coreos/bootupd/pull/820

HuijingHei:
Close this as did in https://github.com/coreos/bootupd/pull/855

PR #817, opened 2025-01-17
Release 0.2.26

cgwalters:
Nothing huge here, but we want to keep up regular releases in general. I specifically want the fix for dropping the RPM dep.


HuijingHei (approved):
LGTM

PR #820, opened 2025-01-20
Add `blockdev` and update bios.rs

HuijingHei:
Prep for https://github.com/coreos/bootupd/issues/132

---
Add `bootc-blockdev` and local crate `blockdev`
See https://github.com/containers/bootc/pull/1050

---
blockdev.rs: Add `#[allow(dead_code)]` for the unused functions

---
bios.rs: update functions to use blockdev

### DIFF
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -1,68 +1,22 @@
 use std::io::prelude::*;
+use std::os::unix::io::AsRawFd;
 use std::path::Path;
 use std::process::Command;
 
+use crate::blockdev;
 use crate::component::*;
 use crate::model::*;
 use crate::packagesystem;
-use anyhow::{bail, Result};
 
-use crate::util;
-use serde::{Deserialize, Serialize};
+use anyhow::{bail, Result};
 
 // grub2-install file path
 pub(crate) const GRUB_BIN: &str = "usr/sbin/grub2-install";
-
-#[derive(Serialize, Deserialize, Debug)]
-struct BlockDevice {
-    path: String,
-    pttype: Option<String>,
-    parttypename: Option<String>,
-}
-
-#[derive(Serialize, Deserialize, Debug)]
-struct Devices {
-    blockdevices: Vec<BlockDevice>,
-}
 
 #[derive(Default)]
 pub(crate) struct Bios {}
 
 impl Bios {
-    // get target device for running update
-    fn get_device(&self) -> Result<String> {
-        let mut cmd: Command;
-        #[cfg(target_arch = "x86_64")]
-        {
-            // find /boot partition
-            cmd = Command::new("findmnt");
-            cmd.arg("--noheadings")
-                .arg("--nofsroot")
-                .arg("--output")
-                .arg("SOURCE")
-                .arg("/boot");
-            let partition = util::cmd_output(&mut cmd)?;
-
-            // lsblk to find parent device
-            cmd = Command::new("lsblk");
-            cmd.arg("--paths")
-                .arg("--noheadings")
-                .arg("--output")
-                .arg("PKNAME")
-                .arg(partition.trim());
-        }
-
-        #[cfg(target_arch = "powerpc64")]
-        {
-            // get PowerPC-PReP-boot partition
-            cmd = Command::new("realpath");
-            cmd.arg("/dev/disk/by-partlabel/PowerPC-PReP-boot");
-        }
-
-        let device = util::cmd_output(&mut cmd)?;
-        Ok(device)
-    }
-
     // Return `true` if grub2-modules installed
     fn check_grub_modules(&self) -> Result<bool> {
         let usr_path = Path::new("/usr/lib/grub");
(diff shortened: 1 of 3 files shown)


src/blockdev.rs
```diff
@@ -0,0 +1,208 @@
...
+        if part.is_bios_boot_part() {
+            return Ok(Some(part.path));
+        }
+    }
+    log::debug!("Not found any bios_boot partition");
+    Ok(None)
+}
+
+/// Find all bios_boot partitions on the backing devices with mountpoint boot
+#[allow(dead_code)]
```
cgwalters, line 177: I'm fine with this as is but I think it'd be slightly cleaner if this was part of the second patch

src/blockdev.rs
```diff
@@ -0,0 +1,208 @@
+use std::collections::HashMap;
```
cgwalters, line 1: Not a blocker here but maybe we should try to sync this code with https://github.com/containers/bootc/blob/main/lib/src/blockdev.rs ?

I could try to refactor that into a separate internal crate that we sync back into bootupd and coreos-installer?

Or of course we could even try to publish it.

src/bios.rs
```diff
@@ -199,9 +133,13 @@ impl Component for Bios {
             anyhow::bail!("Failed to find adoptable system")
         };
 
-        let device = self.get_device()?;
-        let device = device.trim();
-        self.run_grub_install("/", device)?;
+        let target_root = "/";
+        let devices = blockdev::get_backing_devices(&target_root)?
```
cgwalters, line 137: How about adding a helper function like `blockdev::get_single_backing_device()` that errors out if it finds > 1 (and also errors out if we somehow get zero).

src/blockdev.rs
HuijingHei: > Not a blocker here but maybe we should try to sync this code with https://github.com/containers/bootc/blob/main/lib/src/blockdev.rs ?
> 
> I could try to refactor that into a separate internal crate that we sync back into bootupd and coreos-installer?
> 
> Or of course we could even try to publish it.

Thanks, that would be better and make this easier. 
I tried to copy the `bootc/blockdev.rs`, but seems there are a lot of dependencies that also need to sync, for example `Task` and `bootc_utils::CommandRunExt`, so I give up.

cgwalters:
FYI I did https://github.com/containers/bootc/pull/1048 to try to prepare bootc's `blockdev.rs` for reuse, but again nonblocking

src/blockdev.rs
cgwalters: The idea of https://github.com/coreos/bootupd/pull/820#issuecomment-2603461521 is that the Task dependency is gone. However yes, we'd need to pull in `bootc-utils` too. We could even do that now.

OK I did just that in https://github.com/coreos/bootupd/pull/821

HuijingHei: Cool, thanks!

HuijingHei: Can we import https://github.com/containers/bootc/blob/main/lib/src/blockdev.rs like what you did in https://github.com/coreos/bootupd/pull/821 ?

src/bios.rs
```diff
@@ -215,9 +153,14 @@ impl Component for Bios {
 
     fn run_update(&self, sysroot: &openat::Dir, _: &InstalledContent) -> Result<InstalledContent> {
         let updatemeta = self.query_update(sysroot)?.expect("update available");
-        let device = self.get_device()?;
-        let device = device.trim();
-        self.run_grub_install("/", device)?;
+        let sysroot = sysroot.recover_path()?;
+        let dest_root = sysroot.to_str().unwrap_or("/");
+        let devices = blockdev::get_backing_devices(&dest_root)?
```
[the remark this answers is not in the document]
HuijingHei: Sorry, seems this is a little confusing, `findmnt` find the source of mount point, for example on RAID1 is `/dev/md127`, what we want is call `find_parent_devices()` to get the parent device.

src/bios.rs
```diff
@@ -215,9 +153,14 @@ impl Component for Bios {
 
     fn run_update(&self, sysroot: &openat::Dir, _: &InstalledContent) -> Result<InstalledContent> {
         let updatemeta = self.query_update(sysroot)?.expect("update available");
-        let device = self.get_device()?;
-        let device = device.trim();
-        self.run_grub_install("/", device)?;
+        let sysroot = sysroot.recover_path()?;
```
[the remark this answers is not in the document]
HuijingHei: Make sense, update to use `format!("/proc/self/fd/{}", sysroot.as_raw_fd())`, is this that you meant?

src/bios.rs
HuijingHei: Update to use `blockdev::get_single_device()`.

src/blockdev.rs
cgwalters: https://github.com/containers/bootc/pull/1050

HuijingHei:
Close this as did in https://github.com/coreos/bootupd/pull/824

PR #821, opened 2025-01-21
Use bootc-utils for command helpers

cgwalters:
This enables more code sharing. I didn't do a full port here, but this starts the ball rolling. In particular I think `run_and_parse_json()` is nice and elegant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler2"
+version = "2.0.0"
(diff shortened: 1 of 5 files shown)


HuijingHei (approved):
LGTM, so fast.

HuijingHei:
Maybe in future, also refactor the `CommandRunExt` into a separate internal crate and publish?

travier:
Are we planning to publish this as an independent crate on crates.io? Otherwise I can see this causing problem with packaging in Fedora unless we vendor things which we don't do generally.

cgwalters:
> Are we planning to publish this as an independent crate on crates.io?

Maybe.

> Otherwise I can see this causing problem with packaging in Fedora unless we vendor things which we don't do generally.

Some projects for sure do, I think as you know I personally disagree with the approach for about 60% of the Fedora Rust packaging. And bootupd does vendor consistently across both Fedora+RHEL right now: https://src.fedoraproject.org/rpms/rust-bootupd/blob/rawhide/f/rust-bootupd.spec#_13

travier:
I disagree as well with the no-vendoring approach. I hadn't checked that we already vendor the code so that should not be an issue. Thanks

PR #834, opened 2025-02-05
Release 0.2.27 & Cargo.toml: Add homepage & Revert "Use bootc..."

travier:
Revert "Use bootc-utils for command helpers"

Temporarily revert until we publish this crate.

See: https://github.com/coreos/bootupd/issues/833

This reverts commit 7f46422223bad1b95d9f100576d335a8aabcb9bf.

---

Cargo.toml: Add homepage

---

Release 0.2.27

This release introduces the hidden `migrate-static-grub-config`
subcommand to help migrate system currently using a dynamically
generated GRUB config to a static one.

This is hidden command as it should only be use by distributions to
enable that transition as needed via the bootloader update unit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ dependencies = [
 
 [[package]]
 name = "bootupd"
-version = "0.2.26"
+version = "0.2.27"
 dependencies = [
  "anyhow",
  "bincode",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,21 +3,6 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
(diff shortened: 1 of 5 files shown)


cgwalters:
> Temporarily revert until we publish this crate.

Why? I think we should stop publishing bootupd to crates.io, it makes no sense to do.

travier:
> > Temporarily revert until we publish this crate.
> 
> Why? I think we should stop publishing bootupd to crates.io, it makes no sense to do.

OK, then we need to update our release process instead.

cgwalters:
I did hit an issue with vendoring and git dependencies in bootc, see https://github.com/containers/bootc/pull/1068/commits/58fa21e54fa6868d5a266072b822454536398ef9

I think it'd be possible to apply the same thing here.

PR #835, opened 2025-02-05
Cargo.toml: Add homepage

travier:
Fixes a warning when preparing the release.

---

Will also serve as a test for the failure in https://github.com/coreos/bootupd/pull/834

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -7,7 +7,7 @@ permissions:
 
 jobs:
   crossarch-check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     name: Build on ${{ matrix.arch }}
 
     strategy:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ version = "0.2.26"
 authors = ["Colin Walters <walters@verbum.org>"]
 edition = "2021"
 rust-version = "1.75.0"
+homepage = "https://github.com/coreos/bootupd"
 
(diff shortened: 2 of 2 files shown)


cgwalters (approved):
I'm fine with this, however if we want to switch over to not publishing, that should squash warnings like this.
To do so, use e.g. https://github.com/containers/bootc/blob/5e0df36305cfda93eab0ca70678b618d1f2844f0/cli/Cargo.toml#L8

travier:
I started looking at what it would take us to stop publishing and do our own releases but that's a significant amount of changes and I would prefer to not block the 0.2.27 release on that. Another option is to backport just https://github.com/coreos/bootupd/pull/790 into Fedora for now.

travier:
Added a commit to temporarily build using Ubuntu-22.04 runners as the builds regularly fails on the 24.04 ones.

.github/workflows/cross.yml
```diff
@@ -7,7 +7,7 @@ permissions:
 
 jobs:
   crossarch-check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
cgwalters, line 10: Did you mean 24.04 ?

travier: Latest is now 24.04. This downgrades to 22.04 to make the tests pass. It's similar to build issues that we've seen in other repos: https://github.com/coreos/afterburn/issues/1156

I have not figured out where the failure is exactly.

Maybe we should run this cross-build on the new aarch64 runners as that should be faster as well.

cgwalters (approved):

PR #838, opened 2025-02-06
contrib/packaging/spec: Sync with Fedora changes

travier:
github/workflows/cross: Use ubuntu-22.04 runners

---

contrib/packaging/spec: Sync with Fedora changes

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -7,7 +7,7 @@ permissions:
 
 jobs:
   crossarch-check:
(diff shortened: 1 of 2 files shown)


travier:
Hum, for some reason, the systemd unit is not included in the crate tarball (the entire contrib directory is not there).

cgwalters:
> (the entire contrib directory is not there).

Add `contrib` to

include = ["src", "LICENSE", "Makefile", "systemd"] in `cargo.toml`

PR #843, opened 2025-02-11
Drop dbg

cgwalters:
rpm_cmd: Drop leftover dbg!

Oops.

---

Deny `clippy::dbg_macro`

So we don't get more `dbg!` sneaking in.

Signed-off-by: Colin Walters <walters@verbum.org>

---


jmarrero (approved):
lgtm

PR #839, opened 2025-02-06
bios+ppc64le: Support being passed a whole device

cgwalters:
And automatically find the PReP partition.

Closes: https://github.com/coreos/bootupd/issues/837

PR #844, opened 2025-02-11
lib: Update to bootc 1.1.5

cgwalters:
On general principle but this is also prep for
https://github.com/coreos/bootupd/pull/839


jmarrero (approved):
lgtm

PR #845, opened 2025-02-11
rpm_cmd: Also drop debugging `ls`

cgwalters:
This was another debugging thing that got accidentally committed.


jmarrero (approved):
lgtm

PR #846, opened 2025-02-11
Rework vendoring

cgwalters:
xtask: Fix deprecated `default_features`

cargo warns about this.

---

xtask: Sync with bootc

In particular now that we have a `git` dependency, this turns
out to reveal some deficiencies in how e.g. the Fedora RPM
vendoring support works.

Sync the code from https://github.com/containers/bootc/pull/1068/commits/58fa21e54fa6868d5a266072b822454536398ef9

---

contrib/rpm: Sync vendor config from bootc

---

contrib/rpm: Two more fixes

- Use `rust-%{crate}` for name because that's what Fedora
  does; the previous spec file sync was just broken
- Use a source archive generated from git, not a `.crate`; we
  will stop publishing to crates.io.

Signed-off-by: Colin Walters <walters@verbum.org>

---

### DIFF
--- a/contrib/packaging/bootupd.spec
+++ b/contrib/packaging/bootupd.spec
@@ -2,14 +2,14 @@
 
 %global crate bootupd
 
-Name:           bootupd
+Name:           rust-%{crate}
 Version:        0.2.9
 Release:        1%{?dist}
 Summary:        Bootloader updater
 
 License:        Apache-2.0
 URL:            https://github.com/coreos/bootupd
-Source0:        %{url}/releases/download/v%{version}/bootupd-%{version}.crate
+Source0:        %{url}/releases/download/v%{version}/bootupd-%{version}.tar.zstd
 Source1:        %{url}/releases/download/v%{version}/bootupd-%{version}-vendor.tar.zstd
 ExcludeArch:    %{ix86}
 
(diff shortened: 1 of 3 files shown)


contrib/packaging/bootupd.spec
```diff
@@ -2,14 +2,14 @@
 
 %global crate bootupd
 
-Name:           bootupd
+Name:           rust-%{crate}
```
jmarrero, line 5: this will not create a big issue in Fedora/RHEL? Changing the package name?

jmarrero (approved):

contrib/packaging/bootupd.spec
cgwalters: This is just the *source* name, and it's already named this in Fedora https://src.fedoraproject.org/rpms/rust-bootupd but it isn't in RHEL https://gitlab.com/redhat/centos-stream/rpms/bootupd

cgwalters: See https://github.com/coreos/bootupd/pull/838 which introduced this change

cgwalters: IOW, one pain point we are going to need to eat here is that because it'd be hugely painful (AFAIK) to rename the *source* package in RHEL, we will have spec file divergence for the forseeable future there.

PR #848, opened 2025-02-12
contrib/spec: Add BR on git

cgwalters:
Since we use autosetup -Sgit


jmarrero (approved):
lgtm

travier:
It should probably be `git-core` instead: https://discussion.fedoraproject.org/t/f42-change-proposal-switch-to-git-core-self-contained/138975

travier:
https://github.com/coreos/bootupd/pull/850

PR #855, opened 2025-02-18
efi: support updating multiple EFIs in mirrored setups (RAID1)

HuijingHei:
The prep PR is https://github.com/coreos/bootupd/pull/824

Fixes https://github.com/coreos/bootupd/issues/132


src/efi.rs
```diff
@@ -93,11 +73,26 @@ impl Efi {
...
-    pub(crate) fn ensure_mounted_esp(&self, root: &Path) -> Result<PathBuf> {
-        let mut mountpoint = self.mountpoint.borrow_mut();
+    // Get esp device list on all devices
+    fn get_esp_devices(&self) -> Option<Vec<String>> {
+        let mut esp_devices = vec![];
+        if let Some(esp_device) = self.get_esp_device() {
+            esp_devices.push(esp_device.to_string_lossy().into_owned());
+        } else {
+            esp_devices = blockdev::find_colocated_esps("/").expect("get esp devices");
+        };
```
champtar, line 83: Have you considered / tried the case with 2 disks / 2 separate OS ? (we do that in our lab, 2 or even 3 rpm-ostree based OS, each one on one disk, all installed via anaconda)
I'm asking because quickly looking at the code, I would use `find_colocated_esps("/")` first and `get_esp_device()` as a fallback.

The safest might be:
1) mounted ESP(s)
2) colocated ESP(s)
3) get_esp_device

If you have a bootc installed OS, and on a second disk an anaconda installed OS (not even ostree based), right now you are going to always pick the anaconda ESP

HuijingHei: > I would use find_colocated_esps("/") first and get_esp_device() as a fallback.

`get_esp_devices()` includes:
- `get_esp_device()` will find CoreOS ESP label `/dev/disk/by-partlabel/EFI-SYSTEM`, or Anaconda ESP label `/dev/disk/by-partlabel/EFI\\x20System\\x20Partition`
- `find_colocated_esps("/") `will list ESP partitions on the devices with mountpoint `/boot`, but it does not mean it is always mounted, for example, coreos does not mount esp after booted, see [doc](https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_mounted_filesystems)

So I think it is OK to keep it as it is?

HuijingHei: > The safest might even be:
> 
> mounted ESP(s)
> colocated ESP(s)
> get_esp_device

What I did: when running update, `get_esp_devices()` would list the all esp devices, then in the loop device will check mounted ESP first, if not, will mount the esp then upgrade. I think you meant that need to check mounted firstly in the beginning, will try that, thanks!

HuijingHei: I do not have such env, but IMU, if second OS (sdb) booted, sda will not be mounted from the booted OS, in this case, we do not scan sda.

> sda: normal anaconda install
> sdb: ESP not mounted, not using anaconda/coreos label

skip sda, and `get_esp_devices()` will call `find_colocated_esps("/")` to find the mount point /boot device, then get ESP part on the same device

> sda: normal anaconda install
> sdb & sdc: raid 1 mirrored, ESP not mounted, not using anaconda/coreos label

skip sda, this is the case that we need to resolve actually, `find_colocated_esps("/")` to find the mount point `/boot` device which is `/dev/md126`, then need to get the backing devices list `sdb & sdc` and find ESP part on each device

> sda: normal coreos install
> sdb: also a coreos install, ESP not mounted

When I tried to boot VM with 2 coreos, failed with:
```
[    9.373774] rdcore[957]: Error: System has 2 devices with a filesystem labeled 'boot': ["/dev/vdb3", "/dev/vda3"]
[FAILED] Failed to start CoreOS Ignition Ensure Unique Boot Filesystem.
```
So we can skip this env.

HuijingHei: cc @travier to confirm about my [reply](https://github.com/coreos/bootupd/pull/855#discussion_r2000542257), not sure if my understanding is correct, thanks!

HuijingHei: > All 3 wants to have the by-partlabel symlink

If booted sda, then it will find ESP part in sda, I do not think it will retrieve all the devices.

HuijingHei:
@travier could you help to review when you have time? Thanks!

travier:
Dusty added some logic in https://github.com/openshift/os/pull/1795 about making sure that the ESP that we are finding are part of the same device that the one the /boot partition is in and we should do the same for both the RAID and non RAID case.

HuijingHei:
> Dusty added some logic in [openshift/os#1795](https://github.com/openshift/os/pull/1795) about making sure that the ESP that we are finding are part of the same device that the one the /boot partition is in and we should do the same for both the RAID and non RAID case.

Agree, and what I did in this patch followed the same way.

src/efi.rs
```diff
@@ -267,16 +272,32 @@ impl Component for Efi {
...
-        let diff = updatef.relative_diff_to(&esp)?;
-        log::trace!("applying adoption diff: {}", &diff);
-        filetree::apply_diff(&updated, &esp, &diff, None).context("applying filesystem changes")?;
+        let esp_devices = self
+            .get_esp_devices()
+            .expect("get esp devices before adopt");
+        let sysroot = sysroot.recover_path()?;
+
+        for esp_dev in esp_devices {
+            let dest_path = if let Some(dest_path) = self.check_mounted_esp(&sysroot)? {
```
dustymabe, line 285: I don't really understand why we needed to create a new `check_mounted_esp()` function (i.e. breaking up `ensure_mounted_esp()`). If we do really need this maybe we can split it into a separate commit with rationale?

IMO now when we call `ensure_mounted_esp(&sysroot, &esp_dev)` we should ensure that the `esp_dev` we passed in is the one that is mounted on the mountpoint. i.e. if somehow a different ESP is mounted there then we shouldn't be happy about it.

src/efi.rs
```diff
@@ -392,24 +434,37 @@ impl Component for Efi {
     }
 
     fn validate(&self, current: &InstalledContent) -> Result<ValidationResult> {
-        if !is_efi_booted()? && self.get_esp_device().is_none() {
+        let esp_devices = self.get_esp_devices();
```
dustymabe, line 437: we are calling `get_esp_devices()` multiple times. Perhaps we should run it once and store the data so that we aren't going through the same logic (and system calls) over and over again?

src/efi.rs
```diff
@@ -267,16 +272,32 @@ impl Component for Efi {
...
+
+        for esp_dev in esp_devices {
+            let dest_path = if let Some(dest_path) = self.check_mounted_esp(&sysroot)? {
+                dest_path.join("EFI")
+            } else {
+                self.ensure_mounted_esp(&sysroot, &esp_dev)?.join("EFI")
+            };
+
+            let esp = openat::Dir::open(&dest_path).context("opening EFI dir")?;
+            validate_esp(&esp)?;
```
dustymabe, line 292: suggest to rename the function `validate_esp_fstype()`.

src/efi.rs
```diff
@@ -137,6 +142,7 @@ impl Efi {
         if let Some(mount) = self.mountpoint.borrow_mut().take() {
             Command::new("umount")
                 .arg(&mount)
+                .arg("-l")
```
dustymabe, line 145: I still can't really think of why this would be needed.

The only reason I can think of is the buffers hadn't been flushed yet, which maybe what we should do instead is run a `sync` after `apply_diff` rather than immediately unmounting and only `umount` after the `sync` is done.

src/efi.rs
```diff
@@ -392,24 +434,37 @@ impl Component for Efi {
...
+
         let mut errs = Vec::new();
-        for f in diff.changes.iter() {
-            errs.push(format!("Changed: {}", f));
-        }
-        for f in diff.removals.iter() {
-            errs.push(format!("Removed: {}", f));
+        let esps = esp_devices.ok_or_else(|| anyhow::anyhow!("No esp device found!"))?;
+        let dest_root = Path::new("/");
+        for esp_dev in esps.iter() {
```
dustymabe, line 449: Since we now have multiple places where we basically need to iterate over esp_devices and mount them and then `do_some_work()` I wish there was an easy way just to make that a funcion and then `yield` like we can in `python` so that we don't have to copy/paste this same code in many places.

Do you know of way to do that? I did some searching around rust book and the net and didn't find an easy answer.

src/efi.rs
HuijingHei: Agree that this makes sense, sorry for the above misunderstanding.
Can I do some check that if there is only 1 disk, then would call `get_esp_device()` ?

dustymabe: > Can I do some check that if there is only 1 disk, then would call get_esp_device() ?

You could, but what would be the benefit?

I could see it being beneficial to make that optimization (and thus keep two separate code paths) if the performance was much different between one versus the other and this was an operation that ran often. In this case, though, this isn't something that runs often and I think it would be much better to just have a single code path that we maintain if it can satisfy all of our requirements.

HuijingHei: SGTM, will update, thanks!

src/efi.rs
HuijingHei: Updated, thanks!

src/efi.rs
HuijingHei: > Can we just say if /boot isn't mounted we'll find the parent devices of / (or target_root) and use that instead?

Check that `findmnt -J -v --output=SOURCE,FSTYPE,OPTIONS,UUID /` will get `"source": "composefs"` (which is not real device), would use `/sysroot` instead and get `"source": "/dev/vda4"`

src/efi.rs
HuijingHei: Sorry that I should check more about this, see that there is `syncfs` after `apply_diff` (see https://github.com/coreos/bootupd/blob/main/src/filetree.rs#L402), but failed with 
```
umount: /boot/efi: target is busy.
[DEBUG bootupd::efi] Unmounting
error: Failed adopt and update: unmount after adopt: Failed to unmount "/boot/efi": Child ["umount" "/boot/efi"] exited: exit status: 32
```

If append `-l` after `umount`, no error, so I think we should keep it, WDYT?

src/efi.rs
HuijingHei: They are used in different subcommands, but agree that need to make the changes to store the data about <parent devices>, <bios_boot partitions>, <esp partitions>, a little tricky that they should be stored at runtime, any suggestions for this?

```
# mount -o remount,rw /boot && rm -f /boot/bootupd-state.json
# bootupctl update -vvvv
[TRACE bootupd] executing cli
[TRACE bootupd] executing cli
[TRACE bootupd::bootupd] No saved state
[TRACE bootupd::bootupd] Remaining known components: 2
[DEBUG bootupd::blockdev] Find parent devices: ["/dev/vda"]
[DEBUG bootupd::blockdev] Find bios_boot partitions: ["/dev/vda1"]
[TRACE bootupd::component] Adoptable: ContentMetadata { timestamp: 2025-04-22T07:11:03.919Z, version: "42.20250422.20.0" }
[DEBUG bootupd::blockdev] Find parent devices: ["/dev/vda"]
[DEBUG bootupd::blockdev] Find esp partitions: ["/dev/vda2"]
[TRACE bootupd::efi] No efivars mount at "/sys/firmware/efi/efivars"
[TRACE bootupd::efi] No efivars mount at "/sys/firmware/efi/efivars"
[TRACE bootupd::component] Adoptable: ContentMetadata { timestamp: 2025-04-22T07:11:03.919Z, version: "42.20250422.20.0" }
[DEBUG bootupd::efi] Unmounting
[DEBUG bootupd::blockdev] Find parent devices: ["/dev/vda"]
[DEBUG bootupd::blockdev] Find bios_boot partitions: ["/dev/vda1"]
[TRACE bootupd::component] Adoptable: ContentMetadata { timestamp: 2025-04-22T07:11:03.919Z, version: "42.20250422.20.0" }
[DEBUG bootupd::blockdev] Find parent devices: ["/dev/vda"]
[DEBUG bootupd::bios] Install grub modules on /dev/vda
Adopted and updated: BIOS: grub2-tools-1:2.12-28.fc42.x86_64

[DEBUG bootupd::blockdev] Find parent devices: ["/dev/vda"]
[DEBUG bootupd::blockdev] Find esp partitions: ["/dev/vda2"]
[TRACE bootupd::efi] No efivars mount at "/sys/firmware/efi/efivars"
[TRACE bootupd::efi] No efivars mount at "/sys/firmware/efi/efivars"
[TRACE bootupd::component] Adoptable: ContentMetadata { timestamp: 2025-04-22T07:11:03.919Z, version: "42.20250422.20.0" }
[DEBUG bootupd::blockdev] Find parent devices: ["/dev/vda"]
[DEBUG bootupd::blockdev] Find esp partitions: ["/dev/vda2"]
[DEBUG bootupd::efi] Mounted at "/boot/efi"
[TRACE bootupd::efi] applying adoption diff: additions: 0 removals: 0 changes: 0
[TRACE bootupd::efi] Unmounted
[DEBUG bootupd::efi] Unmounting
Adopted and updated: EFI: grub2-efi-x64-1:2.12-28.fc42.x86_64,shim-x64-15.8-3.x86_64
```

HuijingHei: Do experiments like [change](https://github.com/coreos/bootupd/compare/main...HuijingHei:bootupd:get-devices?expand=1), initialize the parent devices at the beginning of `update` or `adopt_and_update`. Not sure if this is what we expect.

dustymabe:
Is this ready for another round of review or are we waiting to rebase once https://github.com/coreos/bootupd/pull/923 merges?

HuijingHei:
> Is this ready for another round of review or are we waiting to rebase once #923 merges?

Thank you for the asking! Yes, would be better to wait for #923 merge.

HuijingHei:
Need https://github.com/coreos/bootupd/pull/932 merge first.

src/blockdev.rs
```diff
@@ -5,20 +5,29 @@ use anyhow::{bail, Context, Result};
...
+    // Fallback to /sysroot if the command fails
+    let source = if let Ok(fsinfo) = crate::filesystem::inspect_filesystem(&bootdir, ".") {
+        fsinfo.source
+    } else {
+        let sysroot = target_root.join("sysroot");
+        let sysrootdir = openat::Dir::open(&sysroot)
+            .with_context(|| format!("Opening sysroot {}", sysroot.display()))?;
+        let fsinfo = crate::filesystem::inspect_filesystem(&sysrootdir, ".")?;
+        fsinfo.source
+    };
```
dustymabe, line 30: Would everything above this point in the function be better served as a for loop that loops over `boot` and then `sysroot`. If something is found for `boot` it will exit the loop early.

src/util.rs
```diff
@@ -5,6 +5,7 @@ use std::process::Command;
 use anyhow::{bail, Context, Result};
 use openat_ext::OpenatDirExt;
 
+#[allow(dead_code)]
 pub(crate) trait CommandRunExt {
     fn run(&mut self) -> Result<()>;
 }
```
dustymabe: Why not delete?

src/bios.rs
```diff
@@ -167,18 +167,11 @@ impl Component for Bios {
...
-            anyhow::bail!("Failed to find parent device");
-        };
-
-        if let Some(next) = parent_devices.next() {
-            anyhow::bail!(
-                "Found multiple parent devices {parent} and {next}; not currently supported"
-            );
+        for parent in rootcxt.devices.iter() {
+            self.run_grub_install(rootcxt.path.as_str(), &parent)?;
+            log::debug!("Install grub modules on {parent}");
```
dustymabe: Maybe a bad merge? `Install -> Installed`

src/bios.rs
```diff
@@ -195,20 +188,11 @@ impl Component for Bios {
...
```
dustymabe: `Install -> Installed`

HuijingHei: Updated, thank you!

src/util.rs
HuijingHei: Removed, thank you.

src/blockdev.rs
HuijingHei: SGTM, looks more neat, thank you!

dustymabe:
I can maybe run a more complicated test on Monday, but this is looking good to me.

dustymabe:
> One thing that might be an enhancement over what we have here is some indication in the normal output `bootupctl status` and `bootupctl update` that the software detects there are multiple devices under its control. i.e. when I run `bootupctl status` today I don't know from the output that it recognizes that or not.

This can be a followon PR though if we decide it's a good idea.

HuijingHei:
> > One thing that might be an enhancement over what we have here is some indication in the normal output `bootupctl status` and `bootupctl update` that the software detects there are multiple devices under its control. i.e. when I run `bootupctl status` today I don't know from the output that it recognizes that or not.
> 
> This can be a followon PR though if we decide it's a good idea.

SGTM, thank you so much for the testing.

Issue #856, opened 2025-02-18
Do not try to copy a dir in the ESP if it does not exists


travier:
It looks like we always try to copy the dir first without checking if it exists during the update.

```
Feb 16 14:31:35 mershl-t495 systemd[1]: Starting bootloader-update.service - Update bootloader on boot...
Feb 16 14:31:36 mershl-t495 bootupctl[1110]: cp: cannot stat 'BOOT': No such file or directory
Feb 16 14:31:36 mershl-t495 bootupctl[1065]: error: Failed adopt and update: applying filesystem changes: Failed to copy BOOT to .btmp.BOOT
Feb 16 14:31:36 mershl-t495 systemd[1]: bootloader-update.service: Main process exited, code=exited, status=1/FAILURE
Feb 16 14:31:36 mershl-t495 systemd[1]: bootloader-update.service: Failed with result 'exit-code'.
Feb 16 14:31:36 mershl-t495 systemd[1]: Failed to start bootloader-update.service - Update bootloader on boot.
```

See: https://discussion.fedoraproject.org/t/sb42-updates-fail-on-system-with-512mb-boot-128mb-boot-efi/145041

HuijingHei:
You are correct, we should check the dir before dong the copy.

PR #858, opened 2025-02-20
update: do not try to copy a dir in the ESP if it does not exists

HuijingHei:
When doing removals, usually copy direcory to temp direcoty, then
do operations in temp dir, but should check if the dir is existed
 in dest ESP.

Fixes https://github.com/coreos/bootupd/issues/856


cgwalters (approved):
It'd be helpful to have a unit test that this patch fixes. This code is complex and I can't say I fully convinced myself this logic is right, but it looks plausible to me.

PR #894, opened 2025-04-02
Sync repo templates ⚙

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/repo-templates/actions/workflows/sync.yml) ([source](https://github.com/coreos/repo-templates/blob/main/.github/workflows/sync.yml)).

Sync with coreos/repo-templates@13cccc6879d53039e593e3263c68d0e5d575a400.


HuijingHei:
We also need to update rust MSRV to 1.84.1, will do this later.

HuijingHei (approved):

PR #895, opened 2025-04-03
Cargo.toml: bump MSRV to 1.84.1

HuijingHei:
Simlar to https://github.com/coreos/coreos-installer/commit/948e9367b5401516fddfe73056662091b7471b00

update rust lint_toolchain to 1.84.1 in https://github.com/coreos/bootupd/pull/894

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -35,5 +35,9 @@ jobs:
           run: |
             set -xeu
             apt update -y
-            apt install -y gcc make cargo libssl-dev pkg-config
+            apt install -y gcc make curl libssl-dev pkg-config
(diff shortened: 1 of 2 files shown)


HuijingHei:
on ppc64:

error: package `bootupd v0.2.27 (/home/runner/work/bootupd/bootupd)` cannot be built because it requires rustc 1.84.1 or newer, while the currently active rustc version is 1.75.0

cgwalters:
Yeah we have too many CI systems with too many build settings...I think doing this will require using the GHA to install the MSRV rust version in our GHA flows.

jcapiitao (approved):

Issue #902, opened 2025-04-09
bootupctl: Failed to start transient service unit: Connection reset by peer


dustymabe:
When upgrading a 4.18 (RHEL 9.4) system to 4.19 (RHEL 9.6) after the update bootupctl status can't run:

```
[core@cosa-devsh ~]$ rpm -q bootupd
bootupd-0.2.27-3.el9.x86_64
[core@cosa-devsh ~]$ sudo bootupctl status 
Failed to start transient service unit: Connection reset by peer
```

The transition here is:

```
bootupd 0.2.18-1.el9 -> 0.2.27-3.el9
```

The test I did was boot `rhcos-418.94.202504080525-0-qemu.x86_64.qcow2` and then upgrade to `419.96.202504090041-0`. After upgrade run status command and observe failure.

cgwalters:
I'm not immediately reproducing this on a rhel-bootc system upgraded from rhel-bootc:9.4 to rhel-bootc:9.6. Can you get a bit more info about this from the journal? Also try `env SYSTEMD_LOG_LEVEL=debug bootupctl status`

HuijingHei:
I start 418.94.202504072147-0 and upgrade to 419.96.202504090041-0, can not reproduce. 

```
[root@cosa-devsh core]# rpm-ostree rebase ostree-unverified-registry:<419.96.202504090041-0>
  bootupd 0.2.18-1.el9 -> 0.2.27-3.el9
```
After reboot

```
[root@cosa-devsh core]# bootupctl status
Running as unit: bootupd.service
Component BIOS
  Installed: grub2-tools-1:2.06-86.el9_4.2.x86_64
  Update: Available: grub2-tools-1:2.06-104.el9_6.x86_64
Component EFI
  Installed: grub2-efi-x64-1:2.06-86.el9_4.2.x86_64,shim-x64-15.8-4.el9_3.x86_64
  Update: Available: grub2-efi-x64-1:2.06-104.el9_6.x86_64,shim-x64-15.8-3.el9_4.x86_64
No components are adoptable.
CoreOS aleph version: 418.94.202504072147-0
Boot method: BIOS

[root@cosa-devsh core]# rpm -q bootupd
bootupd-0.2.27-3.el9.x86_64
[root@cosa-devsh core]# bootupctl update -vvvv
...
[root@cosa-devsh core]# bootupctl status
Running as unit: bootupd.service
Component BIOS
[...]

PR #906, opened 2025-04-16
bootupd: add `--pty` when stdin is a terminal

HuijingHei:
From Colin's pointer:
`systemd-run` breaks when forwarding a physical (`/dev/tty*` with
 label `tty_device_t`) as opposed to virtual console (`/dev/pts`
with label `user_devpts_t`).
When stdin is a terminal, we should definitely use `-t`.

Fixes https://github.com/coreos/bootupd/issues/902


HuijingHei:
Build and test on rhcos:

```
$ cosa run -c
[root@cosa-devsh work]# env SYSTEMD_LOG_LEVEL=debug systemd-run --pipe bootupctl status
Bus n/a: changing state UNSET → OPENING
sd-bus: starting bus by connecting to /run/dbus/system_bus_socket...
Bus n/a: changing state OPENING → AUTHENTICATING
Bus n/a: changing state AUTHENTICATING → HELLO
Sent message type=method_call sender=n/a destination=org.freedesktop.DBus path=/org/freedesktop/DBus interface=org.freedesktop.DBus member=Hello cookie=1 reply_cookie=0 siga
Got message type=method_return sender=org.freedesktop.DBus destination=:1.21 path=n/a interface=n/a member=n/a cookie=4294967295 reply_cookie=1 signature=s error-name=n/a ea
Bus n/a: changing state HELLO → RUNNING
Sent message type=method_call sender=n/a destination=org.freedesktop.systemd1 path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=StartTransiena
Bus n/a: changing state RUNNING → CLOSING
Failed to start transient service unit: Connection reset by peer
Bus n/a: changing state CLOSING → CLOSED

[root@cosa-devsh work]# bootupctl status
Press ^] three times within 1s to disconnect TTY.
Component BIOS
  Installed: grub2-tools-1:2.06-104.el9_6.x86_64
  Update: At latest version
Component EFI
  Installed: grub2-efi-x64-1:2.06-104.el9_6.x86_64,shim-x64-15.8-3.el9_4.x86_64
  Update: At latest version
No components are adoptable.
CoreOS aleph version: 419.96.202504150446-0
Boot method: BIOS
```

HuijingHei:
Run cosa without `-c`:
`$ cosa run --qemu-image data/rhcos-419.96.202504150446-0-qemu.x86_64.qcow2 -m 4096 --bind-ro ./,/run/work`

Then update bootupd (built within rhel) that includes the patch, run `bootupctl status`, seems also detect as terminal, see that we have `Press ^] three times within 1s to disconnect TTY.`.
IMU, if only running with `-c`, it would add `--pty`; without `-c`, it should not add `--pty`, is this right? Does this mean the condition in the patch is not correct?

```
[root@cosa-devsh work]# rpm-ostree usroverlay 
[root@cosa-devsh work]# rm -f /usr/libexec/bootupd && rm -f /usr/bin/bootupctl
[root@cosa-devsh work]# cp bootupd /usr/libexec/bootupd && ln -f /usr/libexec/bootupd /usr/bin/bootupctl
[root@cosa-devsh work]# bootupctl status
Press ^] three times within 1s to disconnect TTY.
Component BIOS
  Installed: grub2-tools-1:2.06-104.el9_6.x86_64
  Update: At latest version
Component EFI
  Installed: grub2-efi-x64-1:2.06-104.el9_6.x86_64,shim-x64-15.8-3.el9_4.x86_64
  Update: At latest version
No components are adoptable.
CoreOS aleph version: 419.96.202504150446-0
Boot method: BIOS
```

Without the patch, run `bootupctl status` without error:
```
[core@cosa-devsh ~]$ sudo bootupctl status
Component BIOS
  Installed: grub2-tools-1:2.06-104.el9_6.x86_64
  Update: At latest version
Component EFI
  Installed: grub2-efi-x64-1:2.06-104.el9_6.x86_64,shim-x64-15.8-3.el9_4.x86_64
  Update: At latest version
No components are adoptable.
CoreOS aleph version: 419.96.202504150446-0
Boot method: BIOS
```

cgwalters:
> Does this mean the condition (using nix::unistd::isatty(std::io::stdin().as_raw_fd())) is not correct?

That looks right to me offhand. (btw though I prefer rustix over nix personally, and in this case the [rustix api](https://docs.rs/rustix/latest/rustix/termios/fn.isatty.html) is slightly cleaner. But anyways that's a basically irrelevant aside.

> IMU, if only running with -c, it would add --pty; without -c, it should not add --pty, is this right? 

No, both physical and virtual consoles are ttys. I am not sure there's really a way to find out the difference short of scraping `/proc/self/fd/0` and seeing whether it's a symlink to `/dev/pts/<something>` or not.

cgwalters:
I think I was wrong, there's no real value in us allocating a new pty when we're already on a tty and doing so is just a workaround. So if the selinux fix is going to take some time, then we could do the above logic (detect physical tty and allocate a pty). Or we could just leave this as a known corner case issue that I think relatively few people will hit.

HuijingHei:
> So if the selinux fix is going to take some time, then we could do the above logic (detect physical tty and allocate a pty). Or we could just leave this as a known corner case issue that I think relatively few people will hit.

I am OK to leave this as it is and wait for https://issues.redhat.com/browse/RHEL-86925 is fixed, WDYT, @dustymabe ?

dustymabe:
> I am OK to leave this as it is and wait for https://issues.redhat.com/browse/RHEL-86925 is fixed, WDYT, @dustymabe ?

SGTM. Feel free to close this.

PR #916, opened 2025-04-21
support riscv64

dustymabe:
Allows update EFI bootloader on riscv64 similar to aarch64 and x86_64.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/1931

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ that's for tools like `grubby` and `ostree`.
 ## Status
 
 bootupd supports updating GRUB and shim for UEFI firmware on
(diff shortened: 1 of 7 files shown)


HuijingHei (approved):

dustymabe:
The CI failure here has to do with the fact that FCOS is now on F42 which now means we're inheriting from Fedora bootc via a git submodule but the `e2e-update.sh` doesn't account for that when doing a `git clone`. 

I'll open a separate PR to try to address this.

dustymabe:
> I'll open a separate PR to try to address this.

https://github.com/coreos/bootupd/pull/917

PR #917, opened 2025-04-23
tests/e2e-update: drop manifest_fork logic; update to f42 grub RPMs

dustymabe:
I feel like this was a bit heavyweight for what we were gaining
from it since we were only modifying one file.

It just so happens also that the `git clone` was not including the
git submodule that we're now using in FCOS for bootc.

Let's just drop the "forking" and do something a bit more lightweight.


HuijingHei (approved):
LGTM

HuijingHei:
Thank you @dustymabe for fixing this.

PR #923, opened 2025-05-07
Add `struct RootContext` with devices

HuijingHei:
Add `struct RootContext` to gather the context up
    
Get pointer by https://github.com/coreos/bootupd/pull/923#issuecomment-2873134953

---

update: pass `RootContext` as parameter

Also remove duplicated sysroot as `openat::Dir::open("/")`

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -12,6 +12,7 @@ use crate::efi;
 use crate::model::{ComponentStatus, ComponentUpdatable, ContentMetadata, SavedState, Status};
 use crate::util;
 use anyhow::{anyhow, Context, Result};
+use camino::{Utf8Path, Utf8PathBuf};
 use clap::crate_version;
 use fn_error_context::context;
 use libc::mode_t;
@@ -406,6 +407,34 @@ pub(crate) fn print_status(status: &Status) -> Result<()> {
     }
 
     Ok(())
+}
+
+pub struct RootContext {
+    pub sysroot: openat::Dir,
+    pub path: Utf8PathBuf,
+    pub devices: Vec<String>
+}
+
+impl RootContext {
+    fn new(sysroot: openat::Dir, path: &str, devices: Vec<String>) -> Self {
+        Self {
+            sysroot,
+            path: Utf8Path::new(path).into(),
+            devices,
+        }
+    }
+}
+
+/// Initialize parent devices to prepare the update
+fn prep_before_update() -> Result<RootContext> {
+    let path = "/";
+    let sysroot = openat::Dir::open(path).context("Opening root dir")?;
+    let devices = crate::blockdev::get_devices(path).context("get parent devices")?;
+    Ok(RootContext::new(
+        sysroot,
+        path,
+        devices
+    ))
 }
 
 pub(crate) fn client_run_update() -> Result<()> {
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -216,16 +216,16 @@ fn ensure_writable_boot() -> Result<()> {
 }
 
 /// daemon implementation of component update
-pub(crate) fn update(name: &str) -> Result<ComponentUpdateResult> {
-    let mut state = SavedState::load_from_disk("/")?.unwrap_or_default();
+pub(crate) fn update(name: &str, rootcxt: &RootContext) -> Result<ComponentUpdateResult> {
+    let mut state = SavedState::load_from_disk(&rootcxt.path)?.unwrap_or_default();
     let component = component::new_from_name(name)?;
     let inst = if let Some(inst) = state.installed.get(name) {
         inst.clone()
     } else {
         anyhow::bail!("Component {} is not installed", name);
     };
-    let sysroot = openat::Dir::open("/")?;
-    let update = component.query_update(&sysroot)?;
+    let sysroot = &rootcxt.sysroot;
+    let update = component.query_update(sysroot)?;
     let update = match update.as_ref() {
         Some(p) if inst.meta.can_upgrade_to(p) => p,
         _ => return Ok(ComponentUpdateResult::AtLatestVersion),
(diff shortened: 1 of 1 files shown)


HuijingHei:
@dustymabe I think you meant to initialize esp devices, store them and read after, this is the first step make parent devices global, initialize once and read then. Not sure if can make esp devices or bios devices like this, or other way?

nikita-dubrovskii:
LGTM

src/blockdev.rs
```diff
@@ -1,10 +1,23 @@
...
 use anyhow::{bail, Context, Result};
 use bootc_blockdev::PartitionTable;
 use fn_error_context::context;
 
+static PARENT_DEVICES: OnceLock<Vec<String>> = OnceLock::new();
+
+// Initialize once
+pub fn init_parent_devices(strings: Vec<String>) {
+    PARENT_DEVICES.set(strings).expect("Already initialized");
+}
```
cgwalters, line 14: This isn't the usual pattern for this. It's better to keep the `static` scoped inside a particular function. Here's an example
https://github.com/bootc-dev/bootc/blob/93b22f4dbc2d54f7cca7c1df3ee59fcdec0b2cf1/ostree-ext/src/globals.rs#L22

And notably this uses `get_or_init` which ensures initialize-at-most-once semantics, which avoids the `expect`.

src/cli/bootupctl.rs
```diff
@@ -129,7 +130,14 @@ impl CtlCommand {
...
     fn run_update() -> Result<()> {
         ensure_running_in_systemd()?;
-        bootupd::client_run_update()
+        let sysroot = openat::Dir::open("/").context("Opening root dir")?;
+        let dest_fd = format!("/proc/self/fd/{}", sysroot.as_raw_fd());
+        let dest_root = std::fs::read_link(dest_fd)?;
+
+        let devices = crate::blockdev::get_devices(&dest_root)
+            .with_context(|| "while looking for parent devices")?;
+        crate::blockdev::init_parent_devices(devices);
```
cgwalters: Here's where I'd suggest moving the `static` instead.

However even then *in theory* the parent devices are scoped to a particular destination root and so it can't actually be static. Which means it's hard to cache in that way.

So if we wanted to avoid querying the devices multiple times, then what we'd need to do is query them up front and then pass them down. That may involve more code churn though.

src/cli/bootupctl.rs
```diff
@@ -130,26 +130,14 @@ impl CtlCommand {
     /// Runner for `update` verb.
     fn run_update() -> Result<()> {
         ensure_running_in_systemd()?;
-        let sysroot = openat::Dir::open("/").context("Opening root dir")?;
-        let dest_fd = format!("/proc/self/fd/{}", sysroot.as_raw_fd());
-        let dest_root = std::fs::read_link(dest_fd)?;
```
cgwalters: Hmm so actually we always operated on `/`, and this code pattern was just a really complicated way to go from `/` -> file descriptor -> `/` effectively.

The only tricky thing here is that eventually we actually do want to operate on the non-booted root! See https://github.com/coreos/bootupd/issues/108

src/cli/bootupctl.rs
```diff
@@ -228,3 +216,15 @@ fn run_status_in_container(json_format: bool) -> Result<()> {
     }
     Ok(())
 }
+
+/// Initialize parent devices to prepare the update
+fn prep_before_update() -> Result<openat::Dir> {
+    let sysroot = openat::Dir::open("/").context("Opening root dir")?;
+    let dest_fd = format!("/proc/self/fd/{}", sysroot.as_raw_fd());
+    let dest_root = std::fs::read_link(dest_fd)?;
```
cgwalters, line 224: Per above I think we can just for now hardcode something like
`let devices = crate::blockdev::get_devices("/")` here.

It actually may make sense to move the `static` to this function as it's the one that's also hardcoding `/`.

dustymabe (approved):
IANA rust expert, but LGTM

HuijingHei:
Thank you all for the review! Does it make sense to do the same thing for `esp devices` and `bios devices` ? Like 
```
struct DeviceParts {
    parent_devices: Vec<String>,
    efi_devices: Option<Vec<String>>,
    bios_devices: Option<Vec<String>>,
}
```
Or just `parent_devices` is enough?

HuijingHei:
> Thank you all for the review! Does it make sense to do the same thing for `esp devices` and `bios devices`, make all of them global, initialize at the beginning and reuse them? Like
> 
> ```
> struct DeviceParts {
>     parent_devices: Vec<String>,
>     efi_devices: Option<Vec<String>>,
>     bios_devices: Option<Vec<String>>,
> }
> ```
> 
> Or just `parent_devices` is enough?

HuijingHei:
Sorry that click the wrong button, reopen it.

cgwalters:
I think Dusty's original comment in https://github.com/coreos/bootupd/pull/855#discussion_r2045394526 was correct in that we're clearly parsing the devices a lot. However, this is turning into a nontrivial architectural change and it's not clear to me that we have any real performance concerns here. (Really, most of this data is cached in the kernel or filesystem, we're just re-parsing files into JSON and then back into data structures, but computers are fast)

Looking at this more I don't think we want a global cache, we just want to gather the context up front and pass it down. So instead of passing just plain "sysroot" we'd pass e.g.
```
struct RootContext {
  sysroot: openat::Dir,
  devices: Vec<String>
}
```

or so? (And actually given all the pain that using openat has caused here, we might as well add `path: Utf8PathBuf` to that struct, so anything that wants to can just access it as an absolute path instead. I have something similar in bootc's install code. To make that work when we eventually do https://github.com/coreos/bootupd/issues/108 we'd just need to also mount the target root)

dustymabe:
> Looking at this more I don't think we want a global cache, we just want to gather the context up front and pass it down. So instead of passing just plain "sysroot" we'd pass e.g...

This SGTM.

HuijingHei:
> > Looking at this more I don't think we want a global cache, we just want to gather the context up front and pass it down. So instead of passing just plain "sysroot" we'd pass e.g...
> 
> This SGTM.

Looks sane to me. Will update according this.

HuijingHei:
@cgwalters, update according to https://github.com/coreos/bootupd/pull/923#issuecomment-2873134953 is done, could you help to review?

Next will update `run_update()` to reuse `RootContext.devices`.

cgwalters (approved):

cgwalters:
Thanks for all your hard work here!

PR #928, opened 2025-05-16
update: pass `RootContext` to `run_update()`

HuijingHei:


src/bootupd.rs
```diff
@@ -244,7 +244,7 @@ pub(crate) fn update(name: &str, rootcxt: &RootContext) -> Result<ComponentUpdat
         .context("Failed to update state")?;
 
     let newinst = component
-        .run_update(&state_guard.sysroot, &inst)
+        .run_update(rootcxt, &inst)
```
cgwalters, line 247: I think this is fine, though it probably makes sense to fold the SysrootGuard into being a lock on RootContext or so.

src/bootupd.rs
```diff
@@ -260,7 +260,7 @@ pub(crate) fn update(name: &str, rootcxt: &RootContext) -> Result<ComponentUpdat
 /// daemon implementation of component adoption
 pub(crate) fn adopt_and_update(name: &str, rootcxt: &RootContext) -> Result<ContentMetadata> {
     let sysroot = &rootcxt.sysroot;
-    let mut state = SavedState::load_from_disk(&rootcxt.path)?.unwrap_or_default();
+    let mut state = SavedState::load_from_disk("/")?.unwrap_or_default();
```
cgwalters, line 263: Why this change?

HuijingHei: I think we always want to load the saved state from "/boot/bootupd-state.json", right? If running `bootupctl update --from-pending`, my concern is that `RootContext.path` will be `/ostree/deploy/<pending>`, and the saved state will be `/ostree/deploy/<pending>/boot/bootupd-state.json`, so I change it back to "/", but I might be wrong.

src/bootupd.rs
HuijingHei: Not sure I catch your meaning, current `StateLockGuard` locks `sysroot` Dir, do you mean to add `RootContext` to `StateLockGuard` ?

src/bootupd.rs
cgwalters: Yes makes sense!

cgwalters (approved):

src/bootupd.rs
cgwalters: It's all good as is!

Issue #746, opened 2024-10-16
Bootupd installation is missing `/boot/grub2/grubenv`


Jmennius:
I was recently installing Fedora Silverblue 41 nightly build (`Fedora-Silverblue-ostree-x86_64-41-20241012.n.0.iso`) and I've noticed that the system was missing `/boot/grub2/grubenv` and `/etc/default/grub` files.

Although the system boots fine, I'm pretty sure this is a bug because grub menu is not automatically hidden, boots cannot be marked as successful and so on (`grubenv` is not created after the install was done under any circumstance automatically).

I was doing an installation via a kickstart file (for advanced encryption option), but I've customized a file that I got from a test install (although I think it was Beta 1.2 image).

I've contacted anaconda team (as it was creating those files before - see `write_defaults()` and `write_config()` in https://github.com/rhinstaller/anaconda/blob/4683a8965ce02443134115e3eeea58ec6df38da3/pyanaconda/modules/storage/bootloader/grub2.py#L253) and they said this is now responsibility of bootupd :)

Jmennius:
Should I instead open a bugzilla issue instead/in addition to?

HuijingHei:
Not sure if we should make `bootupd` to install these 2 files, can we reuse `anaconda`? Any other scenarios that we need them?
Maybe @travier @cgwalters know about this.

travier:
We should probably create `/boot/grub2/grubenv` but likely not `/etc/default/grub` as we don't need it anymore.

Jmennius:
> We should probably create `/boot/grub2/grubenv` but likely not `/etc/default/grub` as we don't need it anymore.

Regarding defaults - not so sure.
If a user creates the file because he wants to configure something - he is most likely to not put the BLSCFG variable as well as other variables.
Right now, when there is no defaults file, ostree grub script will just exit.
If the file exists it wants BLSCFG variable to be set and /boot/grub/.*blscfg* file to be present (which is not the case) to not generate boot entries for grub (which avoid duplicate boot entries)...

travier:
> Regarding defaults - not so sure. If a user creates the file because he wants to configure something - he is most likely to not put the BLSCFG variable as well as other variables. Right now, when there is no defaults file, ostree grub script will just exit.

ostree will also skip re-generating the GRUB config if the system has a static GRUB config which is the default starting with F41 installations.

`/etc/default/grub` is only used on older installations and will be ignored on the newer ones.

travier:
[...]

PR #931, opened 2025-05-19
grubconfig: create `boot/grub2/grubenv` if not existing

HuijingHei:
Fixes https://github.com/coreos/bootupd/issues/746

### DIFF
--- a/src/grubconfigs.rs
+++ b/src/grubconfigs.rs
@@ -1,7 +1,9 @@
 use std::fmt::Write;
+use std::os::unix::io::AsRawFd;
 use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Context, Result};
+use bootc_utils::CommandRunExt;
 use fn_error_context::context;
 use openat_ext::OpenatDirExt;
 
@@ -9,6 +11,7 @@ use openat_ext::OpenatDirExt;
 const GRUB2DIR: &str = "grub2";
 const CONFIGDIR: &str = "/usr/lib/bootupd/grub2-static";
 const DROPINDIR: &str = "configs.d";
+const GRUBENV: &str = "grubenv";
 
 /// Install the static GRUB config files.
 #[context("Installing static GRUB configs")]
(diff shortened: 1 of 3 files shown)


travier:
```
---- grubconfigs::tests::test_write_grubenv stdout ----
Error: Create /boot/grub2/grubenv

Caused by:
    No such file or directory (os error 2)
```

src/grubconfigs.rs
```diff
@@ -124,4 +144,15 @@ mod tests {
...
     }
+    #[test]
+    fn test_write_grubenv() -> Result<()> {
+        let td = tempfile::tempdir()?;
+        let tdp = td.path();
+        std::fs::create_dir_all(tdp.join("boot/grub2"))?;
+        let td = openat::Dir::open(&tdp.join("boot"))?;
+        write_grubenv(&td)?;
+
+        assert!(td.exists("grub2/grubenv")?);
```
travier, line 155:
```suggestion
        assert!(td.exists("boot/grub2/grubenv")?);
```

travier: Ah, no, you redefine td above.

HuijingHei: No worry, the CI failed because not found `/usr/bin/grub2-editenv` as not install `grub2-tools-minimal`

HuijingHei:
> LGTM. I haven't tested this however and I'm not sure how we could do that easily.

Agree, not easy. If using `bootc install`, it will by default install both BIOS and UEFI.

Just build bootupd and add as layered based on `fedora-bootc:42`, push to quay.io, then use [anaconda](https://docs.fedoraproject.org/en-US/bootc/bare-metal/#_using_anaconda) to install from network with `ostreecontainer --url quay.io/hhei/bootc-test` and UEFI firmware, check we can get `/boot/grub2/grubenv`

```
[root@fedora ~]# ls /boot/grub2/grubenv 
/boot/grub2/grubenv
```

HuijingHei:
@travier  I update command `.run()` to `.run_with_cmd_context()`, and this can get more info if failed. Could you help to review again? Thanks!

HuijingHei:
Thank you @travier for the kind review!

dustymabe (commented):
LGTM

PR #932, opened 2025-05-21
efi: rewrite `ensure_mounted_esp()`

HuijingHei:
This is the preparation for https://github.com/coreos/bootupd/pull/855.

The main target for this PR is: "efi: rewrite `ensure_mounted_esp()`", but find there are much to update consequently.

---
efi: rewrite `ensure_mounted_esp()`

Split into 2 parts:
- `get_mounted_esp()` to get mounted point esp
- `mount_esp_device()` to mount the passed esp device, return
the mount point

Then in `ensure_mounted_esp()` call the above 2 functions:
firstly check if esp is already mounted, if not, mount the passed
esp device.

---
efi: update `install()` and `update()` according to split change

---
adopt_update: pass `RootContext` as parameter

---
efi: update `validate()` according to the split change

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -328,8 +328,21 @@ pub(crate) fn status() -> Result<Status> {
     // Process the remaining components not installed
     log::trace!("Remaining known components: {}", known_components.len());
     for (name, _) in known_components {
-        // Only run `query_adopt_state()` is enough
-        // When do adopt and update for each component, will check more
+        // To determine if not-installed components can be adopted:
+        //
+        // `query_adopt_state()` checks for existing installation state,
+        // such as a `version` in `/sysroot/.coreos-aleph-version.json`,
+        // or the presence of `/ostree/deploy`.
+        //
+        // `component.query_adopt()` performs additional checks,
+        // including hardware/device requirements.
+        // For example, it will skip BIOS adoption if the system is booted via EFI
+        // and lacks a BIOS_BOOT partition.
+        //
+        // Once a component is determined to be adoptable, it is added to the
+        // adoptable list, and adoption proceeds automatically.
+        //
+        // Therefore, calling `query_adopt_state()` alone is sufficient.
         if let Some(adopt_ver) = crate::component::query_adopt_state()? {
             ret.adoptable.insert(name.to_string(), adopt_ver);
         } else {
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -276,7 +276,7 @@ pub(crate) fn adopt_and_update(name: &str, rootcxt: &RootContext) -> Result<Cont
         SavedState::acquire_write_lock(sysroot).context("Failed to acquire write lock")?;
 
     let inst = component
-        .adopt_update(&state_guard.sysroot, &update)
+        .adopt_update(&rootcxt, &update)
         .context("Failed adopt and update")?;
     state.installed.insert(component.name().into(), inst);
 
@@ -327,8 +327,10 @@ pub(crate) fn status() -> Result<Status> {
 
     // Process the remaining components not installed
     log::trace!("Remaining known components: {}", known_components.len());
-    for (name, component) in known_components {
-        if let Some(adopt_ver) = component.query_adopt()? {
+    for (name, _) in known_components {
+        // Only run `query_adopt_state()` is enough
+        // When do adopt and update for each component, will check more
+        if let Some(adopt_ver) = crate::component::query_adopt_state()? {
             ret.adoptable.insert(name.to_string(), adopt_ver);
         } else {
             log::trace!("Not adoptable: {}", name);
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -264,9 +264,8 @@ impl Component for Efi {
         "EFI"
     }
 
-    fn query_adopt(&self) -> Result<Option<Adoptable>> {
-        let esp = self.open_esp_optional()?;
-        if esp.is_none() {
+    fn query_adopt(&self, devices: &Option<Vec<String>>) -> Result<Option<Adoptable>> {
+        if devices.is_none() {
             log::trace!("No ESP detected");
             return Ok(None);
         };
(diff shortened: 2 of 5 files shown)


src/efi.rs
```diff
@@ -303,18 +313,27 @@ impl Component for Efi {
         log::debug!("Found metadata {}", meta.version);
         let srcdir_name = component_updatedirname(self);
         let ft = crate::filetree::FileTree::new_from_dir(&src_root.sub_dir(&srcdir_name)?)?;
-        let destdir = &self.ensure_mounted_esp(Path::new(dest_root))?;
 
-        let destd = &openat::Dir::open(destdir)
-            .with_context(|| format!("opening dest dir {}", destdir.display()))?;
+        let destpath = if let Some(destdir) = self.get_mounted_esp(Path::new(dest_root))? {
```
HuijingHei, line 317: Should find out whether this should be `src_root` or `dest_root`.

HuijingHei: I think it should be `dest_root` that will be as mount point.

src/efi.rs
```diff
@@ -265,31 +258,48 @@ impl Component for Efi {
...
-        let Some(meta) = self.query_adopt()? else {
+        let esp_devices = blockdev::find_colocated_esps(&sysroot.devices)?;
+        let Some(meta) = self.query_adopt(&esp_devices)? else {
             anyhow::bail!("Failed to find adoptable system")
         };
 
-        let esp = self.open_esp()?;
-        validate_esp(&esp)?;
+        // Confirm that esp_devices is Some(value)
+        let esp_devices = esp_devices.unwrap();
```
cgwalters, line 270: But we also confirm that just below with a runtime error and not a panic, which I think is right; so we can just drop this line.

src/efi.rs
```diff
@@ -97,29 +76,29 @@ impl Efi {
...
-            util::ensure_writable_mount(&mnt)?;
-            log::debug!("Reusing existing {mnt:?}");
-            return Ok(mnt);
+    // Get mounted point for esp
+    pub(crate) fn get_mounted_esp(&self, root: &Path) -> Result<Option<PathBuf>> {
+        // First check all potential mount points without holding the borrow
+        let found_mount = ESP_MOUNTS.iter().map(|&mnt| root.join(mnt)).find(|mnt| {
+            mnt.exists()
+                && rustix::fs::statfs(mnt).map_or(false, |st| st.f_type == libc::MSDOS_SUPER_MAGIC)
+                && util::ensure_writable_mount(mnt).is_ok()
```
cgwalters, line 85: This "swallows" errors. Instead I think we should use `try_fold`; see e.g. this thread https://github.com/bootc-dev/bootc/pull/1299#discussion_r2087544274

cgwalters (commented):
Looks reasonable to me overall, I just have some nits

src/efi.rs
HuijingHei: Can not drop as will get error, use `.unwrap_or_default()` instead.

src/efi.rs
HuijingHei: Updated, could you help to review? Thanks!

cgwalters (commented):
It looks like the cosa build is legitimately failing.

src/efi.rs
cgwalters: It looks better yes. However now that I look even more closely I think it'd likely be simpler to just go back to explicit iteration? But OK as is too.

HuijingHei:
Seems the `try_fold()` iter all items without return at the first match, checking more.

src/efi.rs
```diff
@@ -303,18 +321,28 @@ impl Component for Efi {
         log::debug!("Found metadata {}", meta.version);
         let srcdir_name = component_updatedirname(self);
         let ft = crate::filetree::FileTree::new_from_dir(&src_root.sub_dir(&srcdir_name)?)?;
-        let destdir = &self.ensure_mounted_esp(Path::new(dest_root))?;
```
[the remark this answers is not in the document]
HuijingHei: Good question. After change, the function `ensure_mounted_esp()` will check if esp is already mounted, else mount the passed `esp device`, return the mount point and append path `/EFI`.

For installation, during `bootc install to-disk`, the esp device is not mounted and will call `get_esp_device()` to get esp device via `/dev/disk/by-partlabel`, but when running `cosa build`, not sure if it would also call bootc, find that esp device is already mounted, and can not find esp device via `/dev/disk/by-partlabel` (do not why). Before I use `esp_device = self.get_esp_device()` before `ensure_mounted_esp()`, but `cosa build` failed.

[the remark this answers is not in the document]
HuijingHei: Maybe can use `get_esp_partition(device)` (via sfdisk) instead of `get_esp_device()` (via `/dev/disk/by-partlabel`) ?

src/efi.rs
```diff
@@ -321,10 +321,16 @@ impl Component for Efi {
         log::debug!("Found metadata {}", meta.version);
         let srcdir_name = component_updatedirname(self);
         let ft = crate::filetree::FileTree::new_from_dir(&src_root.sub_dir(&srcdir_name)?)?;
-        // get esp device via /dev/disk/by-partlabel
-        let esp_device = self
-            .get_esp_device()
-            .ok_or_else(|| anyhow::anyhow!("Failed to find ESP device"))?;
+
+        // Get esp device via /dev/disk/by-partlabel
+        // If failed, change to sfdisk
```
dustymabe, line 326: I would add a lot more to this comment. Basically all of what is in the commit message.

```
During `bootc install to-disk` the esp device is not mounted but
we can call `get_esp_device()` and it will find the device via
the `/dev/disk/by-partlabel/` symlink. When running `cosa build` and
building a disk image using `OSBuild` the EFI partitions is mounted
but no udev `/dev/disk/by-partlabel/` symlinks exist because it's 
just a chroot and not a full linux environment. In that case
fallback to using `blockdev` to find the partition.
```

src/efi.rs
```diff
@@ -321,10 +321,16 @@ impl Component for Efi {
...
+        let esp_device = if let Some(esp) = self.get_esp_device() {
+            esp
+        } else {
+            let esp = blockdev::get_esp_partition(device)?
```
dustymabe, line 330: I wonder what the downside would be of just changing `self.get_esp_device()` to `self.get_esp_device(device)` and just use `blockdev::get_esp_partition(device)` inside of that function and not worry about `/dev/disk/by-label` at all.

It appears this is the only use of `self.get_esp_device()` that I see at this time.

Even better, get rid of `self.get_esp_device()` altogether and just use blockdev here:

```
let found_esp = blockdev::get_esp_partition(device)?                    
    .ok_or_else(|| anyhow::anyhow!("Failed to find ESP device"))?;
let esp = PathBuf::from(found_esp)                                                

```

of course, the variable names could be improved :)

dustymabe: Doing this would also mean that we wouldn't need a huge comment explaining why we use `/dev/disk/by-label` in one case and inspect the partitions in another.

src/efi.rs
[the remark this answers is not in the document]
HuijingHei: According to [comment](https://github.com/coreos/bootupd/pull/932#discussion_r2112572571), use `blockdev` to find the partition, instead of `partlabel`.

src/bootupd.rs
```diff
@@ -327,8 +327,10 @@ pub(crate) fn status() -> Result<Status> {
 
     // Process the remaining components not installed
     log::trace!("Remaining known components: {}", known_components.len());
-    for (name, component) in known_components {
-        if let Some(adopt_ver) = component.query_adopt()? {
+    for (name, _) in known_components {
+        // Only run `query_adopt_state()` is enough
+        // When do adopt and update for each component, will check more
```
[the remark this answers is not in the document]
dustymabe: looks better

dustymabe (approved):
LGTM - merge when ready

Issue #926, opened 2025-05-15
Extend `generate-update-metadata` to read from `/usr` (Unification of boot loader updates, phase 1)


travier:
Extend the `generate-update-metadata` logic to not query the RPM db but instead pick the versions from the new paths where the grub2 & shim binaries will be installed in `/usr/lib/(shim|grub)/%{version}-%{release}/EFI/...` when we land https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1.

See:
- https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1
- https://gitlab.com/fedora/bootc/tracker/-/issues/61
- https://gitlab.com/fedora/bootc/base-images/-/merge_requests/90

Draft PRs for GRUB & shim that will use the new paths:
- https://src.fedoraproject.org/rpms/grub2/pull-request/136
- https://src.fedoraproject.org/rpms/grub2/pull-request/140
- https://src.fedoraproject.org/rpms/shim/pull-request/4

travier:
Ah, there is no dist for shim as it's the same build for all releases

travier:
So this makes installing the shim & grub binaries under versioned directories less interesting.

lsandov1:
I have reworked MRs [1,2] (no scratch builds this time) based on your comments and the only piece different is new `efi` folder on top of `/usr/lib/`, where grub and shim will place their files on top

```
root@fedora rpms-unification]# rpm -ql grub2-efi-x64 shim-x64 | grep /usr/lib/efi
/usr/lib/efi/grub2/2.12-34.fc42/EFI/fedora/grubx64.efi
/usr/lib/efi/shim/15.8-4/EFI/BOOT/BOOTX64.EFI
/usr/lib/efi/shim/15.8-4/EFI/BOOT/fbx64.efi
/usr/lib/efi/shim/15.8-4/EFI/fedora/BOOTX64.CSV
/usr/lib/efi/shim/15.8-4/EFI/fedora/mmx64.efi
/usr/lib/efi/shim/15.8-4/EFI/fedora/shim.efi
/usr/lib/efi/shim/15.8-4/EFI/fedora/shimx64.efi
```

but maybe this does not solve the issue.

@travier @HuijingHei propose any directory hierarchy and I follow it, whatever if fits better for your scripting.

HuijingHei:
Copy @cgwalters 's comment here in case it is buried.

[...]

PR #935, opened 2025-05-27
extend-payload-to-esp:  extends any src_input_dir to esp 

### LINKED ISSUE #651: Support for Raspberry Pi firmwares/bootloaders
Raspberry Pi requires a bunch of files to be available in the root of the ESP partition (traditionally mounted as `/boot/efi` on Fedora) in order to boot. In the case of Fedora, this means the firmware, uboot, device trees and probably more. This is the content of `/boot/efi` on Fedora Server for aarch64:
```
$ ls /boot/efi
bcm2710-rpi-2-b.dtb*       bcm2710-rpi-cm3.dtb*       bcm2711-rpi-400.dtb*  bcm2711-rpi-cm4-io.dtb*  bcm2712-rpi-5-b.dtb*        bootcode.bin*  fixup4cd.dat*  fixup4x.dat*   fixup_db.dat*  rpi-u-boot.bin*  start4.elf*    start_db.elf*
bcm2710-rpi-3-b.dtb*       bcm2710-rpi-zero-2.dtb*    bcm2711-rpi-4-b.dtb*  bcm2711-rpi-cm4s.dtb*    bcm2712-rpi-cm5-cm4io.dtb*  config.txt*    fixup4.dat*    fixup_cd.dat*  fixup_x.dat*   start4cd.elf*    start4x.elf*   start.elf*
bcm2710-rpi-3-b-plus.dtb*  bcm2710-rpi-zero-2-w.dtb*  bcm2711-rpi-cm4.dtb*  bcm2712d0-rpi-5-b.dtb*   bcm2712-rpi-cm5-cm5io.dtb*  EFI/           fixup4db.dat*  fixup.dat*     overlays/      start4db.elf*    start_cd.elf*  start_x.elf*
```

However, `bootupd` currently only manages the `/boot/efi/EFI` directory, so there's no way to put these files under `/boot/efi` without any kind of post-processing run after `bootupd backend install`.

### LINKED ISSUE #766: support dynamically extending/adding update payloads
See this https://github.com/ondrejbudai/fedora-bootc-raspi/blob/cbff66a8bdb82b6276814765ffdb82c4ec4efa9c/bootupctl-shim#L7

Basically today the bootloader update payload is often part of the base image, but we need to support dynamically extending it with layered image content as well.

---

In the short term, what may suffice is to offer an explicit user-hookable phase that can copy things from the container to `/boot` as part of `bootc install to-filesystem`.

say-paul:
should fix: #651 
 fixes : #766 
 To support raspberry pi 4, which works well for rpm-ostree, but in bootc.
 This PR add command `extend-payload-to-esp`  which will enable bootup install to move the files to esp.
usage:
```
RUN dnf install -y uboot-images-armv8

# rpm -ql uboot-images-armv8 | grep rpi
# /usr/share/uboot/rpi_arm64
# /usr/share/uboot/rpi_arm64/u-boot.bin

RUN bootupctl backend extend-payload-to-esp /usr/share/uboot/rpi_arm64
```
 
what extend-payload-to-esp does:
- stages the files to /usr/lib/efi/firmware/{name}/{version}/EFI/
- stores rpm details and timestamp in /usr/lib/efi/firmware/{name}/{version}/EFI.json
```
bash-5.2# tree /usr/lib/efi/firmware/uboot/
/usr/lib/efi/firmware/uboot/
`-- 1:2025.04-1
    |-- EFI
    |   `-- u-boot.bin
    `-- EFI.json
```
- `bootupd install` and `bootupd update` is updated to read from the above path and install to esp.
- use the metadata(EFI.json) to update the bootupd-state

rpi4 -usecase: https://github.com/fedora-iot/iot-distro/issues/82#issuecomment-3051419262

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,6 +183,7 @@ jobs:
               echo "✓ extend-payload completed successfully"
               # Now test install to disk with extended firmware
               bootc install to-disk --skip-fetch-check \
+                -vv \
                 --disable-selinux --generic-image --via-loopback /target/myimage-extend.raw
             '
             
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,3 +139,195 @@ jobs:
               bootupctl backend generate-update-metadata -vvv
               cat ${updates}/EFI.json | jq
             '
+
+      - name: Test install after extend-payload
+        run: |
+          set -xeuo pipefail
+          sudo truncate -s 5G myimage-extend.raw
+          sudo podman run --rm --privileged -v .:/target --pid=host --security-opt label=disable \
+            -v /var/lib/containers:/var/lib/containers \
+            -v /dev:/dev \
+            localhost/bootupd:latest bash -c '
+              # Create test firmware directory and files
+              mkdir -p /usr/share/uboot/rpi/overlays
+              echo "test uboot binary content" > /usr/share/uboot/rpi/u-boot.bin
+              echo "i2c device tree overlay" > /usr/share/uboot/rpi/overlays/i2c.dtb
+              
+              # Create a fake RPM database for testing
+              mkdir -p /usr/lib/sysimage/rpm
+              echo "fake rpm database" > /usr/lib/sysimage/rpm/Packages
+              
+              # Create mock rpm script using a here-document for clarity
+              cat << '"'EOT'"' > /usr/local/bin/rpm
+              #!/bin/bash
+              if [[ "$*" == *"-q"* ]] && [[ "$*" == *"-f"* ]]; then
+                  echo "uboot-images-2023.04-2.fc42.noarch,1681234567"
+                  exit 0
+              fi
+              exec /usr/bin/rpm.orig "$@"
+              EOT
+              
+              # Backup original rpm and make our mock executable
+              cp /usr/bin/rpm /usr/bin/rpm.orig
+              chmod +x /usr/local/bin/rpm
+              export PATH="/usr/local/bin:$PATH"
+              
+              # Run extend-payload-to-esp first
+              bootupctl backend extend-payload-to-esp /usr/share/uboot/rpi
+              
+              # Verify firmware was extended correctly
+              test -d /usr/lib/efi/firmware || { echo "firmware directory not created"; exit 1; }
+              firmware_ver_dir=$(find /usr/lib/efi/firmware -name "*2023.04*" -type d | head -1)
+              test -n "${firmware_ver_dir}" || { echo "firmware version directory not found"; exit 1; }
+              test -f "${firmware_ver_dir}/EFI/u-boot.bin" || { echo "u-boot.bin not copied"; exit 1; }
+              echo "✓ extend-payload completed successfully"
+              # Now test install to disk with extended firmware
+              bootc install to-disk --skip-fetch-check \
+                --disable-selinux --generic-image --via-loopback /target/myimage-extend.raw
+            '
+            
+          # Verify firmware files were installed to ESP
+          sudo losetup -P -f myimage-extend.raw
+          device=$(losetup -a myimage-extend.raw --output NAME -n)
+          esp_part=$(sudo sfdisk -l -J "${device}" | jq -r '.partitiontable.partitions[] | select(.type == "C12A7328-F81F-11D2-BA4B-00A0C93EC93B").node')
+          sudo mount "${esp_part}" /mnt/
+          
+          # Check that firmware files were copied to ESP during install
+          if sudo test -f /mnt/u-boot.bin; then
+            sudo grep -q "test uboot binary content" /mnt/u-boot.bin || { echo "u-boot.bin content incorrect on ESP"; exit 1; }
+            echo "✓ Firmware files correctly installed to ESP"
+          else
+            echo "Note: u-boot.bin not found on ESP (firmware install integration may need work)"
+          fi
+          
+          sudo umount /mnt
+          sudo losetup -D "${device}"
+          sudo rm -f myimage-extend.raw
+
+      - name: Test update after extend-payload
+        run: |
+          set -xeuo pipefail
+          sudo truncate -s 5G myimage-update.raw
+          sudo podman run --rm --privileged -v .:/target --pid=host --security-opt label=disable \
+            -v /var/lib/containers:/var/lib/containers \
+            -v /dev:/dev \
+            localhost/bootupd:latest bash -c '
+              # Create initial test firmware directory and files
+              mkdir -p /usr/share/uboot/rpi/overlays
+              echo "initial uboot binary content v1.0" > /usr/share/uboot/rpi/u-boot.bin
+              echo "initial i2c device tree overlay" > /usr/share/uboot/rpi/overlays/i2c.dtb
+              
+              # Create a fake RPM database for testing
+              mkdir -p /usr/lib/sysimage/rpm
+              echo "fake rpm database" > /usr/lib/sysimage/rpm/Packages
+              
+              # Create mock rpm script that returns initial package data
+              cat << '"'EOT'"' > /usr/local/bin/rpm
+              #!/bin/bash
+              if [[ "$*" == *"-q"* ]] && [[ "$*" == *"-f"* ]]; then
+                  echo "uboot-images-2023.04-1.fc42.noarch,1681234567"
+                  exit 0
+              fi
+              exec /usr/bin/rpm.orig "$@"
+              EOT
+              
+              # Backup original rpm and make our mock executable
+              cp /usr/bin/rpm /usr/bin/rpm.orig
+              chmod +x /usr/local/bin/rpm
+              export PATH="/usr/local/bin:$PATH"
+              
+              # Run initial extend-payload-to-esp
+              bootupctl backend extend-payload-to-esp /usr/share/uboot/rpi
+              
+              # Verify initial firmware was extended correctly
+              test -d /usr/lib/efi/firmware || { echo "firmware directory not created"; exit 1; }
+              firmware_ver_dir=$(find /usr/lib/efi/firmware -name "*2023.04-1*" -type d | head -1)
+              test -n "${firmware_ver_dir}" || { echo "initial firmware version directory not found"; exit 1; }
+              test -f "${firmware_ver_dir}/EFI/u-boot.bin" || { echo "initial u-boot.bin not copied"; exit 1; }
+              grep -q "initial uboot binary content v1.0" "${firmware_ver_dir}/EFI/u-boot.bin"
+              echo "✓ initial extend-payload completed successfully"
+              
+              # Install to disk with initial firmware
+              bootc install to-disk --skip-fetch-check \
+                --disable-selinux --generic-image --via-loopback /target/myimage-update.raw
+              
+              # Now simulate a firmware update by creating new firmware files
+              echo "updated uboot binary content v2.0" > /usr/share/uboot/rpi/u-boot.bin
+              echo "updated i2c device tree overlay" > /usr/share/uboot/rpi/overlays/i2c.dtb
+              echo "new overlay for v2" > /usr/share/uboot/rpi/overlays/spi.dtb
+              
+              # Update mock rpm to return new version
+              cat << '"'EOT'"' > /usr/local/bin/rpm
+              #!/bin/bash
+              if [[ "$*" == *"-q"* ]] && [[ "$*" == *"-f"* ]]; then
+                  echo "uboot-images-2023.04-2.fc42.noarch,1681234999"
+                  exit 0
+              fi
+              exec /usr/bin/rpm.orig "$@"
+              EOT
+              
+              # Run updated extend-payload-to-esp
+              bootupctl backend extend-payload-to-esp /usr/share/uboot/rpi
+              
+              # Verify updated firmware was extended correctly (only v2.0 should exist now)
+              updated_firmware_ver_dir=$(find /usr/lib/efi/firmware -name "*2023.04-2*" -type d | head -1)
+              test -n "${updated_firmware_ver_dir}" || { echo "updated firmware version directory not found"; exit 1; }
+              test -f "${updated_firmware_ver_dir}/EFI/u-boot.bin" || { echo "updated u-boot.bin not copied"; exit 1; }
+              grep -q "updated uboot binary content v2.0" "${updated_firmware_ver_dir}/EFI/u-boot.bin"
+              test -f "${updated_firmware_ver_dir}/EFI/overlays/spi.dtb" || { echo "new spi.dtb not copied"; exit 1; }
+              
+              # Verify old version (2023.04-1) was removed
+              old_firmware_ver_dir=$(find /usr/lib/efi/firmware -name "*2023.04-1*" -type d | head -1)
+              test -z "${old_firmware_ver_dir}" || { echo "old firmware version should have been removed but still exists: ${old_firmware_ver_dir}"; exit 1; }
+              
+              echo "✓ updated extend-payload completed successfully (old version cleaned up)"
+              
+              # Run bootupctl update to apply the updated firmware to ESP
+              bootupctl update
+              echo "✓ bootupctl update completed successfully"
+            '
+            
+          # Verify updated firmware files were applied to ESP
+          sudo losetup -P -f myimage-update.raw
+          device=$(losetup -a myimage-update.raw --output NAME -n)
+          esp_part=$(sudo sfdisk -l -J "${device}" | jq -r '.partitiontable.partitions[] | select(.type == "C12A7328-F81F-11D2-BA4B-00A0C93EC93B").node')
+          sudo mount "${esp_part}" /mnt/
+          
+          # Check that updated firmware files were applied to ESP during update
+          if sudo test -f /mnt/u-boot.bin; then
+            sudo grep -q "updated uboot binary content v2.0" /mnt/u-boot.bin || { echo "u-boot.bin was not updated on ESP"; exit 1; }
+            echo "✓ Updated firmware files correctly applied to ESP"
+          else
+            echo "Warning: u-boot.bin not found on ESP after update"
+            exit 1
+          fi
+          
+          # Check that new overlay file was also copied
+          if sudo test -f /mnt/overlays/spi.dtb; then
+            sudo grep -q "new overlay for v2" /mnt/overlays/spi.dtb || { echo "spi.dtb content incorrect on ESP"; exit 1; }
+            echo "✓ New overlay files correctly applied to ESP"
+          else
+            echo "Warning: new spi.dtb not found on ESP after update"
+          fi
+          
+          # Verify checksums and state integrity
+          echo "🔍 Validating firmware checksums and state integrity..."
+          sudo podman run --rm --privileged -v .:/target --pid=host --security-opt label=disable \
+            -v /var/lib/containers:/var/lib/containers \
+            -v /dev:/dev \
+            localhost/bootupd:latest bash -c '
+              # Run bootupctl validate to check all checksums
+              bootupctl validate || { echo "bootupctl validate failed - checksum mismatch detected"; exit 1; }
+              echo "✓ All file checksums validated successfully"
+              
+              # Check that bootupd-state.json reflects the updated firmware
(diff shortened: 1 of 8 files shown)


cgwalters (dismissed):
Thanks for working on this!

src/efi.rs
```diff
@@ -162,6 +162,39 @@ impl Efi {
         clear_efi_target(&product_name)?;
         create_efi_boot_entry(device, espdir, vendordir, &product_name)
     }
+
+    /// Copy firmware files from /boot/efi to target directory
+    fn copy_firmware_blobs(&self, src_root: &openat::Dir, dest_root: &str) -> Result<()> {
+        let src_efi = "boot/efi";
```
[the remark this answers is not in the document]
say-paul: Ran: #766 with he current implmentation with an additional step like of creating the files in ` /boot`:
```
# Build from the current git into a c9s-bootc container image.
# Use e.g. --build-arg=base=quay.io/fedora/fedora-bootc:41 to target
# Fedora or another base image instead.
#
ARG base=quay.io/centos-bootc/centos-bootc:stream9

FROM $base as build
# This installs our package dependencies, and we want to cache it independently of the rest.
# Basically we don't want changing a .rs file to blow out the cache of packages.
RUN <<EORUN
set -xeuo pipefail
dnf -y install cargo git openssl-devel
EORUN

RUN mkdir -p /boot/firmware/efi && \
    touch /boot/firmware/efi/dummy.{dtb,dat,efi,elf,bin}
# Now copy the source
COPY . /build
WORKDIR /build
# See https://www.reddit.com/r/rust/comments/126xeyx/exploring_the_problem_of_faster_cargo_docker/
# We aren't using the full recommendations there, just the simple bits.
RUN --mount=type=cache,target=/build/target --mount=type=cache,target=/var/roothome \
    make && make install-all DESTDIR=/out
  
# FROM $base
# # Clean out the default to ensure we're using our updated content
# RUN rpm -e bootupd
# COPY --from=build /out/ /
# # Sanity check this too
# RUN bootc container lint --fatal-warnings
  
    
``` 
From #755 
> Content shipped in an RPM with files in /boot that are present in the initial base image build where bootupctl backend generate-update-payload is invoked

The build container has the dummy files:
```
bash-5.2# ls /boot/firmware/efi/          
dummy.bin  dummy.dat  dummy.dtb  dummy.efi  dummy.elf
```

but the final build bu using the copy commands  doesn't have the files,  :confused:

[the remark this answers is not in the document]
say-paul: I think this function can be modified into a generic one to copy files, but I am not sure how to trigger it from the container build process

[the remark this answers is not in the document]
cgwalters: > but the final image that is built using the COPY --from=build /out/ / doesn't have the files, 😕

The `build` image is only intended to hold the result of building the bootupd binary itself. If you want to modify the final image for other reasons (e.g. injecting kernel arguments, adding debug tools or other configuration, and as in this case inject arbitrary other content) then you should put it on the last stage.

Or better even - don't modify the dockerfile, make a new one and have it be `FROM localhost/bootupd`.

say-paul:
It should now stage the updates for any user input path,
I think we have to additionally chain this in  efi::{install,run_update}
Its very similar to `generate_update_metadata`  but with an user_input.

travier:
A lot of questions are unanswered here:
- Is this file versioned?
- Is it always installed under the same name?
- Is it tied to a kernel version?
- What happens when we rollback?

Overall, I would suggest we work toward https://github.com/coreos/bootupd/issues/926 which should get us to https://github.com/coreos/bootupd/issues/766.

say-paul:
@travier please let me know if I am missing anything.

cgwalters:
Can you add an example canonical usage of this in the initial PR comment or so at least? A test would be even more ideal.

Basically to demonstrate how this is used.

I think again just to summarize two important goals:
- no changes to `bootc install` are required
- `bootupctl upgrade` Just Works to also upgrade

say-paul:
I can add test to it, but I want to validate my approach.
I ran the following to test it manually:

I usedc Containerfile:
```
ARG base=quay.io/fedora/fedora-bootc:41

FROM $base as build
# This installs our package dependencies, and we want to cache it independently of the rest.
# Basically we don't want changing a .rs file to blow out the cache of packages.
RUN <<EORUN
set -xeuo pipefail
dnf -y install cargo git openssl-devel
EORUN
# Now copy the source
COPY . /build
WORKDIR /build
# See https://www.reddit.com/r/rust/comments/126xeyx/exploring_the_problem_of_faster_cargo_docker/
# We aren't using the full recommendations there, just the simple bits.
RUN --mount=type=cache,target=/build/target --mount=type=cache,target=/var/roothome \
    make && make install-all DESTDIR=/out

FROM $base
# Clean out the default to ensure we're using our updated content
RUN rpm -e bootupd
COPY --from=build /out/ /
RUN dnf -y install uboot-images-armv8
RUN bootupctl backend extend-payload-to-esp  /usr/share/uboot/rpi_arm64
```
Build on rpi4:
`[root@192 bootupd]# podman build -t bootc_payload_extend --build-arg=base=quay.io/fedora/fedora-bootc:42 .`

Validation:
```
[root@192 bootupd]# podman run -it localhost/bootc_payload_extend:latest /bin/bash
bash-5.2# tree /usr/lib/efi/firmware/
/usr/lib/efi/firmware/
`-- uboot
    `-- 1:2025.04-2
        `-- EFI
            `-- boot
                `-- efi
                    `-- u-boot.bin
```

the final destination is in accordance to #926 , #938
I am unaware of how the things proceeds next,  but i think generate-update-metadata() from #938 can pick it up and process it to for installation.

say-paul:
Extended install to copy from the extend-payload.
 Build the image with bib, I could install uboot to esp.

```
λ sudo fdisk -l /dev/nbd0
Disk /dev/nbd0: 10 GiB, 10737418240 bytes, 20971520 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: D209C89E-EA5E-4FBD-B161-B461CCE297E0

Device        Start      End  Sectors  Size Type
/dev/nbd0p1    2048  1028095  1026048  501M EFI System
/dev/nbd0p2 1028096  3125247  2097152    1G Linux filesystem
/dev/nbd0p3 3125248 20971486 17846239  8.5G Linux filesystem
λ mkdir ~/qcow2-mount
sudo mount /dev/mapper/nbd0p1 ~/qcow2-mount
[~/git_repos/bootupd]$(testing-frimware-copy)
λ ls -l ~/qcow2-mount
total 688
drwxr-xr-x. 4 root root   8192 Jul  5 05:56 EFI
-rwxr-xr-x. 1 root root 690712 Jul  5 05:56 u-boot.bin
[~/git_repos/bootupd]$(testing-frimware-copy)
λ ls -al ~/qcow2-mount
total 704
drwxr-xr-x. 3 root    root     16384 Jan  1  1970 .
drwx--x---+ 1 saypaul saypaul   2038 Jul  5 06:17 ..
drwxr-xr-x. 4 root    root      8192 Jul  5 05:56 EFI
-rwxr-xr-x. 1 root    root    690712 Jul  5 05:56 u-boot.bin
[~/git_repos/bootupd]$(testing-frimware-copy)
```

say-paul:
bootupd-state.json: (adds 'firmware' to EFI) - probably should not have  EFI.json in fireware::filetree, which should be a minor fix.

```
λ cat ~/qcow2-mount-p2/bootupd-state.json | jq
{
  "installed": {
    "EFI": {
      "meta": {
        "timestamp": "2025-03-27T10:27:59Z",
        "version": "grub2-efi-aa64-1:2.12-28.fc42.aarch64,shim-aa64-15.8-3.aarch64"
      },
      "filetree": {
        "children": {
          "BOOT/BOOTAA64.EFI": {
            "size": 971654,
            "sha512": "sha512:8891ba415802b97828b7939b81c0235610f3648fd257df8e4b754e1fa4899da9e202fd1be95bbcb095816525b601a6ecfa3ec94be08de5f5fa1911cffc683d13"
          },
          "BOOT/fbaa64.efi": {
            "size": 91096,
            "sha512": "sha512:a1289697eb9686c171446fd7ae2b1fd7358b7aa8de229f6d22993852f4f5be41f0b783f50b33505167014e85b07deadd6fed2533c68eb4577743bf6c5df61ff3"
          },
          "fedora/BOOTAA64.CSV": {
            "size": 112,
            "sha512": "sha512:4ceedbfec87e417308a71313d22b6b81aa3b00c70f55d35c18bd97e88d45c0429032cb822b31340a8c105e47bb45d4c44871d6de29bb397d324b735752c42b7c"
          },
          "fedora/grubaa64.efi": {
            "size": 4210160,
            "sha512": "sha512:2e84d665dbac56f4b0fa369023841c8e8c16bd0d68f0f3bcbfdfeae7236f5396e03d1852238d85b3a2e0d3c4f64d443bf66d5f79e1fa22e304440a9a72fb3e7c"
          },
          "fedora/mmaa64.efi": {
            "size": 886296,
            "sha512": "sha512:fda90b956737d74b6d66b491b66573bd47f857da1a2ef8e619167173b1b870991b0a904961714462794dcaebfbd0d026c0bd8afdfcdb21932f0a7c43163855d4"
          },
          "fedora/shim.efi": {
            "size": 971654,
            "sha512": "sha512:8891ba415802b97828b7939b81c0235610f3648fd257df8e4b754e1fa4899da9e202fd1be95bbcb095816525b601a6ecfa3ec94be08de5f5fa1911cffc683d13"
          },
          "fedora/shimaa64.efi": {
            "size": 971654,
            "sha512": "sha512:8891ba415802b97828b7939b81c0235610f3648fd257df8e4b754e1fa4899da9e202fd1be95bbcb095816525b601a6ecfa3ec94be08de5f5fa1911cffc683d13"
          }
        }
      },
      "adopted-from": null,
      "firmware": {
        "uboot": {
          "meta": {
            "timestamp": "2025-04-20T21:03:54Z",
            "version": "uboot-images-armv8-1:2025.04-2.fc42.noarch"
          },
          "filetree": {
            "children": {
              "EFI.json": {
                "size": 91,
                "sha512": "sha512:09f2147475392cea035a408e269c13dcf8d7ef2d3c7a9fd34328131de179c14c2dd27510f3c22807d571bfa7ad151a1b0787ad0b5a4113be3dfc98a4d8951c3f"
              },
              "u-boot.bin": {
                "size": 690712,
                "sha512": "sha512:10fe3903b2f94a3e3c1e7e95009388c2628c50a84ebe98d8c43f370856a0250feaa44044df92bea293a1612703e253e8d9d4a60192d30d355a2958133986fafe"
              }
            }
          },
          "adopted-from": null
        }
      }
    }
  },
  "pending": null,
  "static-configs": {
    "timestamp": "2025-07-04T23:50:49Z",
    "version": "0.2.28"
  }
}
```

say-paul:
@cgwalters I have updated the  comments [here](https://github.com/coreos/bootupd/pull/935#issue-3093610029) with the details + example.

kfox1111:
Curious about a mode dealing with a rpi config.txt. There a way to have it land one and then not update it if its there already or alternately only if unchanged?

HuijingHei:
Since Colin and Timothee are busy with other things, I can help to see this. Let me know if I misunderstood it.

So does this mean that you only want to extend single file `/usr/share/uboot/rpi_arm64/u-boot.bin`,  not files under `/boot/efi` like `bcm2711-firmware`, is this right?

say-paul:
> So does this mean that you only want to extend single file /usr/share/uboot/rpi_arm64/u-boot.bin, not files under /boot/efi like bcm2711-firmware, is this right?

@HuijingHei On high level i want to copy content of a dir to /boot/efi that includes both containers: /boot/efi  and /usr/share/uboot/rpi_arm64/ to esp of the image.

this is the complete usecase : https://github.com/fedora-iot/iot-distro/issues/82#issuecomment-3051419262

HuijingHei:
> > So does this mean that you only want to extend single file /usr/share/uboot/rpi_arm64/u-boot.bin, not files under /boot/efi like bcm2711-firmware, is this right?
> 
> @HuijingHei On high level i want to copy content of a dir to /boot/efi that includes both containers: /boot/efi and /usr/share/uboot/rpi_arm64/ to esp of the image.
> 
> this is the complete usecase : [fedora-iot/iot-distro#82 (comment)](https://github.com/fedora-iot/iot-distro/issues/82#issuecomment-3051419262)

Agree with Colin and Timothee that there are partly duplication with https://github.com/coreos/bootupd/issues/926, e.g. `stages the input files to /usr/lib/efi/firmware/{name}/{version}/EFI/`, maybe we should make them independent functions and let both can reuse it. Would you like to change this (or I will do the change)? And it would be better to add test for [comment](https://github.com/fedora-iot/iot-distro/issues/82#issuecomment-3051419262) in the PR. WDYT?

src/efi.rs
```diff
@@ -462,6 +622,84 @@ impl Component for Efi {
         Ok(meta)
     }
 
+    fn extend_payload(&self, sysroot_path: &str, src_input: &str) -> Result<Option<bool>> {
+        let dest_efidir_base = Path::new(sysroot_path).join("usr/lib/efi").join("firmware");
```
say-paul, line 660: @HuijingHei  do you suggest overloading  the function  to put files in `usr/lib/efi/(firmware|shim|grub)`

bam80:
Anything left?

.github/workflows/ci.yml
```diff
@@ -139,3 +139,196 @@ jobs:
               bootupctl backend generate-update-metadata -vvv
               cat ${updates}/EFI.json | jq
             '
+
+      - name: Test install after extend-payload
+        run: |
+          set -xeuo pipefail
+          sudo truncate -s 5G myimage-extend.raw
+          sudo podman run --rm --privileged -v .:/target --pid=host --security-opt label=disable \
```
cgwalters, line 147: This is not at all a new problem, and I won't say we should block on this but...

This is greatly exacerbating the problem that it's just a pain to run these tests outside of GHA (short of https://github.com/nektos/act etc.)

There's an effort to adapt our tests to TMT, but that gets into larger CI issues.

Anyways again not a blocker, but just noting.

.github/workflows/ci.yml
```diff
@@ -139,3 +139,196 @@ jobs:
...
+            -v /dev:/dev \
+            localhost/bootupd:latest bash -c '
+              # Create test firmware directory and files
+              mkdir -p /usr/share/uboot/rpi/overlays
+              echo "test uboot binary content" > /usr/share/uboot/rpi/u-boot.bin
+              echo "i2c device tree overlay" > /usr/share/uboot/rpi/overlays/i2c.dtb
+              
+              # Create a fake RPM database for testing
+              mkdir -p /usr/lib/sysimage/rpm
+              echo "fake rpm database" > /usr/lib/sysimage/rpm/Packages
```
cgwalters, line 158: Hmmm....this seems highly likely to confuse other code in bootupd at some point.

As is right now, I am pretty sure it will result in bootupd not installing shim and grub, right?

.github/workflows/ci.yml
cgwalters: Actually thinking about this a bit more, why can't this test work how we'd expect things to work in production?

Basically if we change this to be a `podman build` that installs the files instead that should all Just Work, right?

src/cli/bootupctl.rs
```diff
@@ -73,6 +73,8 @@ pub enum CtlBackend {
     Generate(super::bootupd::GenerateOpts),
     #[clap(name = "install", hide = true)]
     Install(super::bootupd::InstallOpts),
+    #[clap(name = "extend-payload-to-esp", hide = true)]
```
cgwalters, line 76: I don't think we should `hide = true` this - it's a user visible feature.

src/efi.rs
```diff
@@ -860,4 +1103,170 @@ Boot0003* test";
...
+                println!("✓ Package version: {}", parsed.version);
+            }
+            Ok(Some(false)) => {
+                panic!("extend_payload returned false - expected success");
+            }
+            Ok(None) => {
+                panic!("extend_payload returned None - expected success");
+            }
+            Err(e) => {
+                panic!("extend_payload failed when it should have succeeded: {}", e);
```
cgwalters, line 1266: It'd avoid all this nesting to just use `.unwrap()` or `.expect()` on the result which seems preferable to me.
And we may as well do the same for the `Option`

src/efi.rs
```diff
@@ -860,4 +1103,170 @@ Boot0003* test";
...
+        {
+            use std::os::unix::fs::PermissionsExt;
+            let mut perms = fs::metadata(&mock_rpm_script)?.permissions();
+            perms.set_mode(0o755);
+            fs::set_permissions(&mock_rpm_script, perms)?;
+        }
+
+        let original_path = std::env::var("PATH").unwrap_or_default();
+        let new_path = format!("{}:{}", mock_rpm_dir.path().display(), original_path);
+        std::env::set_var("PATH", &new_path);
```
cgwalters, line 1177: This is fully `unsafe` for good reason in Rust 2024: https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html#stdenvset_var-remove_var

It's more painful but basically if we wanted to mock RPM we'd have to explicitly pass it as a parameter.

src/efi.rs
```diff
@@ -506,6 +656,99 @@ impl Component for Efi {
...
+
+        let meta_from_src = packagesystem::query_files(sysroot_path, [path_to_query])
+            .context(format!("Querying RPM metadata for {:?}", src_input_path))?;
+
+        let version_string_part =
+            meta_from_src.version.split(',').next().ok_or_else(|| {
+                anyhow!("RPM query returned an empty or malformed version string")
+            })?;
+
+        let parts: Vec<&str> = version_string_part.split('-').collect();
```
cgwalters, line 682: Hmm @HuijingHei is reworking the version parsing in https://github.com/coreos/bootupd/pull/977 etc.

src/efi.rs
```diff
@@ -506,6 +656,99 @@ impl Component for Efi {
...
+        let (pkg_name, version_release_str) = if parts.len() >= 3 {
+            (
+                parts[0].to_string(),
+                format!(
+                    "{}-{}",
+                    parts[parts.len() - 2],
+                    parts[parts.len() - 1]
```
cgwalters, line 689: Try using e.g.

```
match parts.as_slice() {
  [name, version, release] => { ... }
  _ => anyhow::bail!()
```

src/efi.rs
```diff
@@ -409,14 +466,106 @@ impl Component for Efi {
...
-            let destpath = &self.ensure_mounted_esp(rootcxt.path.as_ref(), Path::new(&esp))?;
+        for esp in esp_devices.iter() {
+            let destpath = &self.ensure_mounted_esp(rootcxt.path.as_ref(), Path::new(esp))?;
             let destdir = openat::Dir::open(&destpath.join("EFI")).context("opening EFI dir")?;
             validate_esp_fstype(&destdir)?;
             log::trace!("applying diff: {}", &diff);
             filetree::apply_diff(&updated, &destdir, &diff, None)
                 .context("applying filesystem changes")?;
+            // update firmware
+            let firmware_base_dir_path = Path::new("usr/lib/efi/firmware");
```
cgwalters, line 479: Where does this path come from? Is it something you made up here?

src/efi.rs
```diff
@@ -377,15 +379,70 @@ impl Component for Efi {
             .arg(destpath)
             .current_dir(format!("/proc/self/fd/{}", src_root.as_raw_fd()))
             .run()?;
+
+        let mut found_firmware = BTreeMap::new();
+        // Scan and install supplemental firmware
+        let firmware_base_dir_path = Path::new("usr/lib/efi/firmware");
+        if src_root.exists(firmware_base_dir_path)? {
```
cgwalters, line 386: The code past this is getting really deeply nested, can we factor out the inner part here into a helper function?

cgwalters (commented):
Sorry for the delay in review and thanks for working on this.

I have some initial comments.

alexlarsson:
> We had a chat a while back and this should now be possible with what we have in #938.

@travier Can #938 be used to copy firmware to the root of the esp? It looks like it can only copy into the EFI subdirectory?

martinezjavier:
@say-paul I've proposed https://github.com/coreos/bootupd/pull/1073 that I think complements your `extend-payload-to-esp` sub-command

PR #936, opened 2025-05-28
Update Dockerfile to use bootc

cgwalters:
- This is just way easier to understand and use
- This also specifically inherits the bits we have in bootc to set up rust to cache to a volume so incremental builds work


Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+set -xeuo pipefail
+dnf -y install cargo git openssl-devel
+EORUN
+# Now copy the source
+COPY . /build
+WORKDIR /build
+# See https://www.reddit.com/r/rust/comments/126xeyx/exploring_the_problem_of_faster_cargo_docker/
+# We aren't using the full recommendations there, just the simple bits.
+RUN --mount=type=cache,target=/build/target --mount=type=cache,target=/var/roothome \
+    make && make install DESTDIR=/out
```
HuijingHei, line 20: Overall LGTM, one minor comment, do we need `make install install-grub-static install-systemd-unit DESTDIR=/out`?

HuijingHei (approved):

HuijingHei:
> Check that `centos-bootc:stream9` creates 2 partitions: ESP and root , ESP part should be `/dev/loop0p1`, but `centos-bootc-dev:stream9` creates 3 parts with first reserved part, ESP part should be `/dev/loop0p2`.
> 
> Find `centos-bootc-dev:stream9` is too old that with `stream9.20240610.0` and did not update for almost 1 year, it is reasonable to use `centos-bootc:stream9`, and should fix to use modern `/dev/loop0p1`.

I have no permission to push here, create new https://github.com/coreos/bootupd/pull/937, which should fix the error.

.github/workflows/ci.yml
```diff
@@ -28,6 +28,13 @@ jobs:
     runs-on: [ "${{ matrix.runner }}" ]
 
     steps:
+      - name: Get a newer podman for heredoc support (from debian testing)
+        run: |
+          set -eux
+          echo 'deb [trusted=yes] https://ftp.debian.org/debian/ testing main' | sudo tee /etc/apt/sources.list.d/testing.list
+          sudo apt update
+          sudo apt install -y crun/testing podman/testing skopeo/testing
+
```
HuijingHei, line 37: Sorry that I might misunderstood, does this mean to install debian packages on Ubuntu?

PR #937, opened 2025-05-29
ci: get esp part using `sfdisk`

HuijingHei:


cgwalters:
Thanks for tracking this down!

Issue #933, opened 2025-05-21
"Ignoring downgrade" with older grub2-efi installed


miabbott:
I upgraded my system to Fedora 42 (specifically a custom build of Silverblue 42 using the Universal Blue tooling), which includes `grub2-efi-ia32-2.12-28.fc42.x86_64` and `grub2-efi-x64-2.12-28.fc42.x86_64`, etc

When I check the output of `bootupctl status`, I see the following:

```
$ sudo bootupctl status 
Running as unit: bootupd.service
Component EFI
  Installed: grub2-efi-ia32-1:2.12-21.fc41.x86_64,grub2-efi-x64-1:2.12-21.fc41.x86_64,shim-ia32-15.8-3.x86_64,shim-x64-15.8-3.x86_64
  Update: Ignoring downgrade
No components are adoptable.
Boot method: EFI
```

It looks like the `shim` packages have successfully upgraded the EFI binaries, but the `grub2-efi` packages have not.

When I try `bootupctl update`, it reports no updates avaiilable.

```
$ sudo bootupctl update
Running as unit: bootupd.service
No update available for any component.
```

Nothing in the journal points to a failure or error.

cgwalters:
Parsing version numbers gets into the thorny issue of cross-distribution version strings. There's of course https://uapi-group.org/specifications/specs/version_format_specification/ and maybe we could try to constrain our inputs to match that.

cgwalters:
This is I think probably a relatively unusual case in that the build for Fedora N-1 is chronologically newer than the version for Fedora N.

However...instead of going all the way to comparing version numbers in the same way RPM does, we could simply incorporate the data from `/usr/lib/os-release` into our calculations. If the combination of either `$ID` or `$VERSION_ID` have changed, we assume that we need to apply the update, regardless of timestamp.

But *inside* an "$ID-$VERSION_ID" (an OS stream/release/version) I think it's pretty safe to just rely on timestamps and not tie ourselves to any specific package system comparison method.

cgwalters:
You are right that trying to adapt to the `/usr/lib/efi` format would force comparing version numbers.

However, one option here is to suggest that OSes using this conform to https://uapi-group.org/specifications/specs/version_format_specification/ instead.

One way to do that I think for Fedora would be to start by ignoring the epoch. If a component there ever needed to bump epoch, we'd have to instead map it to one of `~` or `^`.

But actually since nowadays RPM supports those too I think we could just hopefully not ever bump epoch again.

PR #938, opened 2025-05-29
Extend `generate-update-metadata()` to read from `/usr`

HuijingHei:
For `generate-update-metadata()`, we move legacy dir from `usr/lib/ostree-boot/efi/EFI` to `usr/lib/bootupd/updates/`, with this patch, will move `usr/lib/efi/(shim|grub)/%{version}-%{release}/EFI/`
 to `usr/lib/bootupd/updates/`.

Still need to query the RPM db to get metadata.

See https://github.com/coreos/bootupd/issues/926


HuijingHei:
Ready for reviewing now.

I did testing under fedora-bootc container, and upgrade `grub2-efi-x64 shim-x64` (to https://github.com/coreos/bootupd/issues/926#issuecomment-2913933348), remove the old `/usr/lib/bootupd/updates/*` and run `generate-update-metadata` to create metadata.

```
# cat myrepo.repo
[myrepo]
name=My Custom Repository
baseurl=http://localhost:8080/bootloader/
enabled=1
gpgcheck=0
```
Build code:
```
# make && make install-all DESTDIR=/srv/bootupd-test/out
```
Run testing under fedora-bootc container:
```
podman run --privileged -it -v ./:/opt -w /opt -v ./tmp/myrepo.repo:/etc/yum.repos.d/myrepo.repo --net host quay.io/fedora/fedora-bootc:42 bash


bash-5.2# dnf upgrade -y grub2-efi-x64 shim-x64 --repo myrepo

bash-5.2# rpm -ql grub2-efi-x64 shim-x64 | grep usr/lib/efi
/usr/lib/efi/grub2/2.12-34.fc43/EFI/fedora/grubx64.efi
/usr/lib/efi/shim/15.8-4/EFI/BOOT/BOOTX64.EFI
/usr/lib/efi/shim/15.8-4/EFI/BOOT/fbx64.efi
/usr/lib/efi/shim/15.8-4/EFI/fedora/BOOTX64.CSV
/usr/lib/efi/shim/15.8-4/EFI/fedora/mmx64.efi
/usr/lib/efi/shim/15.8-4/EFI/fedora/shim.efi
/usr/lib/efi/shim/15.8-4/EFI/fedora/shimx64.efi

dnf reinstall -y grub2-efi-x64 shim-x64 --repo my
bash-5.2# cp -rav out/usr /
bash-5.2# rm -rf /usr/lib/bootupd/updates/*

bash-5.2# /usr/bin/bootupctl backend generate-update-metadata / -vvv
[TRACE bootupd] executing cli
[DEBUG bootupd::ostreeutil] Using dbpath "--dbpath=/usr/lib/sysimage/rpm"
Generated update layout for BIOS: grub2-tools-1:2.12-34.fc43.x86_64
[DEBUG bootupd::ostreeutil] Using dbpath "--dbpath=/usr/lib/sysimage/rpm"
Generated update layout for EFI: grub2-efi-x64-1:2.12-34.fc43.x86_64,shim-x64-15.8-4.x86_64
[DEBUG bootupd::efi] Unmounting

### check the result:
bash-5.2# cat /usr/lib/bootupd/updates/BIOS.json 
{"timestamp":"2025-05-30T03:39:14Z","version":"grub2-tools-1:2.12-34.fc43.x86_64"} 
bash-5.2# cat /usr/lib/bootupd/updates/EFI.json 
{"timestamp":"2025-05-30T06:29:19Z","version":"grub2-efi-x64-1:2.12-34.fc43.x86_64,shim-x64-15.8-4.x86_64"}
```

src/efi.rs
```diff
@@ -381,10 +382,42 @@ impl Component for Efi {
     }
 
     fn generate_update_metadata(&self, sysroot_path: &str) -> Result<ContentMetadata> {
-        let ostreebootdir = Path::new(sysroot_path).join(ostreeutil::BOOT_PREFIX);
         let dest_efidir = component_updatedir(sysroot_path, self);
 
-        if ostreebootdir.exists() {
+        // New EFI dir /usr/lib/efi
```
cgwalters: Do you know if this this specific to the new Fedora grub/shim packages? I suspect it is...we may want to somehow make this a build time or even runtime conditional so in theory it's more pluggable/controllable by others.

I'd at least factor it out into a `const` somewhere that explains where it came from.

Hmm actually, this topic also strongly relates to https://github.com/coreos/bootupd/issues/766 right?

src/efi.rs
```diff
@@ -381,10 +382,42 @@ impl Component for Efi {
...
+
+                    Command::new("mv")
+                        .arg(&efi)
+                        .arg(&dest_efidir)
+                        .run_with_cmd_context()?;
+                    acc.append(&mut files);
+                    Ok(acc)
+                });
+            packagesystem::query_files(sysroot_path, all_files?.into_iter())?
+        } else if ostreebootdir.exists() {
```
cgwalters: Would it be cleaner perhaps if we checked first for `ostreebootdir`, and migrated it to `usr/lib/efi` if that doesn't exist? It should be an error if both exist.

Then we get closer to thinking of the `ostreebootdir` one as legacy.

src/efi.rs
```diff
@@ -381,10 +382,42 @@ impl Component for Efi {
...
+                    // Get all files path
+                    let mut files = crate::util::filenames(&efidir)?
+                        .into_iter()
+                        .map(|mut f| {
+                            f.insert_str(0, efi.to_str().unwrap());
+                            f
+                        })
+                        .collect::<Vec<String>>();
+
+                    Command::new("mv")
```
cgwalters: I think we have an opportunity here to make `usr/lib/efi` the standard path actually if it exists, maybe we could make things as simple as the equivalent of `ln -sr /usr/lib/efi /usr/lib/bootupd/updates/EFI`?

EDIT: Ah I see it's not that simple based on `find_all_efi_dirs`.

Hmmm...but actually I like the idea of that layout, what if we tried to adopt *that* as the standard and migrate our current EFI layout to it? (It'd break updates for older bootupd though...without having dual layouts for a while, ug)

src/efi.rs
```diff
@@ -615,6 +656,29 @@ fn find_file_recursive<P: AsRef<Path>>(dir: P, target_file: &str) -> Result<Vec<
     Ok(result)
 }
 
+// Find EFI dirs under usr/lib/efi
+// for exmaple: usr/lib/efi/shim/15.8-4/EFI, usr/lib/efi/grub2/2.12-34.fc42/EFI
+fn find_all_efi_dirs(sysroot_lib: &Path) -> Result<Vec<PathBuf>> {
```
cgwalters: BTW looking at this layout, I think we can now avoid invoking `rpm` to query the file information for these which would be a huge side benefit.

So again I think this relates to https://github.com/coreos/bootupd/issues/766 in that perhaps we make this layout our new "API" for adding content in the ESP?

Actually thinking about things here...you know, it probably wouldn't be terribly hard to change what rpm-ostree does to automatically do this instead (via an opt-in). That'd require some coordination but the powerful benefit is we'd effectively automatically "backport" support for `/usr/lib/efi` even for older OSes which seems like it'd help us a lot here.

cgwalters (approved):
Overall looks sane to me, a variety of comments. I don't have anything blocking.

I think we should be testing this scenario in CI, I doubt we are? I am not up to date on things, does it require us pulling in a copr?

src/efi.rs
HuijingHei: Thank you so much for the reviewing!

> Do you know if this this specific to the new Fedora grub/shim packages? I suspect it is...we may want to somehow make this a build time or even runtime conditional so in theory it's more pluggable/controllable by others.

Yes. 
> 
> I'd at least factor it out into a `const` somewhere that explains where it came from.

SGTM.
> Hmm actually, this topic also strongly relates to #766 right?

Actually this is related to issue https://github.com/coreos/bootupd/issues/926#issuecomment-2913933348, but we can extend it support #766 too.

src/efi.rs
```diff
@@ -615,6 +656,29 @@ fn find_file_recursive<P: AsRef<Path>>(dir: P, target_file: &str) -> Result<Vec<
...
+    const LIBDIRS: &[&str] = &["grub2", "shim"];
```
say-paul: Can we have : `const LIBDIRS: &[&str] = &["grub2", "shim","."];`
So that we can keep in things in `/usr/lib/efi/<version>/EFI/<files>` ,  EFI.json metadata can contain the rpm details.

src/efi.rs
HuijingHei: > Hmmm...but actually I like the idea of that layout, what if we tried to adopt _that_ as the standard and migrate our current EFI layout to it? (It'd break updates for older bootupd though...without having dual layouts for a while, ug)

Yes, you are right, that will break old bootupd. The current EFI.json is like:
```
# cat EFI.json | jq
{
  "timestamp": "2025-03-27T10:27:15Z",
  "version": "grub2-efi-x64-1:2.12-28.fc42.x86_64,shim-x64-15.8-3.x86_64"
}
```
We only get `/usr/lib/efi/grub2/2.12-34.fc43/EFI` (-> `grub2-2.12-34.fc43`) and `/usr/lib/efi/shim/15.8-4/EFI` (-> `shim-15.8-4`) from the path without rpmdb, does this make sense? For silverblue we install both `shim-ia32` and `shim-x64`, it only shows once, we do not care about this if we only concerns the version.
```
{
  "timestamp": "<now>",
  "version": "grub2-2.12-34.fc43,shim-15.8-4"
}
```

When we do the update, we sync all the files under `/usr/lib/bootupd/updates/EFI` to `/boot/efi/EFI`, means we only apply once, but if there are 2 or more directories, we need to sync each EFI directory, any good suggestion for this?

src/efi.rs
HuijingHei: Maybe can remove the limitation and scan all EFI directories like `/usr/lib/efi/<name>/<version>/EFI/<files>` ?

src/efi.rs
HuijingHei: Want to clarify this, what we currently do in `generate-update-metadata()` is move files under legacy ostree  `/usr/lib/ostree-boot/efi/EFI` to `/usr/lib/bootupd/updates/EFI`, then invoke rpm to query the file information (need to insert the `/boot/efi/EFI` to get the correct path) and create `EFI.json` to include the package info

What we want to change is: retrieve `/usr/lib/efi` and get EFI path, then add the package info from path and create EFI.json. The change might be easy.

What I am concern is for the installation and update, if there are 2 or more directories, we need to change the current logic to sync each EFI directory, instead of only `/usr/lib/bootupd/updates/EFI`. WDYT?

src/efi.rs
say-paul: That can be one method,  we would only need to find a name for putting things to esp, which is trivial.

travier: Let's not consider files outside of a `<name>/<version>` directory here as we need to be able to know "where" they come from / attach they to a package/source, even if it's user specified.

src/efi.rs
cgwalters: > What we want to change is: retrieve /usr/lib/efi and get EFI path, then add the package info from path and create EFI.json. The change might be easy.

Yeah I think so.

> What I am concern is for the installation and update, if there are 2 or more directories, we need to change the current logic to sync each EFI directory, instead of only /usr/lib/bootupd/updates/EFI. WDYT?

Yeah, I think that would make sense. However it would mean we need to bridge between the current list of files in `bootupd-state.json` vs the split directories.

Anyways in the *short* term what you're doing here (not changing the payload layout) is probably what we have to do in order to retain backwards compat (i.e. older clients can upgrade).

But after this work lands it'd probably be useful to try to start some work on making `/usr/lib/efi` style layout be supported; something like `bootupctl backend install --format-version=2` as an opt in or so.

HuijingHei: > start some work on making `/usr/lib/efi` style layout be supported;

SGTM.

src/efi.rs
say-paul: is reliance on `<name>/<version>` to remove dependency on `rpm -qf` which does not work when files are copied around?

HuijingHei: > is reliance on `<name>/<version>` to remove dependency on `rpm -qf` which does not work when files are copied around?

Yes, I think it will not work. In future, will rely on that to get meta for package.

HuijingHei:
Look at more about https://github.com/coreos/bootupd/pull/938#discussion_r2136203241, to clarify:

- Firstly checks `/usr/lib/efi/(shim|grub)/%{version}-%{release}/EFI/` EFI path exists, then add EFI.json about the "(shim|grub)/%{version}-%{release}"
- If the path `/usr/lib/efi` doesn’t exist, check the legacy OSTree location `/usr/lib/ostree-boot/efi/EFI/`, use rpm to query the files and translate them like `/usr/lib/efi/(shim|grub)/%{version}-%{release}/EFI/`, then add EFI.json about the "(shim|grub)/%{version}-%{release}"

- For the installed /boot/efi files, use rpm to query the files and translate them to `/usr/lib/efi/(firmware)/%{version}-%{release}/EFI/boot/efi/*`, also append EFI.json about "(firmware)/%{version}-%{release}", like:
```
{
  "timestamp": "2025-03-27T10:27:59Z",
  "version": "grub2-2.12-28.fc42,shim-15.8-3,bcm2711-firmware-20250319-1.80f23d4.fc42,bcm2835-firmware-20250319-1.80f23d4.fc42,bcm283x-firmware-20250319-1.80f23d4.fc42,bcm283x-overlays-20250319-1.80f23d4.fc42"
}
bash-5.2# dnf install -y bcm2711-firmware
Package                                                   Arch           Version                                                   Repository                            Size
Installing:
 bcm2711-firmware                                         aarch64        20250319-1.80f23d4.fc42                                   fedora                            10.3 MiB
Installing dependencies:
 bcm2835-firmware                                         aarch64        20250319-1.80f23d4.fc42                                   fedora                            12.1 MiB
 bcm283x-firmware                                         aarch64        20250319-1.80f23d4.fc42                                   fedora                            73.6 KiB
 bcm283x-overlays                                         aarch64        20250319-1.80f23d4.fc42                                   fedora                             1.0 MiB
...
bash-5.2# find /boot/efi/ -type f -exec rpm -qf {} \; 2>/dev/null | sort | uniq
bcm2711-firmware-20250319-1.80f23d4.fc42.aarch64
bcm2835-firmware-20250319-1.80f23d4.fc42.aarch64
bcm283x-firmware-20250319-1.80f23d4.fc42.aarch64
bcm283x-overlays-20250319-1.80f23d4.fc42.aarch64
```

Consequently, need to change the install to copy each directory to the destination "/boot/efi". WDYT?

HuijingHei:
> `generate-update-metadata` to accept any source_location(in this case:/usr/share/uboot/rpi_arm64)

Not sure how to point the file `/usr/share/uboot/rpi_arm64/u-boot.bin` precisely, or we only care about it? I see there are many `u-boot.bin` under different directories.

HuijingHei:
> The package installs the binary for all the boards, but rpi needs it be in `/boot/efi`
> 
> Also want to understand what will happen if: `cp /usr/share/uboot/rpi_arm64/u-boot.bin /boot/efi/u-boot.bin` is run before generate-update-metadata() is executed, as `rpm -qf` will show "not owned by any package", will it still be moved to /usr/lib/efi/firmware?

I think probably no, as it can not find the correct path.

HuijingHei:
> @HuijingHei will it be possible to add /usr/lib/efi/(grub|shim|**firmware**) to the parser as per the #935

Yes, I think so.

HuijingHei:
CI failed error:

```
error: boot data installation failed: installing component EFI: No such file or directory (os error 2)

[2025-06-23T09:02:20.994Z] Traceback (most recent call last):

[2025-06-23T09:02:20.994Z]   File "/run/osbuild/bin/org.osbuild.bootupd", line 82, in <module>

[2025-06-23T09:02:20.994Z]     r = main(_args, _args["options"])

[2025-06-23T09:02:20.994Z]   File "/run/osbuild/bin/org.osbuild.bootupd", line 75, in main

[2025-06-23T09:02:20.994Z]     subprocess.run(cmd, check=True)

[2025-06-23T09:02:20.994Z]     ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^

[2025-06-23T09:02:20.994Z]   File "/usr/lib64/python3.13/subprocess.py", line 577, in run

[2025-06-23T09:02:20.994Z]     raise CalledProcessError(retcode, process.args,

[2025-06-23T09:02:20.995Z]                              output=stdout, stderr=stderr)

[2025-06-23T09:02:20.995Z] subprocess.CalledProcessError: Command '['chroot', '/run/osbuild/mounts/ostree/deploy/fedora-coreos/deploy/81e9a6bc0f04476229d4887ac2316146fdee69ecebf8e9fa9b1824508bfa152b.0', '/usr/bin/bootupctl', 'backend', 'install', '--device=/dev/loop0', '--with-static-configs', '/run/osbuild/mounts']' returned non-zero exit status 1.
```

In this PR, we copy the EFI content from `ostree-boot` to new `/usr/lib/{pkg}/{version}/EFI`, not `/usr/lib/bootupd/updates`, so this failed.
One option is move the EFI content to `/usr/lib/bootupd/updates` and make bootupd works like before, and change in later PR teach bootupd to use new dir.
Another option is teach bootupd works to adapt the new `/usr/lib/{pkg}/{version}/EFI`.
WDYT?

cgwalters:
> One option is move the EFI content to /usr/lib/bootupd/updates and make bootupd works like before, and change in later PR teach bootupd to use new dir.

Yes per previous discussion it's a backwards incompatible change to use the new path - the old bootupd (before this PR) won't know what to do with data generated by bootupd after this PR.

Let's define some terms here:

## Layouts:

### /boot/efi/EFI

The default place for Fedora grub/shim RPM packages.

### /usr/lib/ostree-boot

Where rpm-ostree moves the above to by default today and what bootupd consumes.

### /usr/lib/bootupd/updates/EFI

This is content owned by bootupd.

### /usr/lib/efi/&lt;component&gt;/&lt;version&gt;

What the proposed new Fedora layout uses (and we could try to champion cross-OS/distro).

---

So for now we need to go:

/usr/lib/ostree-boot -> /usr/lib/bootupd/updates/EFI
/usr/lib/efi/&lt;component&gt; -> /usr/lib/bootupd/updates/EFI

But ideally in a future PR we teach bootupd to *also* know how to directly check for updates in /usr/lib/efi/&lt;component&gt; for EFI. We'll need to ship the code that knows how to consume those long before people can switch the default though.

HuijingHei:
> But ideally in a future PR we teach bootupd to _also_ know how to directly check for updates in /usr/lib/efi/<component> for EFI. We'll need to ship the code that knows how to consume those long before people can switch the default though.

You are right, look at more about this and above comments, what I did in this PR now:
- Transfer `/usr/lib/ostree-boot` -> `/usr/lib/efi/<component>` in `generate-update-metadata()`
- Extend `install(), update(), adopt()` to read files from `/usr/lib/efi/<component>`
WDYT?

HuijingHei:
Updated to use path instead of `rpm` when testing using new shim & grub
```
podman run --privileged -it -v ./:/opt -w /opt -v ./tmp/myrepo.repo:/etc/yum.repos.d/myrepo.repo --net host quay.io/fedora/fedora-bootc:42 bash
bash-5.2# cp -rav /opt/bootupd-test/out/usr /
bash-5.2# rm -rf /usr/lib/bootupd/updates/*
bash-5.2# dnf upgrade -y grub2-efi-x64 shim-x64 --repo myrepo
bash-5.2# bootupctl backend generate-update-metadata -vvv
[TRACE bootupd] executing cli
[DEBUG bootupd::ostreeutil] Using dbpath "--dbpath=/usr/lib/sysimage/rpm"
Generated update layout for BIOS: grub2-tools-1:2.12-34.fc43.x86_64
Generated update layout for EFI: grub2-2.12-34.fc43,shim-15.8-4
[DEBUG bootupd::efi] Unmounting
bash-5.2# cat /usr/lib/bootupd/updates/BIOS.json
{"timestamp":"2025-07-15T13:57:09Z","version":"grub2-tools-1:2.12-34.fc43.x86_64"}
bash-5.2# cat /usr/lib/bootupd/updates/EFI.json
{"timestamp":"2025-07-16T07:22:20.290457693Z","version":"grub2-2.12-34.fc43,shim-15.8-4"}
```

See the `EFI.json` is a little different from before
```
bash-5.2# cat /usr/lib/bootupd/updates/EFI.json
{"timestamp":"2025-06-17T16:35:00Z","version":"grub2-efi-x64-1:2.12-32.fc42.x86_64,shim-x64-15.8-3.x86_64"}
```

HuijingHei:
One issue about this PR is the old timestamp `1970-01-01T00:00:00Z`, which might hit https://github.com/coreos/bootupd/issues/933 when upgrade.
```
    "EFI": {
      "meta": {
        "timestamp": "1970-01-01T00:00:00Z",
        "version": "grub2-2.12-34.fc43,shim-15.8-4"
      },
```

travier:
Thanks a lot for this. This is looking good from a quick look. I'll test it tomorrow.

travier:
If I'm not mistaken, this is doing the same as https://gitlab.com/fedora/bootc/base-images/-/merge_requests/90 but more cleanly so I'll close that PR.

HuijingHei:
> If I'm not mistaken, this is doing the same as https://gitlab.com/fedora/bootc/base-images/-/merge_requests/90 but more cleanly so I'll close that PR.

Yes, I think so. Also add CI test and the EFI.json content is (from [log](https://github.com/coreos/bootupd/actions/runs/16511728134/job/46694847621?pr=938)):
```
{
  "timestamp": "2025-07-25T01:47:40.351545925Z",
  "version": "shim-15-15.el8_2,grub2-2.06-107.el9"
}
```
The timestamp is using now as https://github.com/coreos/bootupd/commit/1ec371ae3504b20891e1a03d29b24408d9623cd7 to workaround https://github.com/coreos/bootupd/issues/933, let me know if you have other way. Thank you so much for the kind review!

.github/workflows/ci.yml
```diff
@@ -106,3 +106,36 @@ jobs:
           # Verify we injected static configs
           jq -re '.["static-configs"].version' /boot/bootupd-state.json
           [ $(sudo stat -c "%a" /boot/grub2/grub.cfg) == "600" ]
+
+      - name: bootupctl generate-update-metadata
+        run: |
+          set -xeuo pipefail
+          # This will be helpful when we extend /usr
+          sudo podman run --rm -it --privileged localhost/bootupd:latest \
```
cgwalters, line 114: I've been trying harder to move load-bearing test logic out of GHA yaml and into something that's easier to run outside of GHA.

In this case this script could live in e.g. `tests/move-content-to-usr.sh` or something.

src/efi.rs
```diff
@@ -654,6 +698,29 @@ fn find_file_recursive<P: AsRef<Path>>(dir: P, target_file: &str) -> Result<Vec<
     Ok(result)
 }
 
+/// Get EFI path under usr/lib/efi, eg. usr/lib/efi/<package>/<version>/EFI
+fn get_efi_path_from_usr(
+    sysroot: &Utf8Path,
+) -> Result<impl Iterator<Item = Result<Utf8PathBuf>> + use<'_>> {
```
cgwalters, line 704: OK as is but I think this might be cleaner if we returned something like a

```
struct EFIComponent {
  name: String,
  version: String,
  path: Utf8PathBuf
}
```

or so

travier:
> The timestamp is using now as [1ec371a](https://github.com/coreos/bootupd/commit/1ec371ae3504b20891e1a03d29b24408d9623cd7) to workaround #933, let me know if you have other way. Thank you so much for the kind review!

I'm afraid that will work but not be what we want as that means that each new build will get a newer timestamp and we'll trigger an update on each boot? If I understand correctly [we don't check the version but only the timestamp](https://github.com/coreos/bootupd/issues/933#issuecomment-3020482129)? will have to check as I hadn't realized that.

travier:
So we should probably still query the RPM DB for now for those files to get the "real" timestamp.

src/efi.rs
HuijingHei: Nice suggestion, will update in later PR.

HuijingHei:
> So we should probably still query the RPM DB for now for those files to get the "real" timestamp. Or move to version comparison instead. There is an `rpm` crate that could help us do that: https://docs.rs/rpm/latest/rpm/struct.Evr.html

Agree to move version comparison and will not need to query RPM DB anymore. Thank you for the suggestion, will add that later.

HuijingHei:
> > The timestamp is using now as [1ec371a](https://github.com/coreos/bootupd/commit/1ec371ae3504b20891e1a03d29b24408d9623cd7) to workaround #933, let me know if you have other way. Thank you so much for the kind review!
> 
> I'm afraid that will work but not be what we want as that means that each new build will get a newer timestamp and we'll trigger an update on each boot?

No, if versions are equal (simple checking `current == target`), it will not trigger the update, this works if they are both in the same format (both from rpm DB or both from `/usr/lib/efi` path). If `current != target`, it will compare the timestamp. Agree we need to parse and compare the version (see https://github.com/coreos/bootupd/pull/977)

travier:
> > > The timestamp is using now as [1ec371a](https://github.com/coreos/bootupd/commit/1ec371ae3504b20891e1a03d29b24408d9623cd7) to workaround #933, let me know if you have other way. Thank you so much for the kind review!
> > 
> > 
> > I'm afraid that will work but not be what we want as that means that each new build will get a newer timestamp and we'll trigger an update on each boot?
> 
> No, if versions are equal (simple checking `current == target`), it will not trigger the update, this works if they are both in the same format (both from rpm DB or both from `/usr/lib/efi` path). If `current != target`, it will compare the timestamp. Agree we need to parse and compare the version (see #977)

OK, great, then we should probably make a release with this code so that we can have the changes land in GRUB & shim RPMs.

Issue #943, opened 2025-06-04
bootupctl adopt failed with BIOS only


HuijingHei:
Build the latest code, start VM using BIOS only, and no ESP part, bootupctl adopt failed
```
[root@localhost core]# bootupctl adopt-and-update
Running as unit: bootupd.service
Adopted and updated: BIOS: grub2-tools-1:2.06-104.el9_6.x86_64
error: Failed adopt and update: Failed to find adoptable system

[root@localhost core]# lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
loop0    7:0    0  5.7M  1 loop 
vda    252:0    0   16G  0 disk 
|-vda1 252:1    0    1M  0 part 
|-vda2 252:2    0    1G  0 part /boot
`-vda3 252:3    0   15G  0 part /var
                                /sysroot/ostree/deploy/rhcos/var
                                /etc
                                /sysroot

```

I think it will also fail if booting with UEFI only.

Edit:
Seems this is because of https://github.com/coreos/bootupd/pull/932/commits/fe53bab2d325c18e15faa871bdc44e228b742840, we add it to adopt list, and when do adoption, find it should skip.

HuijingHei:
Should be fixed with https://github.com/coreos/bootupd/pull/944, the next new version will include the fix.

PR #944, opened 2025-06-04
adopt: fix adoption when booting with BIOS firmware without ESP

HuijingHei:
We should support adoption when :
- booting with BIOS firmware without ESP
- booting with UEFI firmware without BIOS BOOT part

Fix regression issue https://github.com/coreos/bootupd/issues/943, which is introduced by https://github.com/coreos/bootupd/commit/fe53bab2d325c18e15faa871bdc44e228b742840


jlebon (approved):
Seems sane to me!

PR #945, opened 2025-06-05
migrate: add tag to install the canonical static GRUB config

HuijingHei:


src/bootupd.rs
```diff
@@ -702,6 +719,150 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
+
+        let mut state_guard = SavedState::acquire_write_lock(sysroot.try_clone()?)
+            .context("Failed to acquire write lock")?;
+        // Backup the current grub.cfg and install static config
+        {
+            const GRUBCONFIG: &str = "grub.cfg";
+            const BACKUP: &str = "grub.cfg.backup";
+            let grubconfig = grub_config_dir.join(GRUBCONFIG);
+
+            #[cfg(any(target_arch = "x86_64", target_arch = "powerpc64"))]
```
cgwalters: Nonblocking and not a new issue but we should probably centralize this into a `#[cfg(grub_arch)]` or so (I think we'd need to do that in a `build.rs`).

src/bootupd.rs
```diff
@@ -702,6 +719,150 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
+            {
+                // On BIOS, /boot/grub2/grub.cfg is symlink to /boot/loader/grub.cfg,
+                // should backup it to /boot/grub2/grub.cfg.bak
+                if !grubconfig.exists() {
+                    // Not existing, maybe UEFI
+                    println!("Not found '{}'", grubconfig.display());
+                    if cfg!(target_arch = "powerpc64") {
```
cgwalters: That said is powerpc64 actually relevant here? It looks like https://mirror.openshift.com/pub/openshift-v4/ppc64le/dependencies/rhcos/ only has a version starting from 4.3...was that not using static configs? (Ugh it's painful to inspect from the outside because that's using the old LUKS setup)

src/bootupd.rs
```diff
@@ -594,95 +603,103 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
+                let mut writer = BufWriter::new(
+                    dirfd
+                        .write_file(
+                            &stripped_config,
+                            (S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH) as mode_t,
+                        )
+                        .context("Failed to open temporary GRUB config")?,
+                );
+                let mut skip = false;
+                for line in BufReader::new(current_config_file).lines() {
```
cgwalters: This logic is pretty unit testable, and also probably worth splitting out into a separate function at least.

systemd/bootloader-migrate.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Static GRUB config migration for 4.1/4.2 born RHCOS nodes
+ConditionPathExists=!/boot/grub2/.grub2-static-migrated
+RequiresMountsFor=/sysroot /boot
+
+[Service]
+Type=oneshot
+ExecCondition=/bin/sh -c '[ "$(ostree config --repo=/sysroot/ostree/repo get sysroot.bootloader 2>/dev/null)" != "none" ]'
+ExecStart=bootupctl adopt-and-update
+ExecStart=/usr/bin/bootupctl migrate-static-grub-config --from-tree
```
cgwalters: What about having `adopt-and-update --with-static-config`?

Makefile
```diff
@@ -37,6 +37,7 @@ install-grub-static:
 .PHONY: install-systemd-unit
 install-systemd-unit:
 	install -m 644 -D -t "${DESTDIR}$(PREFIX)/lib/systemd/system/" systemd/bootloader-update.service
+	install -m 644 -D -t "${DESTDIR}$(PREFIX)/lib/systemd/system/" systemd/bootloader-migrate.service
```
cgwalters: It'd be nice to only enable this service for cases that need it which is basically RHCOS right?

I think a lot of cases now are using `install-systemd-unit` so this will have a large blast radius as is.

Now in theory of course the
`ExecCondition=/bin/sh -c '[ "$(ostree config --repo=/sysroot/ostree/repo get sysroot.bootloader 2>/dev/null)" != "none" ]'` will mean this new service just starts and then does nothing on each boot for those systems, but...it's probably worth turning this around and only actually installing the service there inside openshift/os or so?

cgwalters: And didn't we end up writing something similar to this for the atomic desktops? cc @travier

src/bootupd.rs
```diff
@@ -559,7 +564,11 @@ pub(crate) fn client_run_validate() -> Result<()> {
 }
 
 #[context("Migrating to a static GRUB config")]
-pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
+pub(crate) fn client_run_migrate_static_grub_config(from_tree: bool) -> Result<()> {
+    if cfg!(target_arch = "s390x") {
+        println!("Not supported migration");
```
cgwalters: It's more there's nothing to do because we don't use grub here. I think we can just drop this print.

systemd/bootloader-migrate.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Static GRUB config migration for 4.1/4.2 born RHCOS nodes
+ConditionPathExists=!/boot/grub2/.grub2-static-migrated
+RequiresMountsFor=/sysroot /boot
+
+[Service]
+Type=oneshot
+ExecCondition=/bin/sh -c '[ "$(ostree config --repo=/sysroot/ostree/repo get sysroot.bootloader 2>/dev/null)" != "none" ]'
```
cgwalters: I'd add above this `ConditionArchitecture=!s390x` or so.

systemd/bootloader-migrate.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Static GRUB config migration for 4.1/4.2 born RHCOS nodes
```
cgwalters: I think "4.1/4.2 born RHCOS nodes" isn't something that would make sense to general upstream bootc users. How about

```
# e.g. OpenShift RHCOS 4.1/4.2 nodes
Description=Static GRUB config migration
```

src/bootupd.rs
```diff
@@ -702,6 +719,150 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
+                        anyhow::bail!("Failed to find esp device");
+                    };
+
+                    if let Some(next_esp) = devices.next() {
+                        anyhow::bail!(
+                            "Found multiple esp devices {esp} and {next_esp}; not currently supported"
+                        );
+                    }
+                    esp
+                };
```
cgwalters: This is a copy of code from elsewhere right? It'd be good to have this in a single shared helper if so

src/bootupd.rs
```diff
@@ -702,6 +719,150 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
+
+                let mount_point = efi.ensure_mounted_esp(Path::new(root_path), Path::new(&esp))?;
+                let efi_path = mount_point.join(format!("EFI/{vendor}"));
+                let efidir = openat::Dir::open(&efi_path)
+                    .with_context(|| format!("Opening {}", efi_path.display()))?;
+                println!("Creating a backup of the current GRUB config on EFI");
+                efidir
+                    .copy_file(GRUBCONFIG, BACKUP)
+                    .context("Backuping grub.cfg")?;
+                crate::grubconfigs::install(&sysroot, Some(&vendor), false)?;
```
cgwalters: I notice in the above path we had a `syncfs` but not here?

cgwalters (commented):
Thanks for working on this! An initial review pass here

src/bootupd.rs
HuijingHei: Good catch, we should do the sync after grub installation.

src/bootupd.rs
HuijingHei: Like `#[cfg(grub_arch="bios")]` to match `#[cfg(any(target_arch = "x86_64", target_arch = "powerpc64"))]`,
and `#[cfg(grub_arch="efi")]` to match `#[cfg(any(
                target_arch = "x86_64",
                target_arch = "aarch64",
                target_arch = "riscv64"
            ))]` ?

systemd/bootloader-migrate.service
HuijingHei: Looks more sensible to me.

Makefile
travier: We added a drop-in for the existing unit for the Atomic Desktops: https://pagure.io/workstation-ostree-config/blob/main/f/bootupd.yaml#_17

travier: This is very much a "per-variant" thing here as to how to do this, so we should probably not install this by default.

systemd/bootloader-migrate.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+# e.g. OpenShift RHCOS 4.1/4.2 nodes
+Description=Static GRUB config migration
+ConditionArchitecture=!s390x
+ConditionPathExists=!/boot/grub2/.grub2-static-migrated
+RequiresMountsFor=/sysroot /boot
+
+[Service]
+Type=oneshot
+ExecCondition=/bin/sh -c '[ "$(ostree config --repo=/sysroot/ostree/repo get sysroot.bootloader 2>/dev/null)" != "none" ]'
```
travier: That feels too much on top of the stamp file check above? Why do we need both? Can you sync with Jonathan on the conditions for this unit?

src/bootupd.rs
HuijingHei: See https://github.com/coreos/bootupd/pull/947, also add other 2 functions about ostree config command.

systemd/bootloader-migrate.service
HuijingHei: Yes, I synced with @jlebon about the conditions, as do not want to run the service if `sysroot.bootloader` is none.

Makefile
cgwalters: (I am jetlagged right now and may be missing something obvious but) how is that different from what's being done here? Is it about fetching the static config from the new version in the target root?

src/bootupd.rs
cgwalters: Yeah something like that, anyways it's fine as is!

systemd/bootloader-migrate.service
jlebon: Yeah, I think with the new `adopt-and-update --with-static-config` approach, I'm OK dropping this. (The concern was to not unconditionally update the bootloader before even checking if we have work to do.)

Makefile
jlebon: Right. In the Atomic Desktop case, we want to "static-ify" a grub2-mkconfig-generated `grub.cfg`. In the CoreOS case, we just want to take the canonical GRUB configs from the rootfs. (The "static-ify the dynamic config" approach could work too in those old CoreOS systems of course, but since we don't care about dual boot in our case, we can afford to just use the canonical static configs instead and bring those nodes back in sync with modern CoreOS.)

jlebon: Hmm, unless I'm missing something, the line we're commenting on here is just installing the unit, not enabling it.

HuijingHei: Yes, by default the unit is disabled. We should make it enabled on RHCOS to let it running on each boot. I did testing with starting the service manually after boot.

systemd/bootloader-migrate.service
HuijingHei: ~Agree, we do not need it when change to adopt. Removed, thank you!~ Sorry that I was wrong, see Jonathan's above comment. Without the condition, it will run update even on 4.3+ ( but do not migrate static config as it is already `none`). If we only want to limit the update on 4.1/4.2, I think condition makes sense.

src/bootupd.rs
HuijingHei: I check 4.3 on ppc, it is already using `bootloader=none`, maybe better to remove the condition?

HuijingHei:
Update to `bootupctl adopt-and-update --with-static-config`, welcome review, thank you all!

Makefile
cgwalters: Hmm. This unit is not just enabling static configs, it's also doing the `adopt-and-update` flow i.e. updating the bootloader as well (which makes sense!).

Wouldn't it be simpler to structure this then as a drop-in for the `bootloader-update.service` unit? I think we'd trigger it on the firstboot path in OCP or so?

systemd/bootloader-migrate.service
HuijingHei: And it is not suitable to run on dual boot, for example on x86_64, after adopt for BIOS, the bootloader will be set to `none`, and will skip `EFI`, this will make the OS unboot if booting EFI.

Makefile
HuijingHei: > Wouldn't it be simpler to structure this then as a drop-in for the `bootloader-update.service` unit?

In this case, `bootloader-update.service` will run `update`, and we only need to enable static config for installed component, right? This is a little different from `adopt-and-update --with-static-config` that only enable static config for adoptable component. This would make `with-static-config` independently without update or adopt like the current `migrate`. Or also make `update --with-static-config` work, that means enable static for installed component. WDYT?

jlebon: Yeah, I'm not sure about structuring this as a drop-in for `bootloader-update.service`. But I think it's fine to carry the unit definition in OCP if for whatever reason we don't want it to live in bootupd. That probably makes sense anyway since the conditions under which you want to do this migration can be quite specific (and I guess is why e.g. no systemd unit was added as part of the Atomic Desktop migration work either).

systemd/bootloader-migrate.service
jlebon: Yes, sorry for misunderstanding (but for understanding my concern!). I thought that the new approach would _not_ update the bootloader unconditionally.

Even though we will soon turn on bootloader updates by default, I don't want to conflate that goal with this bug fix, and especially wouldn't want to do it in 4.19 now.

jlebon:
So, combining the two threads here. How about:
- We land `bootupctl adopt-and-update --with-static-config` here without having a unit.
- In OCP 4.19, we add a generator that outputs the migration unit only if `sysroot.bootloader` is `none`. That's nicer because (1) it minimizes as much as possible what runs on unaffected nodes; it won't even show up in the transaction, and (2) it gets rid of the gnarly `ExecCondition`. Accounting for EUS-to-EUS upgrades, we'd need to keep it for 4.20 as well but then could drop it in 4.21.

jlebon:
> In OCP 4.19, we add a generator that outputs the migration unit only if `sysroot.bootloader` is not `none`.

That should be I think "only if `sysroot.bootloader` is unset". s390x sets it to `zipl`. And if somehow it's set to some value other than `none` or `zipl`, we should probably just leave the system alone.

HuijingHei:
> I guess so, but meh... I'm also fine to just keep it in rhel-coreos-config. That would mean we wouldn't be able to remove it until 4.22 instead of 4.21, but that's fine. It probably make testing the final fix easier.

Could you help to review https://github.com/coreos/rhel-coreos-config/pull/23 ?

HuijingHei:
@jlebon I update the logic that will migrate to static grub config if `sysroot.bootloader` is not set instead of checking it is not `none`. Let me know if it is suitable, thank you!

src/grubconfigs.rs
```diff
@@ -13,7 +13,10 @@ use crate::freezethaw::fsfreeze_thaw_cycle;
 const GRUB2DIR: &str = "grub2";
 const CONFIGDIR: &str = "/usr/lib/bootupd/grub2-static";
 const DROPINDIR: &str = "configs.d";
+// The related grub files
 const GRUBENV: &str = "grubenv";
+pub(crate) const GRUBCONFIG: &str = "grub.cfg";
+pub(crate) const BACKUP: &str = "grub.cfg.backup";
```
jlebon: Should this be called `GRUBCONFIG_BACKUP` instead?

src/filetree.rs
jlebon, line 1: These hunks look unrelated/like they should be in a separate commit?

src/efi.rs
```diff
@@ -282,6 +309,25 @@ impl Component for Efi {
         log::trace!("applying adoption diff: {}", &diff);
         filetree::apply_diff(&updated, &destdir, &diff, None)
             .context("applying filesystem changes")?;
+
+        // Backup current config and install static config
+        if with_static_config {
+            // Installs the static config if the OSTree bootloader is either not set
+            // or is set to a value other than `none`.
+            if let Some(bootloader) = ostreeutil::get_ostree_bootloader()? {
+                if bootloader != "none" {
```
jlebon: I think the behaviour here should be the same as in the BIOS case, no?

src/efi.rs
```diff
@@ -282,6 +309,25 @@ impl Component for Efi {
...
+                    println!(
+                        "ostree repo 'sysroot.bootloader' config option is: '{}'",
+                        bootloader
+                    );
+                    self.migrate_static_grub_config(rootcxt.path.as_str(), destdir)?;
+                } else {
+                    println!(
+                        "Already using a static GRUB config, skipped adopting the static config"
+                    );
```
jlebon: Maybe for consistency also, we should have the print statements match between the BIOS and EFI case.

src/bios.rs
```diff
@@ -157,10 +159,62 @@ impl Component for Bios {
...
+    // - Backup "/boot/loader/grub.cfg" to "/boot/grub2/grub.cfg.bak"
+    // - Remove symlink "/boot/grub2/grub.cfg"
+    // - Install new static grub.cfg
+    // - Remove old "/boot/loader/grub.cfg"
+    fn migrate_static_grub_config(&self, sysroot_path: &str, destdir: &openat::Dir) -> Result<()> {
+        use anyhow::Context;
+        use openat_ext::OpenatDirExt;
+
+        let grub_config_dir = Utf8PathBuf::from(sysroot_path).join("boot/grub2");
+        let grubdir = destdir
```
jlebon: Hmm, I don't understand why we sometimes prefix `destdir` and sometimes not in the various path maniulations. If we want to assume that `destdir` can be something other than `/`, then I think some of the paths below need to be adjusted?

src/bios.rs
```diff
@@ -157,10 +159,62 @@ impl Component for Bios {
...
+                .context("Failed to backup GRUB config")?;
+            // Remove the symlink (on BIOS)
+            grubdir.remove_file_optional(grubconfigs::GRUBCONFIG)?;
+            crate::grubconfigs::install(&destdir, None, true)?;
+        } else {
+            anyhow::bail!("'{}' is not a symlink", grubconfig);
+        }
+
+        // Remove the unused file, backuped as /boot/grub2/grub.cfg.bak
+        destdir.remove_file_optional("boot/loader/grub.cfg")?;
```
jlebon: Presumably, this should instead be removing the target of the symlink instead of hardcoding.

src/bios.rs
```diff
@@ -157,10 +159,62 @@ impl Component for Bios {
...
+                current_config, backup_config
+            );
+            std::fs::copy(&current_config, &backup_config)
+                .context("Failed to backup GRUB config")?;
+            // Remove the symlink (on BIOS)
+            grubdir.remove_file_optional(grubconfigs::GRUBCONFIG)?;
+            crate::grubconfigs::install(&destdir, None, true)?;
+        } else {
+            anyhow::bail!("'{}' is not a symlink", grubconfig);
+        }
```
jlebon: I think one way to reflow this is

```rust
if !grubconfig.exists() {
    bail
} else if !grubconfig.is_symlink() {
    bail
}

// dedent
```

src/efi.rs
HuijingHei: Yes, good catch, forget to update this. Thank you for the review!

src/bios.rs
HuijingHei: Currently the destdir is `/`, and see https://github.com/coreos/bootupd/blob/main/src/bootupd.rs#L518 and [definition](https://github.com/coreos/bootupd/blob/main/src/bootupd.rs#L455).

HuijingHei: Agree it is confusing, update to "boot/grub2" to make it clear.

src/bios.rs
```diff
@@ -157,10 +159,66 @@ impl Component for Bios {
...
+
+        let grub_config = grub_config_path.join(grubconfigs::GRUBCONFIG);
+
+        // On BIOS, /boot/grub2/grub.cfg is symlink to /boot/loader/grub.cfg,
+        // backup it to /boot/grub2/grub.cfg.bak
+        if !grub_config.exists() {
+            anyhow::bail!("Not found '{}'", grub_config);
+        } else if !grub_config.is_symlink() {
+            anyhow::bail!("'{}' is not a symlink", grub_config);
+        } else {
```
jlebon, line 169: Minor/optional: mentioned this in a previous comment; since both paths above return, we don't strictly need this in an else block. That allows us to outdent one level.

src/efi.rs
```diff
@@ -245,11 +245,38 @@ impl Component for Efi {
...
+        };
+
+        // destdir is /boot/efi/EFI
+        let efidir = destdir
+            .sub_dir(&vendor)
+            .with_context(|| format!("Opening EFI/{}", vendor))?;
+        println!("Creating a backup of the current GRUB config on EFI");
+        efidir
+            .copy_file(grubconfigs::GRUBCONFIG, grubconfigs::GRUBCONFIG_BACKUP)
+            .context("Backuping grub.cfg")?;
```
jlebon: Maybe

```suggestion
            .context("Failed to backup GRUB config")?;
```

for consistency with BIOS.

src/bios.rs
```diff
@@ -157,10 +159,66 @@ impl Component for Bios {
...
+
+            // Backup the current GRUB config which is hopefully working right now
+            println!(
+                "Creating a backup of the current GRUB config '{}' in '{}'...",
+                current_config, backup_config
+            );
+            std::fs::copy(&current_config, &backup_config)
+                .context("Failed to backup GRUB config")?;
+            // Remove the symlink
+            grub_config_dir.remove_file(grubconfigs::GRUBCONFIG)?;
```
jlebon: Hmm, instead of removing the symlink here, can we just let `grubconfigs::install` overwrite it? That removes the window in which there is no GRUB config.

src/bios.rs
HuijingHei: Got it, sorry for the misunderstanding.

src/bios.rs
HuijingHei: SGTM, thank you!

HuijingHei:
Refer to  Jonathan's pointer https://gist.github.com/jlebon/7190e57cef4d705582ecdfa745bf1a66, I did testing using 4.1 BIOS/UEFI firmware with [bootloader-migrate.service](https://github.com/coreos/rhel-coreos-config/pull/23), install grub.cfg and bootuuid.cfg successfully.

On BIOS
```
[core@localhost ~]$ journalctl  -u bootloader-migrate.service | cat 
Jun 17 02:26:07 localhost.localdomain systemd[1]: Starting Static GRUB config migration...
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: ostree repo 'sysroot.bootloader' config option is not set yet
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Creating a backup of the current GRUB config '/boot/grub2/../loader/grub.cfg' in '/boot/grub2/grub.cfg.backup'...
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Added 01_users.cfg
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Added 05_ignition.cfg
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Added 10_blscfg.cfg
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Added 14_menu_show_once.cfg
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Added 30_uefi-firmware.cfg
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Added 41_custom.cfg
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Installed: grub.cfg
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Installed: bootuuid.cfg
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Static GRUB configuration has been adopted successfully.
Jun 17 02:26:11 localhost.localdomain bootupctl[864]: Adopted and updated: BIOS: grub2-tools-1:2.06-104.el9_6.x86_64
Jun 17 02:26:11 localhost.localdomain systemd[1]: Finished Static GRUB config migration.
```

on EFI
```
[root@localhost grub2]# journalctl  -u bootloader-migrate.service | cat 
Jun 17 02:47:04 localhost.localdomain systemd[1]: Starting Static GRUB config migration...
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: ostree repo 'sysroot.bootloader' config option is not set yet
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Creating a backup of the current GRUB config on EFI
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Added 01_users.cfg
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Added 05_ignition.cfg
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Added 10_blscfg.cfg
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Added 14_menu_show_once.cfg
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Added 30_uefi-firmware.cfg
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Added 41_custom.cfg
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Installed: grub.cfg
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Installed: bootuuid.cfg
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Installed: "redhat/grub.cfg"
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Installed: "redhat/bootuuid.cfg"
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Static GRUB configuration has been adopted successfully.
Jun 17 02:47:07 localhost.localdomain bootupctl[886]: Adopted and updated: EFI: grub2-efi-x64-1:2.06-104.el9_6.x86_64,shim-x64-15.8-4.el9_3.x86_64
Jun 17 02:47:07 localhost.localdomain systemd[1]: Finished Static GRUB config migration.
```

src/grubconfigs.rs
```diff
@@ -109,6 +112,8 @@ pub(crate) fn install(
                 println!("Installed: {target:?}");
             }
             fsfreeze_thaw_cycle(efidir.open_file(".")?)?;
+        } else {
+            println!("Not found /boot/efi/EFI when installing {target:?}");
```
travier: Nit:
```suggestion
            println!("Could not find /boot/efi/EFI when installing {target:?}");
```

src/bios.rs
```diff
@@ -142,10 +144,63 @@ impl Component for Bios {
...
     }
 
+    // Backup the current grub.cfg and install new static config
+    // - Backup "/boot/loader/grub.cfg" to "/boot/grub2/grub.cfg.bak"
+    // - Remove symlink "/boot/grub2/grub.cfg"
+    // - Install new static grub.cfg
+    // - Remove old "/boot/loader/grub.cfg"
+    fn migrate_static_grub_config(&self, sysroot_path: &str, destdir: &openat::Dir) -> Result<()> {
+        use anyhow::Context;
+        use openat_ext::OpenatDirExt;
```
travier: Hum, why place those here and not at the top of the file?

src/bios.rs
```diff
@@ -142,10 +144,63 @@ impl Component for Bios {
...
+        let grub_config_dir = destdir
+            .sub_dir("boot/grub2")
+            .context("Opening boot/grub2")?;
+
+        let grub_config = grub_config_path.join(grubconfigs::GRUBCONFIG);
+
+        // On BIOS, /boot/grub2/grub.cfg is symlink to /boot/loader/grub.cfg,
+        // backup it to /boot/grub2/grub.cfg.bak
+        if !grub_config.exists() {
+            anyhow::bail!("Not found '{}'", grub_config);
```
travier:
```suggestion
            anyhow::bail!("Could not find '{}'", grub_config);
```

src/bios.rs
```diff
@@ -142,10 +144,63 @@ impl Component for Bios {
...
+        } else if !grub_config.is_symlink() {
+            anyhow::bail!("'{}' is not a symlink", grub_config);
```
travier: Should this be an error? If this is not a symlink then we need to keep going as the migration might have been interrupted after we did symlink swap.

src/bios.rs
```diff
@@ -142,10 +144,63 @@ impl Component for Bios {
...
+    // - Remove old "/boot/loader/grub.cfg"
+    fn migrate_static_grub_config(&self, sysroot_path: &str, destdir: &openat::Dir) -> Result<()> {
+        use anyhow::Context;
+        use openat_ext::OpenatDirExt;
+
+        // sysroot_path is /, destdir is Dir of /
+        let grub_config_path = Utf8PathBuf::from(sysroot_path).join("boot/grub2");
+        let grub_config_dir = destdir
+            .sub_dir("boot/grub2")
+            .context("Opening boot/grub2")?;
```
travier: What the difference between those two paths here?

src/bios.rs
```diff
@@ -142,10 +144,63 @@ impl Component for Bios {
...
+        let backup_config = grub_config_path.join(grubconfigs::GRUBCONFIG_BACKUP);
+
+        // Backup the current GRUB config which is hopefully working right now
+        println!(
+            "Creating a backup of the current GRUB config '{}' in '{}'...",
+            current_config, backup_config
+        );
+        std::fs::copy(&current_config, &backup_config).context("Failed to backup GRUB config")?;
+
+        crate::grubconfigs::install(&destdir, None, true)?;
```
travier, line 192: Is this doing an atomic swap with the symlink?

src/efi.rs
```diff
@@ -246,11 +246,38 @@ impl Component for Efi {
...
+
+        // destdir is /boot/efi/EFI
+        let efidir = destdir
+            .sub_dir(&vendor)
+            .with_context(|| format!("Opening EFI/{}", vendor))?;
+        println!("Creating a backup of the current GRUB config on EFI");
+        efidir
+            .copy_file(grubconfigs::GRUBCONFIG, grubconfigs::GRUBCONFIG_BACKUP)
+            .context("Failed to backup GRUB config")?;
+        grubconfigs::install(&sysroot, Some(&vendor), true)?;
```
travier, line 270: Same question, we need to make sure it's doing an atomic swap

src/efi.rs
```diff
@@ -246,11 +246,38 @@ impl Component for Efi {
...
+            openat::Dir::open(sysroot_path).with_context(|| format!("Opening {sysroot_path}"))?;
+        let Some(vendor) = self.get_efi_vendor(&sysroot)? else {
+            anyhow::bail!("Failed to find efi vendor");
+        };
+
+        // destdir is /boot/efi/EFI
+        let efidir = destdir
+            .sub_dir(&vendor)
+            .with_context(|| format!("Opening EFI/{}", vendor))?;
+        println!("Creating a backup of the current GRUB config on EFI");
```
travier: Maybe we should not do this if the file already exists (from a previously interrupted migration), but this is an edge case so OK to not check it as well

travier (commented):
Thanks, mostly LGTM with some questions / things I don't understand / know

src/bios.rs
travier: https://docs.rs/openat-ext/latest/openat_ext/trait.OpenatDirExt.html#method.write_file_contents
it should

src/efi.rs
travier: https://docs.rs/openat-ext/latest/openat_ext/trait.OpenatDirExt.html#method.write_file_contents

src/bios.rs
jlebon: Right, yeah. The flow needs to be idempotent.

Two issues I see:
- We would overwrite the real backup when we rerun. We should check if the backup file already exists before overwriting. (Ahh OK, Timothee commented on that as well on the EFI side).
- We should use `remove_file_optional` instead of `remove_file` below.

Makefile
travier: Yes, the use case for that specific command is very specific to older RHCOS systems right now and should not be the default (or we would have to override it in the Atomic Desktops then).

It's probably safer to override the unit in RHCOS as needed.

src/bios.rs
HuijingHei: `grub_config_path` is `Utf8PathBuf` of `/boot/grub2`, that will be used to backup.
`grub_config_dir` is `Openat::Dir`, that will be used as `read_link` the real path of `/boot/grub2/grub.cfg` and do sync.
So we need both, WDYT?

src/bios.rs
HuijingHei: If `/boot/grub2/grub.cfg` is not symlink, maybe skip backup, then will overwrite the `grub.cfg` with static config, and can not remove the config as it is what we need.
What we want to remove is `/boot/grub2/../loader/grub.cfg`, but now it is `/boot/grub2/grub.cfg`. Need to check before `remove_file`, WDYT?

src/bios.rs
HuijingHei: Move to the top now, thank you for the review!

HuijingHei:
I did testing on 4.1 BIOS with the current change, after upgrade and reboot, check that:
- create backup
- remove the real config of symlink
- install the static grub configuration
```
[root@localhost core]# journalctl  -u bootloader-migrate.service | cat
Jun 19 05:39:00 localhost.localdomain systemd[1]: Starting Static GRUB config migration...
Jun 19 05:39:03 localhost.localdomain bootupctl[874]: ostree repo 'sysroot.bootloader' config option is not set yet
Jun 19 05:39:03 localhost.localdomain bootupctl[874]: Creating a backup of the current GRUB config '/boot/grub2/../loader/grub.cfg' in '/boot/grub2/grub.cfg.backup'...
...
Jun 19 05:39:03 localhost.localdomain bootupctl[874]: Installed: grub.cfg
Jun 19 05:39:03 localhost.localdomain bootupctl[874]: Installed: bootuuid.cfg
Jun 19 05:39:03 localhost.localdomain bootupctl[874]: Removing /boot/grub2/../loader/grub.cfg
Jun 19 05:39:03 localhost.localdomain bootupctl[874]: Static GRUB configuration has been adopted successfully.
Jun 19 05:39:03 localhost.localdomain bootupctl[874]: Adopted and updated: BIOS: grub2-tools-1:2.06-104.el9_6.x86_64
Jun 19 05:39:03 localhost.localdomain systemd[1]: Finished Static GRUB config migration.
```

In case that process is unfinished, need to rerun. 
Unset `sysroot.bootloader` and remove `/boot/bootupd-state.json`, restart the service again:
- check grub.cfg is not symlink
- backup file existing so skip the backup
- do not remove the config as it is not symlink
- install the static grub configuration

```
ostree config --repo=/sysroot/ostree/repo unset sysroot.bootloader
rm -v /boot/bootupd-state.json 

# systemctl restart bootloader-migrate.service

# journalctl  -u bootloader-migrate.service | cat
Jun 19 05:41:15 localhost.localdomain bootupctl[1219]: ostree repo 'sysroot.bootloader' config option is not set yet
Jun 19 05:41:15 localhost.localdomain bootupctl[1219]: '/boot/grub2/grub.cfg' is not a symlink
...
Jun 19 05:41:15 localhost.localdomain bootupctl[1219]: Installed: grub.cfg
Jun 19 05:41:15 localhost.localdomain bootupctl[1219]: Installed: bootuuid.cfg
Jun 19 05:41:15 localhost.localdomain bootupctl[1219]: Static GRUB configuration has been adopted successfully.
Jun 19 05:41:15 localhost.localdomain bootupctl[1219]: Adopted and updated: BIOS: grub2-tools-1:2.06-104.el9_6.x86_64
Jun 19 05:41:15 localhost.localdomain systemd[1]: Finished Static GRUB config migration.
```

src/bios.rs
```diff
@@ -142,10 +145,69 @@ impl Component for Bios {
         crate::component::query_adopt_state()
     }
 
+    // Backup the current grub.cfg and install new static config
+    // - Backup "/boot/loader/grub.cfg" to "/boot/grub2/grub.cfg.bak"
+    // - Remove symlink "/boot/grub2/grub.cfg"
+    // - Install new static grub.cfg
```
travier: I think this happens in one go here so maybe:

```suggestion
    // - Replace "/boot/grub2/grub.cfg" symlink with new static `grub.cfg`
```

src/bios.rs
```diff
@@ -142,10 +145,69 @@ impl Component for Bios {
...
+            std::fs::copy(&current_config, &backup_config)
+                .context("Failed to backup GRUB config")?;
+        }
+
+        crate::grubconfigs::install(&destdir, None, true)?;
+
+        // Remove the real config if it is symlink
+        if current_config != grub_config {
+            println!("Removing {}", current_config);
+            grub_config_dir.remove_file(current_config.as_std_path())?;
```
travier: This is a clean up and not worth stopping the migration for. We can ignore the error and log it instead.

travier (approved):
Two last nits but this LGTM. Thanks a lot

src/bios.rs
travier: https://docs.rs/openat-ext/latest/openat_ext/trait.OpenatDirExt.html#tymethod.remove_file_optional

src/bios.rs
travier: https://github.com/coreos/bootupd/pull/945#discussion_r2156982205

src/bios.rs
travier: I was thinking that maybe we wouldn't need to use both the openat and the path based access methods but I would have to try that to make sure it's actually possible. The code as is looks good so that could be later cleanup if I ever get to it.

src/bios.rs
HuijingHei: I am OK to change to `remove_file_optional`. We still need the check here, right? If `/boot/grub2/grub.cfg` is not symlink here, then it will be removed, and what we want to remove is `/boot/grub2/../loader/grub.cfg`.

travier: We should completely ignore the error here, thus never use `?` and ignore the error. It should not happen but just in case.

travier: The if condition looks good and is needed

PR #947, opened 2025-06-10
migrate: separate ostree command to functions

HuijingHei:
Add `fn get_ostree_bootloader()` to get `sysroot.bootloader` in ostree repo config.
Add `fn set_ostree_bootloader(bootloader: &str)` to set `sysroot.bootloader` value in ostree repo config.

Inspired by https://github.com/coreos/bootupd/pull/945#discussion_r2135764211

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -10,7 +10,7 @@ use crate::coreos;
 ))]
 use crate::efi;
 use crate::model::{ComponentStatus, ComponentUpdatable, ContentMetadata, SavedState, Status};
-use crate::util;
+use crate::{ostreeutil, util};
 use anyhow::{anyhow, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use clap::crate_version;
@@ -561,23 +561,8 @@ pub(crate) fn client_run_validate() -> Result<()> {
 #[context("Migrating to a static GRUB config")]
 pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
     // Did we already complete the migration?
-    let mut cmd = std::process::Command::new("ostree");
-    let result = cmd
-        .args([
-            "config",
-            "--repo=/sysroot/ostree/repo",
-            "get",
-            "sysroot.bootloader",
-        ])
-        .output()
-        .context("Querying ostree sysroot.bootloader")?;
-    if !result.status.success() {
-        // ostree will exit with a non zero return code if the key does not exists
-        println!("ostree repo 'sysroot.bootloader' config option is not set yet");
-    } else {
-        let res = String::from_utf8(result.stdout)
-            .with_context(|| "decoding as UTF-8 output of ostree command")?;
-        let bootloader = res.trim_end();
+    // We need to migrate if bootloader is not none (or not set)
+    if let Some(bootloader) = ostreeutil::get_ostree_bootloader()? {
         if bootloader == "none" {
             println!("Already using a static GRUB config");
             return Ok(());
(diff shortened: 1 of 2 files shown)


HuijingHei:
@cgwalters would you like to review when you get a chance? Thank you!

src/bootupd.rs
```diff
@@ -680,25 +633,103 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
 
     println!("Setting 'sysroot.bootloader' to 'none' in ostree repo config...");
+    set_ostree_bootloader("none")?;
+
+    println!("Static GRUB config migration completed successfully");
+    Ok(())
+}
+
+#[context("Get sysroot.bootloader in ostree repo config")]
+pub(crate) fn get_ostree_bootloader() -> Result<Option<String>> {
```
cgwalters: All good as is - this is correctly just moving code. But as I think about it it'd probably make sense at some point for us to grow an optional dependency on the ostree shared libraries.

src/bootupd.rs
```diff
@@ -680,25 +633,103 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
-        anyhow::bail!("Failed to set 'sysroot.bootloader' to 'none' in ostree repo config");
+        anyhow::bail!("Failed to set 'sysroot.bootloader' to '{bootloader}' in ostree repo config");
     }
+    Ok(())
+}
+
+/// Writes a stripped GRUB config to `stripped_config_name`, removing lines between
+/// `### BEGIN /etc/grub.d/15_ostree ###` and `### END /etc/grub.d/15_ostree ###`.
+fn strip_grub_config_file(
+    current_config_path: &Path,
```
cgwalters: I think it'd be cleaner if this was:
```
     current_config_path: impl Read
```
or so. That way everything in the function would consistently operate on `dirfd` which lends itself more easily to unit testing.

(The callers would take care of opening the file, this line would move there)

cgwalters (approved):
LGTM as is, one note for the future and one optional thing (feel free to merge as is or adjust it now)

src/bootupd.rs
```diff
@@ -586,7 +571,7 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
             "ostree repo 'sysroot.bootloader' config option is currently set to: '{}'",
             bootloader
         );
-    }
+    };
```
jlebon: Hmm, that semicolon shouldn't be necessary, but I'm surprised clippy doesn't complain about it.

src/bootupd.rs
```diff
@@ -680,25 +633,103 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
+            "config",
+            "--repo=/sysroot/ostree/repo",
+            "get",
+            "sysroot.bootloader",
+        ])
+        .output()
+        .context("Querying ostree sysroot.bootloader")?;
+    if !result.status.success() {
+        // ostree will exit with a non zero return code if the key does not exists
+        println!("ostree repo 'sysroot.bootloader' config option is not set yet");
```
jlebon: Should this print statement move to the caller? Feels odd to have a helper print things, but also it then makes it consistent with the other branches in that call on the callers side.

jlebon (approved):
Some minor comments, but LGTM as is too.

src/bootupd.rs
HuijingHei: SGTM, will move it to caller.

src/bootupd.rs
HuijingHei: Agree, it would be helpful for long term. Do you mean ostree shared libraries like "bootc-blockdev" that others can use?
Now just move the 2 ostree command to ostreeutil.

PR #948, opened 2025-06-11
Add fsfreeze_thaw_cycle() + small cleanups

champtar:
I went with a simple implementation of `fsfreeze_thaw_cycle()` compared to the ostree one as we are already protected from SIGTERM, and in case FIFREEZE is slow / stuck, there is no indication calling FITHAW will help, so l prefer to keep it simple.
I can move [grubconfigs: cleanup install()](https://github.com/coreos/bootupd/commit/432d56e48bfc3c380d88804a6127007a45e90e0f) to another PR but it need to be merged first

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -9,6 +9,7 @@ use crate::coreos;
     target_arch = "riscv64"
 ))]
 use crate::efi;
+use crate::freezethaw::fsfreeze_thaw_cycle;
 use crate::model::{ComponentStatus, ComponentUpdatable, ContentMetadata, SavedState, Status};
 use crate::{ostreeutil, util};
 use anyhow::{anyhow, Context, Result};
@@ -17,7 +18,6 @@ use clap::crate_version;
 use fn_error_context::context;
 use libc::mode_t;
 use libc::{S_IRGRP, S_IROTH, S_IRUSR, S_IWUSR};
-use openat_ext::OpenatDirExt;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
@@ -625,21 +625,14 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
 
             strip_grub_config_file(content, &dirfd, stripped_config)?;
 
-            // Sync changes to the filesystem (ignore failures)
-            let _ = dirfd.syncfs();
-
-            // Atomically exchange the configs
+            // Atomically replace the symlink
             dirfd
-                .local_exchange(stripped_config, "grub.cfg")
-                .context("Failed to exchange symlink with current GRUB config")?;
+                .local_rename(stripped_config, "grub.cfg")
+                .context("Failed to replace symlink with current GRUB config")?;
 
-            // Sync changes to the filesystem (ignore failures)
-            let _ = dirfd.syncfs();
+            fsfreeze_thaw_cycle(dirfd.open_file(".")?)?;
 
             println!("GRUB config symlink successfully replaced with the current config");
-
-            // Remove the now unused symlink (optional cleanup, ignore any failures)
-            _ = dirfd.remove_file(stripped_config);
         }
     };
 
@@ -690,8 +683,10 @@ fn strip_grub_config_file(
     }
 
     writer
-        .flush()
-        .context("Failed to write stripped GRUB config")?;
+        .into_inner()
+        .context("Failed to flush stripped GRUB config")?
+        .sync_data()
+        .context("Failed to sync stripped GRUB config")?;
 
     Ok(())
 }
--- a/src/freezethaw.rs
+++ b/src/freezethaw.rs
@@ -1,0 +1,50 @@
+use rustix::fd::AsFd;
+use rustix::ffi as c;
+use rustix::io::Errno;
+use rustix::ioctl::opcode;
+use rustix::{io, ioctl};
+
+use crate::util::SignalTerminationGuard;
+
+fn ioctl_fifreeze<Fd: AsFd>(fd: Fd) -> io::Result<()> {
+    // SAFETY: `FIFREEZE` is a no-argument opcode.
+    // `FIFREEZE` is defined as `_IOWR('X', 119, int)`.
+    unsafe {
+        let ctl = ioctl::NoArg::<{ opcode::read_write::<c::c_int>(b'X', 119) }>::new();
+        ioctl::ioctl(fd, ctl)
+    }
+}
+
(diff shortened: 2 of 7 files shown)


src/bootupd.rs
```diff
@@ -625,21 +625,14 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
...
+            // Atomically replace the symlink
             dirfd
-                .local_exchange(stripped_config, "grub.cfg")
-                .context("Failed to exchange symlink with current GRUB config")?;
+                .local_rename(stripped_config, "grub.cfg")
+                .context("Failed to replace symlink with current GRUB config")?;
 
-            // Sync changes to the filesystem (ignore failures)
-            let _ = dirfd.syncfs();
+            fsfreeze_thaw_cycle(dirfd.open_file(".")?)?;
```
HuijingHei, line 633: A small question here, why not using  `fsfreeze_thaw_cycle(dirfd.as_raw_fd())?`

src/freezethaw.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/// syncfs() doesn't flush the journal on XFS,
+/// and since GRUB2 can't read the XFS journal, the system
+/// could fail to boot.
+///
+/// http://marc.info/?l=linux-fsdevel&m=149520244919284&w=2
+/// https://github.com/ostreedev/ostree/pull/1049
+///
+/// When calling this function you should be protected by SignalTerminationGuard
```
cgwalters: We could accept such a guard as an argument to this function, or honestly pretty safely just create another one?

src/freezethaw.rs
```diff
@@ -0,0 +1,58 @@
...
+fn ioctl_fithaw<Fd: AsFd>(fd: Fd) -> io::Result<()> {
+    // SAFETY: `FITHAW` is a no-argument opcode.
+    unsafe {
+        let ctl = ioctl::NoArg::<{ FITHAW }>::new();
+        ioctl::ioctl(fd, ctl)
+    }
+}
+
+/// syncfs() doesn't flush the journal on XFS,
+/// and since GRUB2 can't read the XFS journal, the system
```
cgwalters, line 28: Do we really know it's only xfs? It seems safer to do this unconditionally.

src/freezethaw.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[cfg(test)]
+mod tests {
+    use rustix::fs::{Mode, OFlags};
+    use rustix::io;
+
+    #[test]
+    #[ignore]
+    fn test_fsfreeze_thaw_cycle() -> io::Result<()> {
+        let dir = rustix::fs::open("/boot/", OFlags::RDONLY, Mode::RUSR)?;
```
cgwalters: We shouldn't affect global state from unit tests. In the general case we should expect unit tests to run without privileges so this should always fail...

I'd just drop the unit test here.

(What I do want to do in general is have a standard pattern to add these types of things as integration tests, but it's a lot more complex)

champtar: it has `#[ignore]`, so you can quickly play with the function, but I'm ok dropping it
For the record I confirmed with strace that it sent the right ioctl, and we can use subfolders, we don't need to use the mount point

src/bootupd.rs
champtar: We need `AsFD`, not `AsRawFD`
Also reading https://github.com/coreos/openat-ext/blob/34502fe72315f127ff6e72b0649551017ff897ec/src/lib.rs#L341, we can't use an fd open with O_PATH for syncfs(), so reopening without O_PATH.

With cap-std-ext we would use [reopen_as_ownedfd](https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.reopen_as_ownedfd)

src/freezethaw.rs
champtar: Will do, ignoring ENOTSUPP errors

src/freezethaw.rs
champtar: > (What I do want to do in general is have a standard pattern to add these types of things as integration tests, but it's a lot more complex)

Can we just gate the integration tests behind something like `#[cfg(integration_test)]` ?

champtar: Dropped the test for now, but fsfreeze_thaw_cycle() is still called via other tests
I now just ignore `EOPNOTSUPP` and `EPERM`

src/freezethaw.rs
champtar: Done, ignoring `EOPNOTSUPP` and `EPERM`

src/freezethaw.rs
champtar: Done, should we also guard against SIGINT ?

cgwalters: I think today since we always run via systemd, we shouldn't get a random stray `SIGINT`, so I'm not too worried about it.

cgwalters:
That's weird that it only fails on ppc64le. Hmmm... I think this may be us running into https://github.com/bytecodealliance/rustix/blob/0a6cb07f8270aab17e534853fb25de1f7a0c9b33/src/ioctl/mod.rs#L293 - with glibc it's a long?

But it's not clear to me why we're not running into a failure on all arches. Hmm it might be because of https://github.com/bytecodealliance/rustix/blob/0a6cb07f8270aab17e534853fb25de1f7a0c9b33/Cargo.toml#L32 where linux-raw-sys is enabled only on some architectures.

We should be able to do this one this way, as in the tests: https://github.com/bytecodealliance/rustix/blob/0a6cb07f8270aab17e534853fb25de1f7a0c9b33/src/ioctl/mod.rs#L373

champtar:
> That's weird that it only fails on ppc64le. Hmmm... I think this may be us running into https://github.com/bytecodealliance/rustix/blob/0a6cb07f8270aab17e534853fb25de1f7a0c9b33/src/ioctl/mod.rs#L293 - with glibc it's a long?
> 
> But it's not clear to me why we're not running into a failure on all arches. Hmm it might be because of https://github.com/bytecodealliance/rustix/blob/0a6cb07f8270aab17e534853fb25de1f7a0c9b33/Cargo.toml#L32 where linux-raw-sys is enabled only on some architectures.
> 
> We should be able to do this one this way, as in the tests: https://github.com/bytecodealliance/rustix/blob/0a6cb07f8270aab17e534853fb25de1f7a0c9b33/src/ioctl/mod.rs#L373

Thanks for digging into this, added 2 small tests to confirm the issue

champtar:
@cgwalters it builds now

cgwalters (approved):

PR #954, opened 2025-06-18
Sync repo templates ⚙

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/repo-templates/actions/workflows/sync.yml) ([source](https://github.com/coreos/repo-templates/blob/main/.github/workflows/sync.yml)).

Sync with coreos/repo-templates@ecc09fcda5542136ef418ea1932bea328af6eb1a.


HuijingHei:
> we edited the workflow here but it should have been edited in the template repo, ore removed from there.

I am OK to enable AI code review. But we need the `dnf install -y grub2-tools-minimal` to do unit testing.

PR #960, opened 2025-06-24
install: attempt to use an already mounted ESP at the target

HuijingHei:
Let's attempt to use an already mounted ESP at the target
dest_root if one is already mounted there in a known ESP location,
if failed, will fallback to mount ESP partition of the device.

See https://github.com/coreos/coreos-assembler/pull/4161

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -22,7 +22,7 @@ parallel build: {
         """)
       }
       shwrap("""
-        make install install-grub-static install-systemd-unit DESTDIR=\$(pwd)/insttree/
+        make install-all DESTDIR=\$(pwd)/insttree/
         tar -c -C insttree/ -zvf insttree.tar.gz .
       """)
       stash includes: 'insttree.tar.gz', name: 'build'
@@ -55,6 +55,7 @@ cosaPod(runAsUser: 0, memory: "4608Mi", cpu: "4") {
       rmdir insttree
       cosa fetch
       cosa build
+      cosa osbuild metal4k
     """)
   }
   // The e2e-adopt test will use the ostree commit we just generated above


src/efi.rs
```diff
@@ -345,14 +345,19 @@ impl Component for Efi {
...
+        } else {
+            // Using `blockdev` to find the partition instead of partlabel because
+            // we know the target install toplevel device already.
+            let esp_device = blockdev::get_esp_partition(device)?
+                .ok_or_else(|| anyhow::anyhow!("Failed to find ESP device"))?;
+            self.mount_esp_device(Path::new(dest_root), Path::new(&esp_device))?
+        };
 
-        let destd = &openat::Dir::open(destpath)
+        let destd = &openat::Dir::open(&destpath)
```
dustymabe: I'm not 100% sure this is going to solve the problem in all cases. Do I understand correctly that `device` passed into this `install()` function will be None (i.e. `device.is_empty()` is true) if we aren't doing BIOS?

If that's the case then won't we run into this same problem IF the ESP isn't already mounted on platforms where BIOS isn't requested? i.e. if the ESP isn't already mounted and BIOS wasn't requested then we'll still fail. In the commit message for https://github.com/coreos/bootupd/commit/f6ef3d1077a7eccf89a751cc5c805613e62857e9 it mentions

>  During `bootc install to-disk` the esp device is not mounted

So I guess one way to test that theory would be to run a `bootc install to-disk` on say aarch64.

I kind of wish we coul just assume the ESP is already mounted in this function.

src/efi.rs
```diff
@@ -345,14 +345,19 @@ impl Component for Efi {
...
         let ft = crate::filetree::FileTree::new_from_dir(&src_root.sub_dir(&srcdir_name)?)?;
 
-        // Using `blockdev` to find the partition instead of partlabel because
-        // we know the target install toplevel device already.
-        let esp_device = blockdev::get_esp_partition(device)?
-            .ok_or_else(|| anyhow::anyhow!("Failed to find ESP device"))?;
-
-        let destpath = &self.ensure_mounted_esp(Path::new(dest_root), Path::new(&esp_device))?;
+        // When building a live image without a device (e.g., metal4k),
+        // a mounted directory will be used to continue the process.
```
dustymabe: Since this problem is going to occur any time we aren't using BIOS (i.e. device isn't passed in) I don't suggest we mention metal4k directly. 

```suggestion
        // Let's attempt to use an already mounted ESP at the target
        // dest_root if one is already mounted there in a known ESP location.
```

src/efi.rs
HuijingHei: > Do I understand correctly that `device` passed into this `install()` function will be None (i.e. `device.is_empty()` is true) if we aren't doing BIOS?

Yes, we need the device to install BIOS, if without device on x64, will skip BIOS, see https://github.com/coreos/bootupd/blob/main/src/bootupd.rs#L88

> If that's the case then won't we run into this same problem IF the ESP isn't already mounted on platforms where BIOS isn't requested? i.e. if the ESP isn't already mounted and BIOS wasn't requested then we'll still fail. In the commit message for [f6ef3d1](https://github.com/coreos/bootupd/commit/f6ef3d1077a7eccf89a751cc5c805613e62857e9) it mentions
> 
> > During `bootc install to-disk` the esp device is not mounted
> 
> So I guess one way to test that theory would be to run a `bootc install to-disk` on say aarch64.

On aarch64 is a little different as we do not include `BIOS` module, and `bootc install to-disk` calls ` bootupctl backend install --write-uuid --device /dev/loop0 /run/bootc/mounts/rootfs` with device (see run [logs](https://github.com/coreos/bootupd/actions/runs/15854361030/job/44695793769?pr=960))

> I kind of wish we coul just assume the ESP is already mounted in this function.

Yes, as we do not have device and can not find esp and can not mount.

HuijingHei: > if the ESP isn't already mounted and BIOS wasn't requested then we'll still fail.

Agree that will be a problem. We have the option `--component` to specify the target component, see https://github.com/coreos/bootupd/blob/main/src/cli/bootupd.rs#L66, maybe we should pass the device and specify `--component EFI` to skip BIOS when building metal4k image (for x64), but I think the workaround also works for that change.

HuijingHei: Include `cosa osbuild metal4k` in CI, and result is PASSED, IMU the workaround will fix the problem.

dustymabe (approved):
LGTM

cgwalters (approved):

Issue #952, opened 2025-06-18
Ensure bootupd-generated GRUB configs have secure permissions and ownership


jan-cerny:
Security compliance frameworks, such as CIS or PCI-DSS, mandate that `/boot/grub2/grub.cfg` and `/boot/grub2/user.cfg` must be owned by `root:root` with file permissions set to `0600` to prevent unauthorized modification.

Currently, bootupd appears to rely on system defaults (e.g., umask) when creating this file, which can result in insufficient permissions (like 0644). This causes failing rules when performing compliance scans of RHEL Image Mode systems.

We propose that bootupd be enhanced to enforce secure-by-default settings for the bootloader configurations it creates. Specifically, when bootupd generates `/boot/grub2/grub.cfg` and `/boot/grub2/user.cfg` , it should explicitly set the file owner to root (UID 0), set the file group to root (GID 0) and et the file permissions to 0600.

Alternatively, bootupd could provide a way how to configure the owner and permissions of the GRUB files at the bootable container image build time, ie. users would drop in a config file for bootupd in their Containerfiles.

travier:
Fixed by https://github.com/coreos/bootupd/pull/961 for new installations. This will not fix it for existing systems, but this is a false positive so users that want to silence that warning can do a one time `chmod 600 ...` on their systems to fix that.

PR #961, opened 2025-06-25
grubconfig: set `/boot/grub2/grub.cfg` file mode to `0600`

HuijingHei:
Copy Colin's comment:
```
One overall issue on this is because we don't have a mechanism to
update the static configs, existing systems will stay as is.
```

See https://github.com/coreos/bootupd/issues/952 &
https://redhat-internal.slack.com/archives/C01BSEK9PM1/p1750152540290679

### DIFF
--- a/src/grubconfigs.rs
+++ b/src/grubconfigs.rs
@@ -17,6 +17,9 @@ const DROPINDIR: &str = "configs.d";
 const GRUBENV: &str = "grubenv";
 pub(crate) const GRUBCONFIG: &str = "grub.cfg";
 pub(crate) const GRUBCONFIG_BACKUP: &str = "grub.cfg.backup";
+// File mode for /boot/grub2/grub.config
+// https://github.com/coreos/bootupd/issues/952
+const GRUBCONFIG_FILE_MODE: u32 = 0o600;
 
 /// Install the static GRUB config files.
 #[context("Installing static GRUB configs")]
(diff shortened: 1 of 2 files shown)


src/grubconfigs.rs
```diff
@@ -103,6 +106,7 @@ pub(crate) fn install(
             efidir
                 .copy_file(&Path::new(CONFIGDIR).join("grub-static-efi.cfg"), target)
                 .context("Copying static EFI")?;
+            efidir.set_mode(target, GRUBCONFIG_FILE_MODE)?;
```
cgwalters: Side note: in the future we could use https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.atomic_write_with_perms to do this in one step.

cgwalters (approved):
One overall issue on this is because we don't have a mechanism to update the static configs, existing systems will stay as is.

src/grubconfigs.rs
HuijingHei: Thank you for the comment, I see your PR in https://github.com/coreos/bootupd/pull/674 is using `cap-std-ext`, maybe we can continue that.

cgwalters:
Hm, not sure offhand why the test is failing. Possibly we're not running the updated code we think we are, or the code is wrong.

PR #963, opened 2025-06-30
Use published bootc-internal-{utils,blockdev} crates

cgwalters:
For now we'll continue to publish bootupd as a crate,
but I think we should eventually stop doing this.
    

Closes: https://github.com/coreos/bootupd/issues/964

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,12 +155,13 @@ dependencies = [
 ]
 
 [[package]]
-name = "bootc-blockdev"
+name = "bootc-internal-blockdev"
 version = "0.0.0"
-source = "git+https://github.com/containers/bootc?rev=v1.3.0#b06d75fed73f26a974423a9db41e19413987fca3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


HuijingHei:
Thank you @cgwalters for the workaround, not sure if my understanding is correct, in this case, the dependencies `bootc-blockdev` and `bootc-utils` do not have a specified version here, so the publish will fail.

```
$ git clone https://github.com/cgwalters/bootupd.git -b sync-bootc-code
$ cargo publish --dry-run
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `bootc-blockdev` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

cgwalters:
Ah right, thanks for testing this out. Yeah I think we're hitting a really longstanding Rust toolchain limitation that requires publishing these sub-crates when what we really want is for them to just be automatically vendored.

I think this PR would need to be reworked to instead rework the sub-crates as modules (very unfortunate ergonomic hit).

HuijingHei (approved):
LGTM

PR #967, opened 2025-07-10
WIP: Add TMT gating test and workflow for bootupd

aaradhak:
Add TMT testing infrastructure including core smoke test with GitHub workflow integration.


tests/tmt/plans/main.fmf
```diff
@@ -0,0 +1,32 @@
+# This prepare is used to control when bootupd is installed using
+# the distribution package or when it is built from source in the test environment
+prepare:
+  - name: Set BOOTUPD_BIN_DIR when built from source
+    when: use_built_from_src is defined and use_built_from_src == true
```
cgwalters, line 5: What defines `use_built_from_src`?

cgwalters (commented):
I'm not opposed to this but it's also important to acknowledge that this is testing *almost* nothing.

As far as I understand without additional work, it will not be testing the source changes to the git repository here, only installing the existing package, which seems misleading at best.

I think what actually is solving that problem is packit which acts as a glue later on top of testing-farm/tmt to build rpms in a copr and then inject them by default.

But that gets to the *next* problem that doing useful bootupd testing requires being in a bootc/coreos system so we'd need to use e.g. https://tmt.readthedocs.io/en/1.42/plugins/provision.html#bootc

aaradhak:
> https://tmt.readthedocs.io/en/1.42/plugins/provision.html#bootc

Thank you for this information about bootc. Right now we are in the process of creating a gating test infrastructure for all the RPMs under CoreOS. It is a work in progress where I am trying to add a basic test in bootupd. 

Now I am planning to add a test from the CI to see if that works - https://github.com/coreos/bootupd/blob/main/.github/workflows/ci.yml

tests/tmt/plans/main.fmf
aaradhak: The definition is missing. I ll add it in `tmt-test.yml`. It should be something like this

```
      use_built_from_src:
        description: 'Built bootupd from source instead of install distro package'
        required: false
        default: 'true'
```

gating.yaml
```diff
@@ -0,0 +1,7 @@
+--- !Policy
```
cgwalters: I don't believe this does anything in upstream tests and should probably only live in the dist-git repo.

cgwalters (commented):
Bigger picture @lsm5 from the container runtimes team is also working on tmt related things (ref https://github.com/containers/skopeo/pull/2640 ) and I think it might make sense for you (or others) working on TMT related things for coreos to at least have a quick sync.

aaradhak:
> Bigger picture @lsm5 from the container runtimes team is also working on tmt related things (ref [containers/skopeo#2640](https://github.com/containers/skopeo/pull/2640) ) and I think it might make sense for you (or others) working on TMT related things for coreos to at least have a quick sync.

Sure will do 

@mike-nguyen @HuijingHei @tlbueno : I would like to tag everyone to inform about this

mike-nguyen:
> https://tmt.readthedocs.io/en/1.42/plugins/provision.html#bootc
Neat! This will be super useful for testing some components.  

As Ash mentioned, the goal is to to have the framework in place and share it with the downstream gating. We want to make it easy to add tests so we can increase test coverage.

lsm5:
Hello, I see rust-bootupd is shipped on Fedora and CentOS Stream, so I'd highly recommend using a Packit workflow to manage Fedora releases as well as trigger TMT tests (on multiple Fedora and CentOS Stream environments). The same set of TMT tests can be run upstream as well as on Fedora and CentOS Stream (dist-git PRs, bodhi updates) with a single source of truth. We're already doing this on podman and many others in the containers org. And I see `bootc` is doing it too.

Btw, CentOS Stream official downstream updates can only be triggered via packit CLI for now because of access restrictions on lookaside cache.

cgwalters:
> We have a horrendous hack for this in bootc (build a container image inside the target environment in each test locally and then bootc install to-existing-root) that basically really wants to be more first-class in tmt.

I filed https://github.com/bootc-dev/bootc/issues/1473 on this and it's very relevant here too

PR #975, opened 2025-07-30
efi: improve `get_efi_component_from_usr` to return `EFIComponent`  struct

HuijingHei:
Inspired by Colin's pointer:
https://github.com/coreos/bootupd/pull/938#discussion_r2231409660
```
return EFIComponent struct:

struct EFIComponent {
  name: String,
  version: String,
  path: Utf8PathBuf
}

### CODE REVIEW

**Comment by HuijingHei** [comment]
Ready for reviewing now, thank you!

File: src/efi.rs [hunk 1]
```diff
@@ -436,27 +436,24 @@ impl Component for Efi {
...
+        let meta = if efilib_path.exists() {
             let mut packages = Vec::new();
-            for efi_dir in get_efi_path_from_usr(&sysroot_path)? {
-                let efi_dir = efi_dir.context("Invalid EFI directory path")?;
-
-                let sysroot_dir =
-                    Dir::open_ambient_dir(sysroot_path, cap_std::ambient_authority())?;
+            let sysroot_dir = Dir::open_ambient_dir(sysroot_path, cap_std::ambient_authority())?;
+            let efi_components_iter = get_efi_component_from_usr(&sysroot_path, EFILIB)?;
+            let mut efi_components: Vec<EFIComponent> = efi_components_iter.flatten().collect();
```
travier: Any specific reason to not do the flatten/collect in the get_efi_component_from_usr function directly?

HuijingHei: Updated, move the collect and sort into `get_efi_component_from_usr` function. Thank you for the reviewing!

PR #977, opened 2025-08-04
package: move to compare `version` instead of `timestamp`

HuijingHei:
Use `rpm::Evr` crate to parse and compare `version` when checking for update.

Inspired by Timothee's pointer:
https://github.com/coreos/bootupd/pull/938#issuecomment-3118960479
Fixes: https://github.com/coreos/bootupd/issues/933

Additional info:
Use rpm `0.16` instead of `0.17` as 1.84 does not support
feature `edition2024` that is requried by 0.17:
```
error: failed to parse manifest at `/root/.cargo/registry/src/
index.crates.io-6f17d22bba15001f/rpm-0.17.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`,
but that feature is not stabilized in this version of Cargo
(1.84.1 (66221abde 2024-11-19)).
```

### DIFF
--- a/src/packagesystem.rs
+++ b/src/packagesystem.rs
@@ -1,9 +1,11 @@
+use std::cmp::Ordering;
 use std::collections::{BTreeMap, BTreeSet};
 use std::io::Write;
 use std::path::Path;
 
 use anyhow::{bail, Context, Result};
 use chrono::prelude::*;
+use rpm::Evr;
 
 use crate::model::*;
 use crate::ostreeutil;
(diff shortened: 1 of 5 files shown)


HuijingHei:
Fix CI error and add related unit test if there is new package.

mike-nguyen:
Is there a reason why the `rpm_evr_compare` function from the `rpm` crate can't be used  (https://docs.rs/rpm/latest/rpm/fn.rpm_evr_compare.html)?

HuijingHei:
> Is there a reason why the `rpm_evr_compare` function from the `rpm` crate can't be used (https://docs.rs/rpm/latest/rpm/fn.rpm_evr_compare.html)?

I think the problem is we need to ignore epoch in our comparison, e.g. `grub2-efi-x64-1:2.12-28.fc42.x86_64` should be the same as `grub2-2.12-28.fc42`, but actually no (as the first one has `epoch` 1, so it is greater than second without `epoch`)

travier:
@lsandov1 we should probably change the installation path to include the epoch as well as the version and release.

HuijingHei:
> Would be good to file an issue in https://github.com/rpm-rs/rpm/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

SGTM, create issue https://github.com/rpm-rs/rpm/issues/275

HuijingHei:
> @lsandov1 we should probably change the installation path to include the epoch as well as the version and release.

WDYT about `%global grub_version_dir %{grub_root_dir}/%{epoch}:%{version}-%{release}` ?

src/packagesystem.rs
```diff
@@ -67,12 +69,182 @@ where
...
-fn test_parse_rpmout() {
-    let testdata = "grub2-efi-x64-1:2.06-95.fc38.x86_64,1681321788 grub2-efi-x64-1:2.06-95.fc38.x86_64,1681321788 shim-x64-15.6-2.x86_64,1657222566 shim-x64-15.6-2.x86_64,1657222566 shim-x64-15.6-2.x86_64,1657222566";
-    let parsed = rpm_parse_metadata(testdata.as_bytes()).unwrap();
-    assert_eq!(
-        parsed.version,
-        "grub2-efi-x64-1:2.06-95.fc38.x86_64,shim-x64-15.6-2.x86_64"
-    );
+fn parse_evr_map(input: &str) -> BTreeMap<String, Evr> {
+    input
+        .split(',')
```
cgwalters, line 74: It does seem like we can reliably split on this since AFAICS `,` is illegal in all of RPM, the uapi versions, dpkg, arch etc.

That said if we are moving to make the versions actually parsed, it would probably make sense to also switch to a proper array instead of a single string.

For compatibility though we'd need to do both for a while at least, i.e. have both `version` (string) and `versions` (array) in the model/JSON.

src/packagesystem.rs
```diff
@@ -67,12 +69,182 @@ where
...
+        .map(|pkg| {
+            // assume it is like "grub2-1:2.12-28.fc42,shim-15.8-3"
+            if !pkg.ends_with(std::env::consts::ARCH) {
+                let (name, version_release) = pkg.split_once('-').unwrap_or((pkg, ""));
+                let evr = Evr::parse(version_release);
+                return (name.to_string(), evr);
+            }
+
+            // assume it is like "grub2-efi-x64-1:2.12-28.fc42.x86_64,shim-x64-15.8-3.x86_64"
+            let nevra = rpm::Nevra::parse(pkg);
```
cgwalters, line 84: Per my earlier comment I think this is OK as is but I'd like to not add new hardcoded RPM dependencies. We can just make this a build-time configuration option in the future.

src/packagesystem.rs
```diff
@@ -67,12 +69,182 @@ where
...
+    }
+    let diffs = compare_package_versions_impl(a, b);
+
+    let mut has_less = false;
+    let mut has_greater = false;
+
+    for (_, ord) in &diffs {
+        match ord {
+            Ordering::Less => {
+                has_less = true;
```
cgwalters, line 142: This can just be `return Ordering::less` right? No need for the bool.

src/packagesystem.rs
```diff
@@ -67,12 +69,182 @@ where
...
+        }
+    }
+    result
+}
+
+// Compare package versions:
+// If any package is Ordering::Less, return Ordering::Less, means upgradable,
+// Else if any package is Ordering::Greater, return Ordering::Greater,
+// Else (all equal), return Ordering::Equal.
+pub(crate) fn compare_package_versions(a: &str, b: &str) -> Ordering {
```
cgwalters, line 129: I think this is OK but it'd be more idiomatic to create a `struct Package` that has `impl Ord`.

And also I think per above we should split on the `,` outside this function, i.e. we're comparing two slices `[Package]` - and if everything compares equal then the longer one wins.

HuijingHei: SGTM, thank you for the pointer, update in https://github.com/coreos/bootupd/pull/983

src/packagesystem.rs
HuijingHei: Nice catch, update in https://github.com/coreos/bootupd/pull/983

src/packagesystem.rs
HuijingHei: > For compatibility though we'd need to do both for a while at least, i.e. have both `version` (string) and `versions` (array) in the model/JSON.

What do you think about like this? Then we do not need to parse the str.
```
{
  "version": "pkg1-1.0,pkg2-2.0",   // old single string
  "versions": [                       // new proper array
    {"name": "pkg1", "evr": "1.0"},
    {"name": "pkg2", "evr": "2.0"}
  ]
}
```

cgwalters: How about replacing `evr` with `rpm-evr` to make explicit what it is? On other operating systems the version might be different.

BTW recently in https://github.com/bootc-dev/bootc/pull/1512/commits/17c03f7ab6b75fb1f71f61f1f3a0200ef01ca26e I added the uapi_versions crate to bootc. It would make sense to me to use it here too *optionally* for systems not using rpm.

But anyways let's start with introducing the parallel `versions` struct with `rpm-evr` or so.

src/packagesystem.rs
HuijingHei: Do you mean that we only import rpm if using `cargo build --features rpm`, else using `cargo build` without rpm, but we also use `cmp()` to compare `rpm::Evr` (see https://docs.rs/rpm/latest/rpm/struct.Evr.html#method.cmp), so we still need rpm, is this right?

cgwalters: I think it's OK to have an rpm feature and use it.

But we could have a generic version that uses https://docs.rs/uapi-version/latest/uapi_version/ to compare versions and require that package systems map to a compatible version scheme (if it's not compatible).

PR #978, opened 2025-08-08
[WIP] feat: enable systemd-boot support

p5:
This is nowhere near ready, and I highly doubt any of this code will make it's way into bootupd without major refactoring...  I'm just trying to make something that works and start the feedback cycle early.

Info logs changed to warns as it was the easiest way to make them show up during `bootc install`.

TODO:
- [ ] Major refactoring
- [ ] Correctly detect (or accept as input) the product name
- [ ] Correctly detect whether to install with sd-boot or grub
- [ ] Boot into a composefs image (since ostree currently lacks support and boots to emergency mode)
- [ ] Everything else

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -96,7 +96,7 @@ pub(crate) fn install(
         let meta = component
             .install(&source_root, dest_root, device, update_firmware)
             .with_context(|| format!("installing component {}", component.name()))?;
-        log::info!("Installed {} {}", component.name(), meta.meta.version);
+        log::warn!("Installed {} {}", component.name(), meta.meta.version);
         state.installed.insert(component.name().into(), meta);
         // Yes this is a hack...the Component thing just turns out to be too generic.
         if let Some(vendor) = component.get_efi_vendor(&source_root)? {
@@ -303,7 +303,7 @@ pub(crate) fn adopt_and_update(
         return Ok(Some(update));
     } else {
         // Nothing adopted, skip
-        log::info!("Component '{}' skipped adoption", component.name());
+        log::warn!("Component '{}' skipped adoption", component.name());
         return Ok(None);
     }
 }
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -20,7 +20,6 @@ use openat_ext::OpenatDirExt;
 use os_release::OsRelease;
 use rustix::fd::BorrowedFd;
 use walkdir::WalkDir;
-use widestring::U16CString;
 
 use crate::bootupd::RootContext;
 use crate::freezethaw::fsfreeze_thaw_cycle;
@@ -47,9 +46,9 @@ pub(crate) const SHIM: &str = "shimx64.efi";
 #[cfg(target_arch = "riscv64")]
 pub(crate) const SHIM: &str = "shimriscv64.efi";
 
-/// Systemd boot loader info EFI variable names
-const LOADER_INFO_VAR_STR: &str = "LoaderInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
-const STUB_INFO_VAR_STR: &str = "StubInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
+// /// Systemd boot loader info EFI variable names
+// const LOADER_INFO_VAR_STR: &str = "LoaderInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
+// const STUB_INFO_VAR_STR: &str = "StubInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
 
 /// Return `true` if the system is booted via EFI
 pub(crate) fn is_efi_booted() -> Result<bool> {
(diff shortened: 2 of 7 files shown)


src/efi.rs
```diff
@@ -153,6 +151,35 @@ impl Efi {
         clear_efi_target(&product_name)?;
         create_efi_boot_entry(device, espdir, vendordir, &product_name)
     }
+
+    /// Find a kernel and optional initramfs/uki file in the update directory for systemd-boot
+    fn find_kernel_and_initrd(dir: &openat::Dir) -> Result<(String, Option<String>)> {
```
cgwalters: bootupd should not have anything to do with kernels. That's the job of bootc (or dnf/apt/etc).

src/efi.rs
```diff
@@ -365,23 +329,94 @@ impl Component for Efi {
...
+            log::warn!("Using systemd-boot, creating entries dir at {}", entries_dir.display());
+            std::fs::create_dir_all(&entries_dir)
+                .with_context(|| format!("creating entries dir {}", entries_dir.display()))?;
+            log::warn!("Installing systemd-boot entry in {}/bootupd.conf", entries_dir.display());
+            // Write loader.conf if it doesn't exist
+            let loader_conf_path = loader_dir.join("loader.conf");
+            if !loader_conf_path.exists() {
+                let mut loader_conf = std::fs::File::create(&loader_conf_path)
+                    .with_context(|| format!("creating loader.conf at {}", loader_conf_path.display()))?;
+                writeln!(loader_conf, "default auto")?;
```
cgwalters: This one should probably come from the container image and not be hardcoded in bootupd.

src/efi.rs
p5: Ah, right.  I was confused as there were no boot entries or kernels being created in /boot during `bootc install` (need to figure out where that happens), so thought it was bootupd's responsibility to set them up the first time.

Done in 
[bfd1cd4](https://github.com/coreos/bootupd/pull/978/commits/bfd1cd49d52623e7e8173c647df8b48a1ed747bc)

src/efi.rs
p5: `bootctl install` actually creates this file too, so what I'll do is allow the user (or distro) to put a custom config in `/usr/lib/bootupd/systemd-boot/loader.conf`, and if that file is detected, bootupd will use it.  Otherwise, the default (empty file) will be used.

p5: Removed the hardcoded config in [bfd1cd4](https://github.com/coreos/bootupd/pull/978/commits/bfd1cd49d52623e7e8173c647df8b48a1ed747bc)

Added optional "copy if present" in [3af4186](https://github.com/coreos/bootupd/pull/978/commits/3af41868bd3b6a2c92b2eb85136918a45a80ce12)

cgwalters:
>  Figure out the public API we wish to expose for enabling sd-boot

First I think we can add a *build time* option which entirely disables GRUB support. Then it's simple on UEFI platforms and where systemd-boot is in the target, we use it by default.

But for dynamic detection one way to do this would be to detect if the target system is both:

- Using EFI and does not have any BIOS partition
- Has https://uapi-group.org/specifications/specs/boot_loader_specification/#standard-conformance-marker-file

Also of course, if the target image doesn't have grub then we should default to systemd-boot.

p5:
Would you mind if I went straight for the dynamic detection?  
I'm trying to gate grub behind a feature flag, but doing so rapidly turns the codebase into a mess because it also means disabling bios.

src/efi.rs
```diff
@@ -50,6 +50,10 @@ pub(crate) const SHIM: &str = "shimriscv64.efi";
 /// Systemd boot loader info EFI variable names
 const LOADER_INFO_VAR_STR: &str = "LoaderInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
 const STUB_INFO_VAR_STR: &str = "StubInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
+#[cfg(target_arch = "aarch64")]
+pub(crate) const SYSTEMD_BOOT_EFI: &str = "usr/lib/systemd/boot/efi/systemd-bootaarch64.efi";
+#[cfg(target_arch = "x86_64")]
+pub(crate) const SYSTEMD_BOOT_EFI: &str = "usr/lib/systemd/boot/efi/systemd-bootx64.efi";
```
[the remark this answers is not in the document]
p5: From what I could tell, systemd-boot does not support risc-v.  Will try disabling the entire sd-boot feature for ppc and riscv.  Or maybe I define it as a value that will _never_ exist just to make Rust happy?

Edit:  Seems to need the introduction of a build.rs file.  Will add it but I don't believe it's strictly necessary as long as we don't enable the systemd-boot feature.

src/systemdbootconfigs.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/// Install files required for systemd-boot
+/// This mostly proxies the bootctl install command
+#[context("Installing systemd-boot")]
+pub(crate) fn install(src_root: &openat::Dir, esp_path: &openat::Dir) -> Result<()> {
+    let esp_path_buf = esp_path.recover_path().context("ESP path is not valid")?;
+    let esp_path_str = esp_path_buf
+        .to_str()
+        .context("ESP path is not valid UTF-8")?;
+    let status = std::process::Command::new("bootctl")
```
travier, line 16: We won't be able to use `bootctl isntall` by default here as that's skipping shim entirely.

p5: Ah, understood.

Would you suggest doing something like [this](https://github.com/systemd/systemd/issues/27234#issuecomment-2424826157) (but pointing to the right paths)?

```
efibootmgr --disk /dev/sda --part 1 --create --label "Linux Secure Boot" --loader '\EFI\Systemd\shimx64.efi' --unicode '\EFI\Systemd\systemd-bootx64.efi'
```

Will need to figure out a way to identify the shim without reintroducing the RPM code.
Though with this approach, we can probably still use `bootctl install`, but optionally followed by copying the shim and running efibootmgr if the shim is detected.

travier: Ah! I did not know that we could pass arguments to shim! If this works then this is great. We should probably install the systemd-boot binary under the os path: `/EFI/<os|fedora|arch>/systemd-bootx64.efi` to not have a "global" instance of it.

travier:
Note that we should not rely on systemd's [systemd-boot-update.service](https://github.com/systemd/systemd/blob/main/units/systemd-boot-update.service) either for updates as it does not update shim.

src/systemdbootconfigs.rs
cgwalters: > We won't be able to use bootctl isntall by default here as that's skipping shim entirely.

For sealed images we do want to skip shim though right?

> Will need to figure out a way to identify the shim without reintroducing the RPM code.

I think it should be the same as detecting grub ideally. I'm not sure we care about grub-but-not-shim (hopefully?) but we do care about the matrix of:

- shim+grub
- shim+systemd-boot
- systemd-boot

So...I think we can reasonably do this by detecting if we have `shim*.efi` in the image.

> Would you suggest doing something like [https://github.com/systemd/systemd/issues/27234#issuecomment-2424826157] (but pointing to the right paths)?

Hmm AFAIK efibootmgr is primarily for manipulating the EFI firmware, which is distinct from the files on disk. We already have code which...currently hardcodes targeting shim, which would definitely have to change for sealed images.

src/systemdbootconfigs.rs
```diff
@@ -0,0 +1,61 @@
...
+    match configdir.open_file("loader.conf") {
+        Ok(mut src_file) => {
+            let mut dst_file = std::fs::File::create(&dst_loader_conf)
+                .context("Failed to create loader.conf in ESP")?;
+            std::io::copy(&mut src_file, &mut dst_file)
+                .context("Failed to copy loader.conf to ESP")?;
+            log::info!("loader.conf copied to {}", dst_loader_conf.display());
+            Ok(())
+        }
+        Err(e) => {
```
cgwalters, line 56: We should only match if this is `ENOENT`, but propagate other errors.

For cap-std-ext we have https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.open_optional but we're not yet using cap-std here...

src/systemdbootconfigs.rs
```diff
@@ -0,0 +1,61 @@
...
+    let esp_path_buf = esp_path.recover_path().context("ESP path is not valid")?;
+    let esp_path_str = esp_path_buf
+        .to_str()
+        .context("ESP path is not valid UTF-8")?;
+    let status = std::process::Command::new("bootctl")
+        .args(["install", "--esp-path", esp_path_str])
+        .status()
+        .context("Failed to execute bootctl")?;
+
+    if !status.success() {
```
cgwalters, line 21: Can use our helper trait `.run_with_cmd_context()` or so

src/systemdbootconfigs.rs
```diff
@@ -0,0 +1,61 @@
...
+use anyhow::{Context, Result};
+use fn_error_context::context;
+
+const CONFIG_DIR: &str = "usr/lib/bootupd/systemd-boot";
+
+/// Install files required for systemd-boot
+/// This mostly proxies the bootctl install command
+#[context("Installing systemd-boot")]
+pub(crate) fn install(src_root: &openat::Dir, esp_path: &openat::Dir) -> Result<()> {
+    let esp_path_buf = esp_path.recover_path().context("ESP path is not valid")?;
```
cgwalters, line 12: Instead of using `recover_path()` a trick is to pass the fd to the child via its working directory. Again in cap-std-ext we have https://docs.rs/cap-std-ext/latest/cap_std_ext/cmdext/trait.CapStdExtCommandExt.html#tymethod.cwd_dir

build.rs
```diff
@@ -0,0 +1,9 @@
+fn main() {
+    if std::env::var("CARGO_FEATURE_SYSTEMD_BOOT").is_ok() {
+        if let Ok(arch) = std::env::var("CARGO_CFG_TARGET_ARCH") {
+            if arch.starts_with("riscv") {
+                panic!("The systemd-boot feature is not supported on RISC-V.");
```
cgwalters, line 5: Hum, this whole build.rs seems like it could use a comment for *why*; I'm sure some RISC-V systems aren't UEFI but it looks like at least some are working on it e.g. https://archive.fosdem.org/2021/schedule/event/firmware_uor/#:~:text=Online%20%2F%206%20%26%207%20February%202021,EDK2%20UEFI%20on%20RISC%2DV

Do we actually fail to build on RISC-V in this case?

src/efi.rs
```diff
@@ -365,6 +363,31 @@ impl Component for Efi {
...
 
+        match bootloader {
+            #[cfg(feature = "systemd-boot")]
+            Bootloader::SystemdBoot => {
+                log::info!("Installing systemd-boot via bootctl");
+                let esp_dir = openat::Dir::open(&destpath)?;
+                crate::systemdbootconfigs::install(src_root, &esp_dir)?;
+                return Ok(InstalledContent {
+                    meta: ContentMetadata {
+                        timestamp: Utc::now(),
```
cgwalters, line 374: This timestamp should be sourced from the `bootctl` binary

src/bootupd.rs
```diff
@@ -715,6 +755,41 @@ fn strip_grub_config_file(
...
 
+/// Determine whether the necessary bootloader files are present for GRUB.
+fn has_grub(source_root: &openat::Dir) -> bool {
+    source_root.open_file(bios::GRUB_BIN).is_ok()
+}
+
+/// Determine whether the necessary bootloader files are present for systemd-boot.
+#[cfg(feature = "systemd-boot")]
+fn has_systemd_boot(source_root: &openat::Dir) -> bool {
+    source_root.open_file(efi::SYSTEMD_BOOT_EFI).is_ok()
```
cgwalters, line 766: Super minor but doing a `stat()` instead of `open()` is slightly more efficient i.e. use `source_root.metadata()`

cgwalters:
> This PR started off as "a super thin wrapper for systemd-boot" but seems to be turning into a larger implementation. I'll certainly keep looking into it, but if the Bootc team are working to a tighter schedule, please feel free to forge on ahead.

Yeah makes sense. Do you want someone else on the team to pick this up? And thanks so much for starting it and sorry for the delay in reviews!

p5:
> Do you want someone else on the team to pick this up?

I'm more than happy to continue looking into this, but I am not wanting to block the bootc systemd-boot implementation with this PR if your team is planning to implement it sooner.  Your team will have far more context in this space and know how the different parts should fit together.

I'm tempted to create a separate branch/draft PR though since people seem to be starting to targeting this one for experimental non-Red-Hat-based images which I don't want to be breaking (although I've made it clear in Discord that it shouldn't be trusted yet :)).

src/bootupd.rs
```diff
@@ -715,6 +755,41 @@ fn strip_grub_config_file(
     Ok(())
 }
 
+/// Determine whether the necessary bootloader files are present for GRUB.
+fn has_grub(source_root: &openat::Dir) -> bool {
+    source_root.open_file(bios::GRUB_BIN).is_ok()
```
smoser: this causes build error on arch other than powerpc64 or x86_64 due to arch specific use of bios crate on line 1.

p5: As a temporary workaround, I've hardcoded the GRUB_BIN value in the function.  Should compile on other architectures now.  

Will be fixed properly when I get chance to implement some of the larger refactors requested in other comments.

As you can probably tell, you're the first to try and run on ARM so please let me know if you face any other issues!

smoser: why isn't there a "thank you" reaction emoji?
thank you.

PR #983, opened 2025-08-15
packagesystem: make `rpm` as build-time configuration

HuijingHei:
This continuous from https://github.com/coreos/bootupd/pull/977,
inspired by Colin's https://github.com/coreos/bootupd/pull/977#discussion_r2279416177.

### DIFF
--- a/src/packagesystem.rs
+++ b/src/packagesystem.rs
@@ -101,6 +101,20 @@ where
     rpm_parse_metadata(&rpmout.stdout)
 }
 
+fn split_name_version(input: &str) -> Option<(String, String)> {
+    // assume it is "grub2-tools-1:2.06-110.el9.x86_64"
+    // strip .arch
+    let main = input.rsplit_once('.')?.0;
+
+    // find last two '-'
+    let mut parts = main.rsplitn(3, '-');
+    let release = parts.next()?; // after last '-'
+    let version = parts.next()?; // between last two '-'
+    let name = parts.next()?; // the rest (may contain '-')
+
+    Some((name.to_string(), format!("{version}-{release}")))
+}
+
 fn parse_evr(pkg: &str) -> Module {
     // assume it is "grub2-1:2.12-28.fc42" (from usr/lib/efi)
     if !pkg.ends_with(std::env::consts::ARCH) {
@@ -112,8 +126,15 @@ fn parse_evr(pkg: &str) -> Module {
     }
 
     let (name_str, rpm_evr) = {
-        let nevra = rpm::Nevra::parse(pkg);
-        (nevra.name().to_string(), nevra.evr().to_string())
+        #[cfg(not(feature = "rpm"))]
+        {
+            split_name_version(pkg).unwrap()
+        }
+        #[cfg(feature = "rpm")]
+        {
+            let nevra = rpm_rs::Nevra::parse(pkg);
+            (nevra.name().to_string(), nevra.evr().to_string())
+        }
     };
 
     let (name, _) = name_str.split_once('-').unwrap_or((&name_str, ""));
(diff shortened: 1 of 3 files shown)


src/packagesystem.rs
```diff
@@ -69,94 +69,103 @@ where
...
+
+fn parse_evr_vec(input: &str) -> Vec<Package> {
+    let mut pkgs: Vec<Package> = input
+        .split(',')
+        .map(|pkg| parse_evr(pkg)) // parse_evr returns owned Package
+        .collect();
+    // Remove duplicates while preserving order
+    let mut seen = BTreeSet::new();
+    pkgs.retain(|pkg| seen.insert((pkg.name.clone(), pkg.version.clone())));
+    pkgs
```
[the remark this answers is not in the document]
HuijingHei: Updated

src/packagesystem.rs
```diff
@@ -69,94 +69,142 @@ where
...
+        return Package {
+            name: name.to_string(),
+            rpm_evr: evr.to_string(),
+        };
+    }
+
+    let (name_str, rpm_evr) = {
+        #[cfg(not(feature = "rpm"))]
+        {
+            let (name, epoch, version, release, _arch) = parse_values(pkg);
```
cgwalters: Let's use https://docs.rs/uapi-version/latest/uapi_version/ to parse versions when we're not using rpm - the idea here is that other package systems will have formats different than rpm's. (dpkg is similar but definitely not the same)

The UAPI version scheme is attempting to create one that can be used exactly by projects like bootupd (also systemd and other uapi projects). Now in practice one would need to create a "mapping" from other systems to it. The spec mentions some incompatibilities, but I think one could create a reliable mapping.

Actually at least we could ask for Fedora that `/usr/lib/efi` uses the UAPI version spec and not RPM which would make it much clearer how to use that scheme on other OSes too.

src/packagesystem.rs
```diff
@@ -69,94 +69,142 @@ where
...
-        .split(',')
-        .map(|pkg| {
-            // assume it is like "grub2-1:2.12-28.fc42,shim-15.8-3"
-            if !pkg.ends_with(std::env::consts::ARCH) {
-                let (name, version_release) = pkg.split_once('-').unwrap_or((pkg, ""));
-                let evr = Evr::parse(version_release);
-                return (name.to_string(), evr);
-            }
+#[derive(Debug, Eq, PartialEq)]
+struct Package {
```
cgwalters: Elsewhere in the code here we use the term "component" in some cases because not everything comes from a "package". Though we also use "component" to distinguish bios/efi, so it'd likely be confusing if we reused that term here.

I guess if I could go back I'd say "module" or "backend" perhaps for bios/efi and "component" instead of "package".

But anyways it's probably going to be way less confusing to call this Package so let's keep it as is.

src/packagesystem.rs
cgwalters: There's a subtlety here in that we probably can't easily add the *name* into a single string and re-parse it into a uapi version as that spec doesn't standardize a separator mechanism.

It's quite tricky as we really have two different cases:

- Status quo of a single string that serializes name+version for multiple packages
- /usr/lib/efi which cleanly splits packages from each other *and* splits the name vs version cleanly

I think we should treat the status quo "single big version string" as the legacy/degenerate case and map other things to it.

Ideally anyone coming *now* into bootupd from a non-rpm distribution ideally can just implement `/usr/lib/efi` with uapi versions or so and in that case we don't need to deal with the legacy of the single `version` string.

HuijingHei: > It's quite tricky as we really have two different cases:
> 
> * Status quo of a single string that serializes name+version for multiple packages
> * /usr/lib/efi which cleanly splits packages from each other _and_ splits the name vs version cleanly
> 
> I think we should treat the status quo "single big version string" as the legacy/degenerate case and map other things to it.

Let me start with `introducing the parallel versions struct with rpm-evr` firstly, then we can ignore to parse the string.

src/packagesystem.rs
HuijingHei: It is package version saved in `ContentMetadata` struct, looks `module` is more clean and neat.

cgwalters:
Looks like this one needs a rebase

PR #986, opened 2025-08-20
Add `versions` to `ContentMetadata`

HuijingHei:
packagesystem: add struct `Module` about package name and evr

Also add `uapi_version::Version` to parse versions according to
Colin's comment https://github.com/coreos/bootupd/pull/983#discussion_r2284399377

---

Add `versions` to `ContentMetadata`

Inspired by Colin's comment https://github.com/coreos/bootupd/pull/977#discussion_r2278838151
We need `versions` to save struct like:
```
  "versions": [
    {"name": "pkg1", "rpm-evr": "1.0"},
    {"name": "pkg2", "rpm-evr": "2.0"}
  ]
```
Then we read and compare versions, do not need to parse the
version in str format.

Co-worked by walters@verbum.org

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -155,7 +155,7 @@ type Components = BTreeMap<&'static str, Box<dyn Component>>;
 #[allow(clippy::box_default)]
 /// Return the set of known components; if `auto` is specified then the system
 /// filters to the target booted state.
-pub(crate) fn get_components_impl(auto: bool) -> Components {
+pub(crate) fn get_components_impl(_auto: bool) -> Components {
     let mut components = BTreeMap::new();
 
     fn insert_component(components: &mut Components, component: Box<dyn Component>) {
@@ -164,7 +164,7 @@ pub(crate) fn get_components_impl(auto: bool) -> Components {
 
     #[cfg(target_arch = "x86_64")]
     {
-        if auto {
+        if _auto {
             let is_efi_booted = crate::efi::is_efi_booted().unwrap();
             log::info!(
                 "System boot method: {}",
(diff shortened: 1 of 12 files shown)


src/packagesystem.rs
```diff
@@ -40,9 +67,14 @@ fn rpm_parse_metadata(stdout: &[u8]) -> Result<ContentMetadata> {
         s.push_str(n);
         s
     });
+
+    // Map the version into Module struct
+    let mut modules_vec: Vec<Module> = pkgs.keys().map(|pkg_str| parse_evr(pkg_str)).collect();
```
HuijingHei, line 72: Not sure if should map the version str into Module struct when using `rpm -q`.

cgwalters (approved):
Looks sane to me!

HuijingHei:
Rebase the main branch to include CI workflow fix, and will fix the CI failed.

HuijingHei:
Rebase to main branch and update test, and will merge this as the CI is passed.

src/bootupd.rs
```diff
@@ -163,7 +164,7 @@ pub(crate) fn get_components_impl(auto: bool) -> Components {
 
     #[cfg(target_arch = "x86_64")]
     {
-        if auto {
+        if _auto {
```
travier, line 167: Nit: Looks weird? Why add the `_`?

src/packagesystem.rs
travier: Hum, I don't understand the question here. Do you mean that if we do `rpm -q` we should rely on timestamps and not versions?

src/bootupd.rs
HuijingHei: The change to `_auto` is to fix the warning that on the other platforms `auto` is unused.

src/packagesystem.rs
HuijingHei: I meant not sure if it is necessary to do the map, and it will compare using version if we do not have the Module struct. But agree to do the map.
Copy the change here:
```
        if let (Some(self_versions), Some(target_versions)) = (&self.versions, &target.versions) {
            compare_package_slices(self_versions, target_versions)
        } else {
            compare_package_versions(&self.version, &target.version)
        }
```

PR #990, opened 2025-08-29
install: pull grubconfigs from the source root

jbtrystram:
When doing install with the `--src-root` argument, make sure grub static configs are copied from the source root rather than the running root.

See https://github.com/bootc-dev/bootc/issues/1559


src/grubconfigs.rs
```diff
@@ -103,8 +116,9 @@ pub(crate) fn install(
             .sub_dir_optional("boot/efi/EFI")
             .context("Opening /boot/efi/EFI")?;
         if let Some(efidir) = dest_efidir {
+            let src_static_efi_path = configdir.recover_path()?.join("grub-static-efi.cfg");
             efidir
-                .copy_file(&Path::new(CONFIGDIR).join("grub-static-efi.cfg"), target)
+                .copy_file(&src_static_efi_path, target)
                 .context("Copying static EFI")?;
```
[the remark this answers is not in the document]
jbtrystram: Nice suggestion ! applied

src/grubconfigs.rs
```diff
@@ -25,6 +26,7 @@ const GRUBCONFIG_FILE_MODE: u32 = 0o600;
 #[context("Installing static GRUB configs")]
 pub(crate) fn install(
     target_root: &openat::Dir,
+    src_root: Option<&openat::Dir>,
```
cgwalters, line 29: Yeah one problem we keep getting bit by here is that the openat crate is old and unmaintained, I'd like to move this project over to cap-std which we use *extensively* in bootc at least.

cgwalters (approved):

src/grubconfigs.rs
jbtrystram: I wasn't aware of that, good to know. I just went with what was existing in the code :)

PR #991, opened 2025-08-29
efi: Extend `get_efi_vendor()` to support `usr/lib/efi`

HuijingHei:
Prepare for step2 (in https://github.com/coreos/bootupd/issues/926) to install efi files from `usr/lib/efi`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,11 @@ FROM $base
 # Clean out the default to ensure we're using our updated content
 RUN rpm -e bootupd
 COPY --from=build /out/ /
+# Remove /var/roothome as workaround
+RUN <<EORUN
+set -xeuo pipefail
+[ -d /var/roothome ] && rm -rf /var/roothome
+EORUN
 # Sanity check this too
 RUN bootc container lint --fatal-warnings
 
--- a/src/component.rs
+++ b/src/component.rs
@@ -51,7 +51,7 @@ pub(crate) trait Component {
     /// This will be run during a disk image build process.
     fn install(
         &self,
-        src_root: &openat::Dir,
+        src_root: &str,
         dest_root: &str,
         device: &str,
         update_firmware: bool,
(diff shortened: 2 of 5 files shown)


src/component.rs
```diff
@@ -225,6 +225,34 @@ mod tests {
                 assert_eq!(x.is_err(), true);
                 std::fs::remove_dir_all(tdp_updates.join("EFI/centos"))?;
                 assert_eq!(component.get_efi_vendor(&td)?, Some("fedora".to_string()));
+                {
+                    let td_usr = tdp.join("usr/lib/efi");
```
jbtrystram: you can keep the `Dir` struct from openat and open a subdirectory : https://github.com/coreos/bootupd/blob/main/src/grubconfigs.rs#L48

src/component.rs
```diff
@@ -225,6 +225,34 @@ mod tests {
...
+                    let td_vendor = td_usr.join("shim/15.8-3/EFI/centos");
```
jbtrystram: then you can use `create_dir` to create a subpath : https://docs.rs/openat/latest/openat/struct.Dir.html#method.create_dir

src/component.rs
HuijingHei: Updated, thank you!

src/component.rs
HuijingHei: Updated, thank you!

HuijingHei:
I will merge this if there is no objection.

Dockerfile
```diff
@@ -23,6 +23,11 @@ FROM $base
 # Clean out the default to ensure we're using our updated content
 RUN rpm -e bootupd
 COPY --from=build /out/ /
+# Remove /var/roothome as workaround
```
travier, line 26: What error is this working around?

travier (approved):
Not tested but looks OK.

Dockerfile
cgwalters: https://github.com/bootc-dev/bootc/issues/1546

src/efi.rs
```diff
@@ -543,11 +545,21 @@ impl Component for Efi {
...
+        let updates = sysroot_path.join(component_updatedirname(self));
+
+        let target = if let Some(p) = [efi_lib.as_path(), updates.as_path(), sysroot_path]
+            .into_iter()
+            .find(|p| p.exists())
+        {
+            p
+        } else {
+            bail!("Failed to find valid target path");
+        };
```
cgwalters, line 560: This is just `.ok_or_else(|| anyhow!("Failed..."))`

src/efi.rs
```diff
@@ -543,11 +545,21 @@ impl Component for Efi {
         }
     }
 
-    fn get_efi_vendor(&self, sysroot: &openat::Dir) -> Result<Option<String>> {
-        let updated = sysroot
-            .sub_dir(&component_updatedirname(self))
-            .context("opening update dir")?;
-        let shim_files = find_file_recursive(updated.recover_path()?, SHIM)?;
+    fn get_efi_vendor(&self, sysroot: &str) -> Result<Option<String>> {
```
cgwalters, line 548: I don't follow why we needed to switch to taking an ambient path here.

HuijingHei: Will try to update to Path later, just thought it could be convenient to pass.

PR #992, opened 2025-08-31
efi: Extend `install()` to support `usr/lib/efi`

HuijingHei:


HuijingHei:
Hold on, the change in the filetree path `grub2/1:2.12-41.fc44/EFI/fedora/grubx64.efi` (from `/boot/bootupd-state.json`) will make the update tricky, will do more testing.

src/filetree.rs
```diff
@@ -68,16 +68,25 @@ const DEFAULT_FILE_MODE: u32 = 0o700;
 use crate::sha512string::SHA512String;
 
 /// Metadata for a single file
-#[derive(Clone, Serialize, Deserialize, Debug, Hash, PartialEq)]
+#[derive(Clone, Serialize, Deserialize, Debug)]
 #[serde(rename_all = "kebab-case")]
 pub(crate) struct FileMetadata {
+    /// File source path
```
cgwalters, line 74: Can you explain a bit more what this is, e.g. include an example and also a rationale for why we need this new field?

cgwalters: OK sorry I see it's here https://github.com/coreos/bootupd/pull/992#issuecomment-3265450522

cgwalters (approved):

Issue #994, opened 2025-09-01
Safety guard: Make sure that we have at least shim & grub packages as part of the update


travier:
In the future we will likely want to be generic, but as a safety check for now, let's make sure that we have at least shim and grub as part of the update payload on x86_64 & aarch64 at least.

We should fail at image build time (generate metadata), not at deploy time.

This is to prevent a future instance of: https://github.com/fedora-silverblue/issue-tracker/issues/671

HuijingHei:
This make sense, and also transfer `usr/lib/ostree-boot` to `usr/lib/efi` seems necessary.

cgwalters:
Maybe, but...what would have fixed this is gating CI before shipping updates which gets into https://gitlab.com/fedora/bootc/tracker/-/issues/48 and also gating packages *pre merge* on basic image mode/bootc sanity tests.

HuijingHei:
Add test to verify the bootupd EFI has more than one component installed, see https://gitlab.com/fedora/bootc/base-images/-/merge_requests/286

PR #995, opened 2025-09-02
efi: transfer `usr/lib/ostree-boot` to `usr/lib/efi`

HuijingHei:
### efi: transfer `usr/lib/ostree-boot` to `usr/lib/efi`
This is required by step2 of https://github.com/coreos/bootupd/issues/926#issuecomment-3095110698, with the change, the files under legacy `/usr/lib/ostree-boot/efi/EFI` will be changed to `/usr/lib/efi/<component>/<version>/EFI/`.

---

### efi: require minimum of 2 components for update metadata generation

Enhance `generate_update_metadata()` with validation to ensure
at least two packages are provided, preventing the creation of
incomplete or invalid update payloads.
See https://github.com/coreos/bootupd/issues/994

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -757,7 +757,11 @@ fn transfer_ostree_boot_to_usr(sysroot: &Path) -> Result<()> {
     let ostreeboot_efi = Path::new(ostreeutil::BOOT_PREFIX).join("efi");
     let ostreeboot_efi_path = sysroot.join(&ostreeboot_efi);
 
-    for entry in WalkDir::new(ostreeboot_efi_path.join("EFI")) {
+    let efi = ostreeboot_efi_path.join("EFI");
+    if !efi.exists() {
+        return Ok(());
+    }
+    for entry in WalkDir::new(&efi) {
         let entry = entry?;
 
         if entry.file_type().is_file() {


src/packagesystem.rs
```diff
@@ -101,6 +101,27 @@ where
     rpm_parse_metadata(&rpmout.stdout)
 }
 
+/// Query the rpm database and get package <name>/<epoch>:<version>-<release>.
+pub(crate) fn query_file(sysroot_path: &str, path: &str) -> Result<String> {
+    let mut c = ostreeutil::rpm_cmd(sysroot_path)?;
+    c.args([
+        "-q",
+        "--queryformat",
+        "%{NAME}/%{EPOCH}:%{VERSION}-%{RELEASE}",
```
travier: We could split on space ` ` here as package names should never have spaces:

```
%{NAME} %{EPOCH} %{VERSION} %{RELEASE}
```

HuijingHei: How about this one?
```
$ rpm -qf --queryformat "%{NAME} %{EVR}" -f /usr/bin/grub2-install 
grub2-tools 1:2.12-32.fc42
$ rpm -qf --queryformat "%{NAME} %{EVR}" -f /bin/bash
bash 5.2.37-1.fc42
```

src/efi.rs
```diff
@@ -451,18 +452,45 @@ impl Component for Efi {
...
+            let efisrc = sysroot_path.join(ostreeutil::BOOT_PREFIX).join("efi/EFI");
+            if efisrc.exists() {
+                let files = WalkDir::new(&efisrc)
+                    .into_iter()
+                    .filter_map(Result::ok)
+                    .filter(|e| e.file_type().is_file())
+                    .map(|e| e.path().to_path_buf())
+                    .collect::<Vec<_>>();
+
+                if !files.is_empty() {
```
HuijingHei: On rawhide with latest grub2 & shim, there are existing empty directories `/usr/lib/ostree-boot/efi/EFI/{BOOT,fedora}`, this is because when installing `efi-filesystem`, `boot/` is transferred to `usr/lib/ostree-boot`, need to remove the directory `/usr/lib/ostree-boot/efi/EFI`. Let me know if I misunderstood this.
```
bash-5.3# rpm -ql efi-filesystem
/boot/efi
/boot/efi/EFI
/boot/efi/EFI/BOOT
/boot/efi/EFI/fedora
```

src/efi.rs
```diff
@@ -451,18 +452,45 @@ impl Component for Efi {
...
+            .context("Opening usr/lib/ostree-boot")?
+        {
+            let cruft = ["loader", "grub2"];
+            for p in cruft.iter() {
+                ostreeboot.remove_all_optional(p)?;
+            }
+
+            let efisrc = sysroot_path.join(ostreeutil::BOOT_PREFIX).join("efi/EFI");
+            if efisrc.exists() {
+                let files = WalkDir::new(&efisrc)
```
cgwalters: I think it'd be cleaner to have `transfer_ostree_boot_to_usr` walk the directory instead.

src/efi.rs
```diff
@@ -451,18 +452,45 @@ impl Component for Efi {
...
+                    .into_iter()
+                    .filter_map(Result::ok)
```
cgwalters: This could silently mask errors (unlikely but possible).

src/efi.rs
```diff
@@ -765,9 +759,53 @@ fn get_efi_component_from_usr<'a>(
...
+    if components.len() == 0 {
+        return Ok(None);
+    }
     components.sort_by(|a, b| a.name.cmp(&b.name));
 
-    Ok(components)
+    Ok(Some(components))
+}
+
+fn transfer_ostree_boot_to_usr(sysroot: &Path, ostreeboot_files: &Vec<PathBuf>) -> Result<()> {
```
cgwalters: This could use a doc comment.

src/efi.rs
HuijingHei: Updated, thank you!

cgwalters (approved):

cgwalters:
I think this change was a good idea, but it's caused a multi-day breakage in bootc CI because of the version interlock it creates where older bootupd fails to understand systems whose layout is generated with newer bootupd.

See https://github.com/bootc-dev/bootc/pull/1926

As of right now it seems this change is rolled out to most targets except Fedora 42, but it didn't propagate consistently or quickly of course.

In retrospect, what we probably should have done is made actually *applying* this change require an opt-in in the base image (config file say), or `bootupctl backend install --format=2` or so. It's not strictly speaking too late to do that.

We're also working on https://github.com/bootc-dev/bootc/pull/1816 which will obviate most issues of skew, but it's hard to avoid in the general case.

With format changes again, the key thing is to decouple *understanding* the change from application of the format.

HuijingHei:
> In retrospect, what we probably should have done is made actually _applying_ this change require an opt-in in the base image (config file say), or `bootupctl backend install --format=2` or so. It's not strictly speaking too late to do that.

Besides this, does it make sense to update bootupd on f42 and c9s?

HuijingHei:
> In retrospect, what we probably should have done is made actually applying this change require an opt-in in the base image (config file say), or bootupctl backend install --format=2 or so. It's not strictly speaking too late to do that.

Could you help to create such an issue for this?

> it could still make sense to do a merge.

Agree!

travier:
> `bootupctl backend install --format=2`

I think you meant `generate-metadata` here

Issue #972, opened 2025-07-24
Do not try to update EFI variables is empty or read only


travier:
According to https://bugzilla.redhat.com/show_bug.cgi?id=2333190, systems sometimes have read only variables and thus any efibootmgr command would fail to update them. We should either:
- make this command fail open (i.e. always ignore efibootmgr failures)
- or check for the read only state and skip it

See also the following issues that are likely related:
- https://github.com/coreos/bootupd/issues/907 & https://github.com/coreos/bootupd/pull/951
- https://github.com/fedora-silverblue/issue-tracker/issues/645
- https://github.com/fedora-silverblue/issue-tracker/issues/614

HuijingHei:
And also https://issues.redhat.com/browse/RHEL-109843, on RT kernel has `CONFIG_EFI_DISABLE_RUNTIME=y`, that means Disable EFI runtime services support by default (from https://cateee.net/lkddb/web-lkddb/EFI_DISABLE_RUNTIME.html), the workaround is add `efi=noruntime` kernel argument.

PR #998, opened 2025-09-05
efi: Do not update EFI variables if not mountpoint, empty or read only

HuijingHei:
Fixes https://github.com/coreos/bootupd/issues/972


src/efi.rs
```diff
@@ -381,7 +402,10 @@ impl Component for Efi {
             .run()?;
         if update_firmware {
             if let Some(vendordir) = self.get_efi_vendor(&Path::new(src_root))? {
-                self.update_firmware(device, destd, &vendordir)?
+                // Only update if EFI firmware variables is readwrite and not empty
+                if self.is_efivars_rw()? {
```
cgwalters: I think it'd be cleaner to just do this check in `update_firmware`; then we should be able to assume the efi directory exists and just need to check if it's readonly and no-op then.

HuijingHei: SGTM, thank you! Also do nothing if the directory is empty (see [bug#2333190](https://bugzilla.redhat.com/show_bug.cgi?id=2333190#c3))

src/efi.rs
```diff
@@ -146,6 +149,19 @@ impl Efi {
...
+        let efi = sysroot
+            .open_dir(EFIVARFS.strip_prefix("/").unwrap())
+            .context("Opening efivars dir")?;
+        let st = rustix::fs::fstatvfs(efi.as_fd())?;
+        // Do nothing if efivars is readonly or empty
+        // See https://github.com/coreos/bootupd/issues/972
+        if st.f_flag.contains(StatVfsMountFlags::RDONLY)
+            || std::fs::read_dir(EFIVARFS)?.next().is_none()
+        {
+            println!("efivars is readonly or empty dir\nSkipped updating EFI firmware variables");
```
travier: I think those should be two calls to log::info?

src/efi.rs
```diff
@@ -381,7 +397,7 @@ impl Component for Efi {
             .run()?;
         if update_firmware {
             if let Some(vendordir) = self.get_efi_vendor(&Path::new(src_root))? {
-                self.update_firmware(device, destd, &vendordir)?
+                self.update_firmware(device, destd, &vendordir)?;
```
travier: Hum, this looks weird as we did not change the return signature of this function.

src/efi.rs
HuijingHei: SGTM, updated!

src/efi.rs
HuijingHei: Seems that both works, and with `;` makes it explicit that the result is only used for error propagation (?) and the success value is discarded. WDYT?
Edit: I am OK to keep it as it is (without ;).

HuijingHei: OK, just keep it as it is without `;`, this also works.

HuijingHei:
Update info to make it shorter, ready to review, thank you!

PR #1002, opened 2025-09-11
package: keep the full name if not `grub2` or `shim`

HuijingHei:
When doing tests for https://github.com/coreos/bootupd/pull/995, find that the package `test-bootupd-payload-1.0-1.x86_64` is parsed to `test-1.0-1` that is confused, maybe should keep the full name if it is not `grub2` or `shim`.

### DIFF
--- a/src/packagesystem.rs
+++ b/src/packagesystem.rs
@@ -116,7 +116,12 @@ fn parse_evr(pkg: &str) -> Module {
         (nevra.name().to_string(), nevra.evr().to_string())
     };
 
-    let (name, _) = name_str.split_once('-').unwrap_or((&name_str, ""));
+    // Only cut the packages name that we know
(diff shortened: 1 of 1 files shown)


src/packagesystem.rs
```diff
@@ -116,7 +116,12 @@ fn parse_evr(pkg: &str) -> Module {
         (nevra.name().to_string(), nevra.evr().to_string())
     };
 
-    let (name, _) = name_str.split_once('-').unwrap_or((&name_str, ""));
+    // Only cut the packages name that we know
+    let (name, _) = if ["grub2", "shim"].iter().any(|p| name_str.starts_with(p)) {
```
travier, line 120: I would do something like
```
while true
    next = split.next()
    if next.starts_with_number() == false
        name.join("-", next)
    else
        break
    fi
endwhile

then is next epoch or version, etc.
```

Issue #1011, opened 2025-09-28
Handle component removal when doing updates


HuijingHei:
When doing the updates, should consider that if one component is removed, we should not update it any more.

Copy Colin's [comment](https://github.com/coreos/bootupd/pull/1010/files#r2383383813):
```
I think actually the problem is we're iterating over the static list of components, when actually I think we should just accept whatever the updated components in the updates set are to start with or so.

Although it raises a different point - it would probably make sense in the future to handle the case where a component is removed. (e.g. an OS update drops out BIOS support and we would then in theory clear out the BIOS partition e.g.)

And to do that, we do need to iterate over the components we know (as this is doing) and intersect them with the update set.
```

HuijingHei:
Close this as done in https://github.com/coreos/bootupd/pull/1018

PR #1018, opened 2025-10-28
bootupd: Handle component removal when doing updates

HuijingHei:
When doing the updates, should consider that if one component is removed, we should not update it any more.
See https://github.com/coreos/bootupd/issues/1011

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -391,7 +391,12 @@ pub(crate) fn status() -> Result<Status> {
         // adoptable list, and adoption proceeds automatically.
         //
         // Therefore, calling `query_adopt_state()` alone is sufficient.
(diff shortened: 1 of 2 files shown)


joelcapitao (approved):
LGTM

travier:
This could make sense for UKI / sealed images where we would remove BIOS support.

HuijingHei:
> This is kind of deceptively simple looking but there's some nontrivial risk here.
> 
> I saw [bootc-dev/bootc#1734](https://github.com/bootc-dev/bootc/issues/1734) get filed recently which I think tangentially relates to this.
> 
> How about adding an "in-use check" too? Detect if the system is booted via EFI, then we refuse to remove that component, etc.

Will make the change next Monday.

Issue #927, opened 2025-05-15
Create "direct" install sub command that will be called in a posttrans on package mode installations


travier:
As part of https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1, for package mode installations, we will update the bootloader (copy from /usr to boot or ESP) as part of the posttrans scriptlet.

This is a slightly different path as the current update path where we have the systemd-run logic, and we notably don't have any prepared metadata.

We can start by creating a command that does a simple copy (like the scripts do) and then iterate on merging the logic between the package mode and the image mode use case.

See change and current sample/basic installation scripts:
- https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1
- https://src.fedoraproject.org/rpms/grub2/pull-request/136
- https://src.fedoraproject.org/rpms/shim/pull-request/4

See related issue: https://github.com/coreos/bootupd/issues/926

PR #1021, opened 2025-11-04
Create "direct" install sub command that will be called in a posttrans on package mode installations

yasminvalim:
TICKET: As part of https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1, for package mode installations, we will update the bootloader (copy from /usr to boot or ESP) as part of the posttrans scriptlet.

This is a slightly different path as the current update path where we have the systemd-run logic, and we notably don't have any prepared metadata.

We can start by creating a command that does a simple copy (like the scripts do) and then iterate on merging the logic between the package mode and the image mode use case.

See change and current sample/basic installation scripts:

https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1
https://src.fedoraproject.org/rpms/grub2/pull-request/136
https://src.fedoraproject.org/rpms/shim/pull-request/4

See related issue: https://github.com/coreos/bootupd/issues/926

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,15 @@ jobs:
         run: |
           set -eux
           echo 'deb [trusted=yes] https://ftp.debian.org/debian/ testing main' | sudo tee /etc/apt/sources.list.d/testing.list
+          sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 6ED0E7B82643E131 78DBA3BC47EF2265
           sudo apt update
+          DIVERSION_PATHS="lib32 libo32 lib64"
+          for PATH_NAME in $DIVERSION_PATHS; do
+            sudo dpkg-divert --remove --package base-files --divert /$PATH_NAME.usr-is-merged /$PATH_NAME || true
+            sudo dpkg-divert --remove --package base-files --divert /.$PATH_NAME.usr-is-merged /$PATH_NAME || true
+          done
+          sudo dpkg --configure -a
+          sudo apt --fix-broken install
           sudo apt install -y crun/testing podman/testing skopeo/testing
 
       - uses: actions/checkout@v5
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -36,8 +36,8 @@ jobs:
             set -xeu
             apt update -y
             apt install -y gcc make curl libssl-dev pkg-config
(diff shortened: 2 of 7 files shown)


HuijingHei:
Thank you @yasminvalim for working on this!
This is tracked by https://github.com/coreos/bootupd/issues/927, there are many dup codes in `install()`, how about make them into one? WDYT?

HuijingHei:
> I've added some unit tests for this feature, but I'm not satisfied with the coverage. Are kola tests recommended in this case?

Since this is for package mode, do we want them also working on CoreOS?

src/efi.rs
```diff
@@ -183,6 +183,95 @@ impl Efi {
...
+
+        // Sync filesystem
+        let efidir =
+            openat::Dir::open(&esp_path.join("EFI")).context("Opening EFI directory for sync")?;
+        fsfreeze_thaw_cycle(efidir.open_file(".")?)?;
+
+        Ok(())
+    }
+
+    /// Shared helper to copy EFI components to all ESP devices
```
travier: As far as I know, package mode installation from Anaconda never have multiple ESP, this is a CoreOS specific thing.

yasminvalim: You're right, I was missing that context. Already fixed it. Thanks for the review. :)

src/component.rs
```diff
@@ -79,6 +79,11 @@ pub(crate) trait Component {
 
     /// Locating efi vendor dir
     fn get_efi_vendor(&self, sysroot: &Path) -> Result<Option<String>>;
+
+    /// Copy from /usr/lib/efi to boot/ESP.
+    fn package_mode_copy_to_boot(&self) -> Result<()> {
+        Ok(())
+    }
```
HuijingHei, line 86: Should be `fn package_mode_copy_to_boot(&self) -> Result<()>`.

yasminvalim: Could you clarify what you'd like changed? The function already has that same signature. Thanks!

HuijingHei: Sorry that I misunderstood that this is `trait` definition that should not have `Ok(())`, but seems it is also OK, so leave it as it is.

yasminvalim:
Sorry folks, I encountered some issues with the git history on my previous branch, so I've pushed this new branch to resolve it. https://github.com/coreos/bootupd/pull/1030

PR #1028, opened 2025-11-18
grubconfig: should use `/sysroot` for the physical root

HuijingHei:
Get hints from bootc code https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/install.rs#L2017


HuijingHei:
Maybe we do not need it, I will close it as I did not hit error.

PR #1030, opened 2025-11-20
Create "direct" install sub command that will be called in a posttrans on package mode installations

yasminvalim:
TICKET: As part of https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1, for package mode installations, we will update the bootloader (copy from /usr to boot or ESP) as part of the posttrans scriptlet.

This is a slightly different path as the current update path where we have the systemd-run logic, and we notably don't have any prepared metadata.

We can start by creating a command that does a simple copy (like the scripts do) and then iterate on merging the logic between the package mode and the image mode use case.

See change and current sample/basic installation scripts:

https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1
https://src.fedoraproject.org/rpms/grub2/pull-request/136
https://src.fedoraproject.org/rpms/shim/pull-request/4

See related issue: https://github.com/coreos/bootupd/issues/926

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -802,6 +802,22 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
     Ok(())
 }
 
+/// Copy bootloader files from /usr/lib/efi to boot/ESP for package mode installations.
+pub(crate) fn copy_to_boot(root: &Path) -> Result<()> {
+    let all_components = get_components_impl(false);
+    if all_components.is_empty() {
+        anyhow::bail!("No components available for this platform.");
+    }
+
+    for component in all_components.values() {
+        component
+            .package_mode_copy_to_boot(root)
+            .with_context(|| format!("Failed to copy component {} to boot", component.name()))?;
+    }
+
+    Ok(())
+}
+
 /// Writes a stripped GRUB config to `stripped_config_name`, removing lines between
 /// `### BEGIN /etc/grub.d/15_ostree ###` and `### END /etc/grub.d/15_ostree ###`.
 fn strip_grub_config_file(
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -802,7 +802,9 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
     Ok(())
 }
 
-/// Copy bootloader files from /usr/lib/efi to boot/ESP for package mode installations.
+/// Copy EFI payloads from `/usr/lib/efi` onto the ESP for package-mode installs. Removals are
+/// scoped to `EFI/<vendor>/` trees present in this merge; other top-level ESP dirs are untouched.
+/// BIOS components are no-ops.
 pub(crate) fn copy_to_boot(root: &Path) -> Result<()> {
     let all_components = get_components_impl(false);
     if all_components.is_empty() {
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -802,9 +802,7 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
     Ok(())
 }
 
-/// Copy EFI payloads from `/usr/lib/efi` onto the ESP for package-mode installs. Removals are
-/// scoped to `EFI/<vendor>/` trees present in this merge; other top-level ESP dirs are untouched.
-/// BIOS components are no-ops.
+/// Copy EFI payloads from `/usr/lib/efi` onto the ESP for package-mode installs.
 pub(crate) fn copy_to_boot(root: &Path) -> Result<()> {
     let all_components = get_components_impl(false);
     if all_components.is_empty() {
--- a/src/cli/bootupd.rs
+++ b/src/cli/bootupd.rs
@@ -38,6 +38,11 @@ pub enum DVerb {
     GenerateUpdateMetadata(GenerateOpts),
     #[clap(name = "install", about = "Install components")]
     Install(InstallOpts),
+    #[clap(
+        name = "copy-to-boot",
+        about = "Copy bootloader files from /usr/lib/efi to ESP (package mode), EFI only"
+    )]
+    CopyToBoot,
 }
 
 #[derive(Debug, Parser)]
@@ -97,6 +102,7 @@ impl DCommand {
         match self.cmd {
             DVerb::Install(opts) => Self::run_install(opts),
             DVerb::GenerateUpdateMetadata(opts) => Self::run_generate_meta(opts),
+            DVerb::CopyToBoot => Self::run_copy_to_boot(),
         }
     }
 
@@ -144,6 +150,11 @@ impl DCommand {
             opts.auto,
         )
         .context("boot data installation failed")?;
+        Ok(())
+    }
+
+    pub(crate) fn run_copy_to_boot() -> Result<()> {
+        bootupd::copy_to_boot(std::path::Path::new("/")).context("copying to boot failed")?;
         Ok(())
     }
 }
--- a/src/component.rs
+++ b/src/component.rs
@@ -86,7 +86,7 @@ pub(crate) trait Component {
     /// Locating efi vendor dir
     fn get_efi_vendor(&self, sysroot: &Path) -> Result<Option<String>>;
 
-    /// Copy from /usr/lib/efi to boot/ESP (package mode)
+    /// Merge `/usr/lib/efi` onto the ESP for package-mode systems (EFI components only).
     fn package_mode_copy_to_boot(&self, root: &Path) -> Result<()>;
 }
 
--- a/src/component.rs
+++ b/src/component.rs
@@ -85,6 +85,9 @@ pub(crate) trait Component {
 
     /// Locating efi vendor dir
     fn get_efi_vendor(&self, sysroot: &Path) -> Result<Option<String>>;
+
+    /// Copy from /usr/lib/efi to boot/ESP (package mode)
+    fn package_mode_copy_to_boot(&self, root: &Path) -> Result<()>;
 }
 
 /// Given a component name, create an implementation.
--- a/src/component.rs
+++ b/src/component.rs
@@ -87,7 +87,6 @@ pub(crate) trait Component {
     fn get_efi_vendor(&self, sysroot: &Path) -> Result<Option<String>>;
 
     /// Merge `/usr/lib/efi` onto the ESP for package-mode systems (EFI only). Deletes only under
-    /// `EFI/<vendor>/` for vendors shipped in this merge; other ESP subtrees are left alone.
     fn package_mode_copy_to_boot(&self, root: &Path) -> Result<()>;
 }
 
--- a/src/component.rs
+++ b/src/component.rs
@@ -86,7 +86,8 @@ pub(crate) trait Component {
     /// Locating efi vendor dir
     fn get_efi_vendor(&self, sysroot: &Path) -> Result<Option<String>>;
 
-    /// Merge `/usr/lib/efi` onto the ESP for package-mode systems (EFI components only).
+    /// Merge `/usr/lib/efi` onto the ESP for package-mode systems (EFI only). Deletes only under
+    /// `EFI/<vendor>/` for vendors shipped in this merge; other ESP subtrees are left alone.
     fn package_mode_copy_to_boot(&self, root: &Path) -> Result<()>;
 }
 
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -854,7 +854,7 @@ impl Component for Efi {
         }
     }
 
-    /// Package mode copy: Simple copy from /usr/lib/efi to boot/ESP.
+    /// Package mode: merge `/usr/lib/efi` onto the ESP (write alongside, then atomic rename).
     fn package_mode_copy_to_boot(&self, root: &Path) -> Result<()> {
         self.package_mode_copy_to_boot_impl(root)
     }
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -19,6 +19,10 @@ use chrono::prelude::*;
 use fn_error_context::context;
 use openat_ext::OpenatDirExt;
 use os_release::OsRelease;
+use rustix::mount::{
+    fsconfig_create, fsconfig_set_path, fsmount, fsopen, move_mount, FsMountFlags, FsOpenFlags,
+    MountAttrFlags, MoveMountFlags, UnmountFlags,
+};
 use rustix::{fd::AsFd, fd::BorrowedFd, fs::StatVfsMountFlags};
 use walkdir::WalkDir;
 use widestring::U16CString;
@@ -87,9 +91,40 @@ pub(crate) fn is_efi_booted() -> Result<bool> {
         .map_err(Into::into)
 }
 
+fn mount_esp(esp_device: &Path, target: &Path) -> Result<()> {
+    use rustix::fs::CWD;
+
+    let fs_fd = fsopen("vfat", FsOpenFlags::empty()).context("fsopen vfat")?;
+    fsconfig_set_path(fs_fd.as_fd(), "source", esp_device, CWD)
+        .context("fsconfig_set_path source")?;
+    fsconfig_create(fs_fd.as_fd()).context("fsconfig_create")?;
+    let mount_fd = fsmount(
+        fs_fd.as_fd(),
+        FsMountFlags::empty(),
+        MountAttrFlags::empty(),
+    )
+    .context("fsmount")?;
+    let target_dir = std::fs::File::open(target).context("open target dir for move_mount")?;
+    let target_fd = unsafe { BorrowedFd::borrow_raw(target_dir.as_raw_fd()) };
+    move_mount(
+        mount_fd.as_fd(),
+        "",
+        target_fd,
+        ".",
+        MoveMountFlags::MOVE_MOUNT_F_EMPTY_PATH,
+    )
+    .context("move_mount")?;
+    Ok(())
+}
+
+struct Mount {
+    path: PathBuf,
+    owned: bool,
+}
+
 #[derive(Default)]
 pub(crate) struct Efi {
-    mountpoint: RefCell<Option<PathBuf>>,
+    mountpoint: RefCell<Option<Mount>>,
 }
 
 impl Efi {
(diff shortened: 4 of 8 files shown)


HuijingHei:
> [#1021 (comment)](https://github.com/coreos/bootupd/pull/1021#issuecomment-3550193095):
> 
> > @travier the command that we want to use is for package mode, also for image mode, is this right?
> 
> This command will be used exclusively on package mode.

Thank you for the confirmation, then it would be easy for us to add such tmt test.

src/cli/bootupd.rs
```diff
@@ -35,6 +35,11 @@ pub enum DVerb {
     GenerateUpdateMetadata(GenerateOpts),
     #[clap(name = "install", about = "Install components")]
     Install(InstallOpts),
+    #[clap(
+        name = "copy-to-boot",
+        about = "Copy bootloader files from /usr/lib/efi to boot/ESP (package mode)"
+    )]
+    CopyToBoot,
```
HuijingHei, line 45: Look at more about this, this means to use like `bootupctl backend copy-to-boot` from root `/`, is this right? How about if we want to use like copying from specified directory, so I think maybe it would better to be under `install`, but need to confirm with @travier , WDYT?

src/efi.rs
```diff
@@ -1129,4 +1126,101 @@ Boot0003* test";
...
+
+    #[test]
+    fn test_package_mode_shim_installation() -> Result<()> {
+        // Test that shim can be installed from /usr/lib/efi to ESP
+        let tmpdir: &tempfile::TempDir = &tempfile::tempdir()?;
+        let tpath = tmpdir.path();
+
+        // Create mock /usr/lib/efi structure with shim
+        let efi_path = tpath.join("usr/lib/efi");
+        let shim_path = efi_path.join("shim/15.8-3/EFI/fedora");
```
ravanelli, line 1284: You are harding coding the versions here, and also in line 1168 and in the other commits also for Grub. Is there any reason? it will fail if we have a diff version.

yasminvalim: Nice catch, it makes sense to avoid hardcoding there. I’ve fixed it to match the pattern used in the other tests.

src/cli/bootupd.rs
HuijingHei: Overall LGTM, just one above small question, any pointer for this? @travier

travier:
I have not reviewed the code yet, but in the meantime, can you test that this change works when you run it as part of a GRUB update? See `%posttrans` in the GRUB package: https://src.fedoraproject.org/rpms/grub2/blob/rawhide/f/grub2.spec#_430

yasminvalim:
Hey  @HuijingHei, I can take a look on how do TMT tests and apply it here. Do you have any examples of this kind of test?

Thanks @travier for your support!

src/cli/bootupd.rs
HuijingHei: Confirm with @travier , this looks good.

HuijingHei:
@yasminvalim I am doing tests with [add-post](https://github.com/coreos/bootupd/compare/main...HuijingHei:bootupd:add-post), also rebase this PR, build package for f44, and install it on package mode, failed when running posttrans, could you help to check? Thank you!
```
[root@citest-1 ~]# dnf install -y /mnt/shared/bootupd/bootupd-0.2.32.23.gd2b568e-1.fc44.x86_64.rpm
Updating and loading repositories:
Repositories loaded.
Package                                                   Arch          Version                                                    Repository                            Size
Installing:
 bootupd                                                  x86_64        0:0.2.32.23.gd2b568e-1.fc44                                @commandline                       3.5 MiB

Transaction Summary:
 Installing:         1 package

Total size of inbound packages is 1 MiB. Need to download 0 B.
After this operation, 4 MiB extra will be used (install 4 MiB, remove 0 B).
Running transaction
[1/3] Verify package files                                                                                                           100% |  76.0   B/s |   1.0   B |  00m00s
[2/3] Prepare transaction                                                                                                            100% |  16.0   B/s |   1.0   B |  00m00s
[3/3] Installing bootupd-0:0.2.32.23.gd2b568e-1.fc44.x86_64                                                                          100% |   5.4 MiB/s |   3.5 MiB |  00m01s
>>> Running %posttrans scriptlet: bootupd-0:0.2.32.23.gd2b568e-1.fc44.x86_64
>>> Finished %posttrans scriptlet: bootupd-0:0.2.32.23.gd2b568e-1.fc44.x86_64
>>> Scriptlet output:
>>> error: unrecognized subcommand 'copy-to-boot'
>>> 
>>> Usage: bootupctl backend [OPTIONS]
>>> 
>>> For more information, try '--help'.
>>> 
Warning: skipped OpenPGP checks for 1 package from repository: @commandline
Complete!
```

travier:
> Just one small question about this, after install the new bootupctl, `/boot/efi` is unmounted, maybe we should not unmount (the same as before installation)? WDYT? @travier

Ah yes, when we run that in a systemd service unit we can mount/unmount things as we are running in our own mount namespace but when running as part of an RPM transaction we should not touch mount points.

HuijingHei (approved):

src/cli/bootupd.rs
```diff
@@ -35,6 +35,11 @@ pub enum DVerb {
     GenerateUpdateMetadata(GenerateOpts),
     #[clap(name = "install", about = "Install components")]
     Install(InstallOpts),
+    #[clap(
+        name = "copy-to-boot",
+        about = "Copy bootloader files from /usr/lib/efi to boot/ESP (package mode)"
```
cgwalters: But are we trying to handle e.g. GRUB BIOS here too?

I guess this is really a question for the Change

src/bootupd.rs
```diff
@@ -688,6 +688,23 @@ pub(crate) fn client_run_migrate_static_grub_config() -> Result<()> {
     Ok(())
 }
 
+/// Copy bootloader files from /usr/lib/efi to boot/ESP for package mode installations.
+pub(crate) fn copy_to_boot() -> Result<()> {
+    let all_components = get_components_impl(false);
+    if all_components.is_empty() {
+        println!("No components available for this platform.");
```
cgwalters: Hmm this seems unlikely to happen? I wonder if we should error.

src/component.rs
```diff
@@ -79,6 +79,11 @@ pub(crate) trait Component {
 
     /// Locating efi vendor dir
     fn get_efi_vendor(&self, sysroot: &Path) -> Result<Option<String>>;
+
+    /// Copy from /usr/lib/efi to boot/ESP.
+    fn package_mode_copy_to_boot(&self) -> Result<()> {
+        Ok(())
```
cgwalters: I think we shouldn't have a default impl here but make each component implement it. That would make it more obvious that grub is a no-op.

src/efi.rs
```diff
@@ -68,6 +68,8 @@ pub(crate) fn is_efi_booted() -> Result<bool> {
 #[derive(Default)]
 pub(crate) struct Efi {
     mountpoint: RefCell<Option<PathBuf>>,
+    /// Track whether we mounted the ESP ourselves (true) or found it pre-mounted (false)
+    did_mount: RefCell<bool>,
```
cgwalters: Hmm seems cleaner to have one RefCell i.e.

```
struct Mount {
  path: PathBuf,
  owned: bool
}

struct Efi {
  mountpoint: RefCell<Option<Mount>>
```

That said even cleaner would be for us to use the new mount API internally and not even affect the global mountns at all

src/efi.rs
```diff
@@ -183,6 +191,89 @@ impl Efi {
...
+                .with_context(|| {
+                    format!(
+                        "Failed to copy {} from {} to {}",
+                        efi_comp.name, efi_comp.path, dest_str
+                    )
+                })?;
+        }
+
+        // Sync filesystem
+        let efidir =
```
cgwalters: I think we want to sync the whole ESP not just the EFI subdirectory - there's not a big different for most use cases.

Also it'd be cleaner if we accepted the ESP as a Dir not a path.

src/efi.rs
```diff
@@ -183,6 +191,89 @@ impl Efi {
...
+            openat::Dir::open(&esp_path.join("EFI")).context("Opening EFI directory for sync")?;
+        fsfreeze_thaw_cycle(efidir.open_file(".")?)?;
+
+        Ok(())
+    }
+
+    /// Copy from /usr/lib/efi to boot/ESP.
+    fn package_mode_copy_to_boot_impl(&self) -> Result<()> {
+        let sysroot = Path::new("/");
```
cgwalters: Why not require the caller to provide this? I can imagine use cases where we want this to operate on a non-mounted root like system recovery.

Among other things...it makes unit tests work.

src/efi.rs
```diff
@@ -183,6 +191,89 @@ impl Efi {
...
+        let sysroot_path =
+            Utf8Path::from_path(sysroot).context("Sysroot path is not valid UTF-8")?;
+
+        let efi_comps = match get_efi_component_from_usr(sysroot_path, EFILIB)? {
+            Some(comps) if !comps.is_empty() => comps,
+            _ => {
+                log::debug!("No EFI components found in /usr/lib/efi");
```
cgwalters: I think there's an open question of if this should be configurable as an error or not.

I think it probably should be.

src/efi.rs
```diff
@@ -183,6 +191,89 @@ impl Efi {
...
+                return Ok(());
+            }
+        };
+
+        let sysroot_dir = openat::Dir::open(sysroot).context("Opening sysroot for reading")?;
```
cgwalters: Let's move this up to be closer to where the path is used.

src/efi.rs
```diff
@@ -414,23 +505,22 @@ impl Component for Efi {
...
+            self.copy_efi_components_to_esp(&src_dir, &destpath, &efi_components)?;
             EFILIB
         } else {
             let updates = component_updatedirname(self);
             let src = updates
                 .to_str()
                 .context("Include invalid UTF-8 characters in path")?;
+            let dest = destpath.to_str().with_context(|| {
+                format!(
+                    "Include invalid UTF-8 characters in dest {}",
```
cgwalters: I would say "Invalid UTF-8: {}" destpath.display()

src/efi.rs
```diff
@@ -948,7 +1042,7 @@ Boot0003* test";
         {
             tmpd.atomic_write(
                 "etc/system-release",
-                "Red Hat Enterprise Linux CoreOS release 4
+                r"Red Hat Enterprise Linux CoreOS release 4
```
cgwalters, line 1182: Why out of curiosity?

src/cli/bootupd.rs
HuijingHei: No, I think we only want to work on EFI, does this make sense to add some info as `Copy bootloader files from /usr/lib/efi to ESP (package mode), only for EFI`?

src/efi.rs
HuijingHei: > Hmm seems cleaner to have one RefCell i.e.
> 
> ```
> struct Mount {
>   path: PathBuf,
>   owned: bool
> }
> 
> struct Efi {
>   mountpoint: RefCell<Option<Mount>>
> ```
LGTM.

> That said even cleaner would be for us to use the new mount API internally

I think you meant for https://github.com/bootc-dev/bootc/blob/main/crates/mount/Cargo.toml which is using rust `edition = "2024"`, we should use the new mount once move to 2024 (now using `edition = "2021"`)

cgwalters: "new mount api" is https://brauner.io/2023/02/28/mounting-into-mount-namespaces.html there's some usage of it in bootc

src/bootupd.rs
yasminvalim: Should we keep it? I added a error when there are no components. WDYT?

src/component.rs
yasminvalim: Done. No default anymore. WDYT?

src/efi.rs
yasminvalim: RefCell is now a single Mount { path, owned }, and we use the new mount API in mount_esp(). Could you take a look when you have a moment?

src/efi.rs
yasminvalim: I applied this suggestions but not sure if its okay. Feel free to take a look.

src/efi.rs
yasminvalim: The test is just to ensure the parser will return the same name even when the file ends with a newline. It makes sense for you? If you think is not a good test scenario let me now so I can delete it :)

Thanks for the review and the comments :)

src/efi.rs
yasminvalim: I am not sure if I understood. Do you mean we should make “no components” configurable (e.g. strict vs allow via a flag)? If yes, I can add a CLI option for it.

src/efi.rs
```diff
@@ -108,8 +145,16 @@ impl Efi {
             if !mnt.exists() {
                 continue;
             }
+            #[cfg(target_os = "linux")]
```
HuijingHei: I do not remember we support target_os is not linux, maybe we should remove that?

yasminvalim: Hey Huijing, I removed this one in line 148. Should remove all _#[cfg(target_os = "linux")]_?

HuijingHei: Yes, I think so.

HuijingHei:
There are a lot of changes since https://github.com/coreos/bootupd/pull/1068 merged, let us get CI and test [PR](https://github.com/coreos/bootupd/pull/1003) passed firstly. After that, would ask for help from Colin or Timothee to have a final review, WDYT?

src/efi.rs
```diff
@@ -87,9 +91,40 @@ pub(crate) fn is_efi_booted() -> Result<bool> {
         .map_err(Into::into)
 }
 
+fn mount_esp(esp_device: &Path, target: &Path) -> Result<()> {
```
cgwalters: This is basically just the `mount` system call as written I think, and we already have ESP mounting code, it's not clear to me why we need a new mounting functionality here?

That said, one thing that we *could* do here by using the new mount api (instead of raw `mount`) is to skip the `move_mount` part, and then just access everything via `mount_fd` (which we wrap in a `cap-std::fs::Dir`) or so.

src/efi.rs
```diff
@@ -180,12 +225,25 @@ impl Efi {
...
+            if let Some(mount) = self.mountpoint.borrow_mut().take() {
+                if rustix::mount::unmount(&mount.path, UnmountFlags::empty()).is_ok() {
+                    log::trace!("Unmounted (new mount API)");
+                } else {
+                    Command::new("umount")
+                        .arg(&mount.path)
+                        .run_inherited()
+                        .with_context(|| format!("Failed to unmount {:?}", mount.path))?;
+                    log::trace!("Unmounted");
+                }
```
cgwalters: I think `umount` should work, if it doesn't, forking off the external binary isn't more likely to work. Can you motivate this change? I think we should stick with one or the other.

If we're hitting issues because it's still in use, one option is to do a lazy unmount. But if it is in use it's likely because we still hold a file descriptor open, which we should have closed.

src/util.rs
```diff
@@ -51,9 +53,17 @@ pub(crate) fn filenames(dir: &openat::Dir) -> Result<HashSet<String>> {
     Ok(ret)
 }
 
+/// Return the available space in bytes on the filesystem containing the given directory.
+/// Uses f_bavail * f_frsize from fstatvfs to avoid partial updates when the partition is full.
+pub(crate) fn available_space_bytes(dir: &openat::Dir) -> Result<u64> {
+    let fd = unsafe { BorrowedFd::borrow_raw(dir.as_raw_fd()) };
```
cgwalters: I think this can just use `.as_fd()`

src/efi.rs
```diff
@@ -236,6 +294,119 @@ impl Efi {
...
+    fn copy_efi_components_to_esp(
+        &self,
+        sysroot_dir: &openat::Dir,
+        esp_dir: &openat::Dir,
+        _esp_path: &Path,
+        efi_components: &[EFIComponent],
+    ) -> Result<()> {
+        // Build a merged source tree in a temp dir (same layout as desired ESP/EFI)
+        let temp_dir = tempfile::tempdir().context("Creating temp dir for EFI merge")?;
+        let temp_efi_path = temp_dir.path().join("EFI");
```
cgwalters: It's a bit unfortunate copy all the files into a tempdir in a different place, then copy them again.

src/efi.rs
```diff
@@ -236,6 +294,119 @@ impl Efi {
...
+            .context("Temp EFI path is not valid UTF-8")?;
+
+        for efi_comp in efi_components {
+            log::info!(
+                "Merging EFI component {} version {} into update tree",
+                efi_comp.name,
+                efi_comp.version
+            );
+            // Copy contents of component's EFI dir (e.g. fedora/) into temp_efi_path so merged
+            // layout is EFI/fedora/..., not EFI/EFI/fedora/...
```
cgwalters: But, we could adjust how we do the copy so it's the right path to start.

src/efi.rs
```diff
@@ -236,6 +294,119 @@ impl Efi {
...
+        let source_dir =
+            openat::Dir::open(&temp_efi_path).context("Opening merged EFI source dir")?;
+        let source_filetree =
+            filetree::FileTree::new_from_dir(&source_dir).context("Building source filetree")?;
+        let current_filetree =
+            filetree::FileTree::new_from_dir(&esp_efi_dir).context("Building current filetree")?;
+        let mut diff = current_filetree
+            .diff(&source_filetree)
+            .context("Computing EFI diff")?;
+        diff.removals.clear();
```
cgwalters: Why is this? Won't this mean we keep leaking old files?

src/efi.rs
```diff
@@ -686,6 +853,11 @@ impl Component for Efi {
             anyhow::bail!("Failed to find {SHIM} in the image")
         }
     }
+
+    /// Package mode: merge `/usr/lib/efi` onto the ESP (write alongside, then atomic rename).
+    fn package_mode_copy_to_boot(&self, root: &Path) -> Result<()> {
+        self.package_mode_copy_to_boot_impl(root)
+    }
```
cgwalters, line 857: Why the indirection here?

src/util.rs
yasminvalim: openat::Dir is only AsRawFd, not AsFd, so I added a wrapper that implements AsFd and used `fstatvfs(wrapper(dir).as_fd())`. Is that okay or should I avoid using a wrappper? Is that any other way that I could implement it?

travier:
Let's put this work on hold for now as we don't have an agreement with the bootloader team for now that this is the path forward for package mode updates.

PR #1039, opened 2025-12-17
Update will fail if update metadata is missing for the current boot type

### LINKED ISSUE #1035: Fail updates if metadata is missing in the image for current boot type
I.e. if we are EFI booted and there is no metadata (EFI.json), then bootupd should fail the update as this is unexpected.

See: https://github.com/coreos/bootupd/pull/1018

HuijingHei:
Fixes https://github.com/coreos/bootupd/issues/1035

### DIFF
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -243,7 +243,12 @@ impl Component for Bios {
     }
 
     fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>> {
-        get_component_update(sysroot, self)
+        let content_metadata = get_component_update(sysroot, self)?;
+        // Failed as expected if booted with BIOS and no update metadata
+        if content_metadata.is_none() && !sysroot.exists("sys/firmware/efi")? {
+            anyhow::bail!("Failed to find BIOS update metadata");
+        }
+        Ok(content_metadata)
     }
 
     fn run_update(&self, rootcxt: &RootContext, _: &InstalledContent) -> Result<InstalledContent> {
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -67,7 +67,7 @@ cosaPod(memory: "7168Mi", cpu: "4") {
       // The previous e2e leaves things only having built an ostree update
       shwrap("cosa build && cosa osbuild qemu")
       // bootupd really can't break upgrades for the OS
-      kola(cosaDir: "${env.WORKSPACE}", extraArgs: "ext.*bootupd*", skipUpgrade: true, skipBasicScenarios: true)
+      kola(cosaDir: "${env.WORKSPACE}", extraArgs: "ext.*bootupd* --qemu-firmware=uefi", skipUpgrade: true, skipBasicScenarios: true)
     }
   } finally {
     archiveArtifacts allowEmptyArchive: true, artifacts: 'tmp/console.txt'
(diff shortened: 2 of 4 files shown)


src/bios.rs
```diff
@@ -243,7 +243,12 @@ impl Component for Bios {
     }
 
     fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>> {
-        get_component_update(sysroot, self)
+        let content_metadata = get_component_update(sysroot, self)?;
+        // Failed as expected if booted with BIOS and no update metadata
+        if content_metadata.is_none() && !sysroot.exists("sys/firmware/efi")? {
```
cgwalters, line 248: why not `!is_efi_booted` instead of testing the `/sys` path?

cgwalters (approved):

src/bios.rs
HuijingHei: I am concern that it might be not suitable for ppc, as `is_efi_booted()` is for `x86_64`, or maybe it is better to move `is_efi_booted()` out of efi.rs. WDYT?

alexlarsson:
This change is breaking the builds for automotive.

 I'm getting this when using image-builder converting the bootc image to a disk image:
```
org.osbuild.bootc.install-to-filesystem: a5b26368688ef424171b4a5692348bcd81f254a8a3646ac0abcad254816b27a0 {
  "kernel-args": [
    "rw",
    "console=tty0",
    "console=ttyS0"
  ],
  "target-imgref": "localhost/aib-c21c2b43c83771df5115a465"
}
device/disk (org.osbuild.loopback): loop0 acquired (locked: False)
mount/- (org.osbuild.ext4): mounting /dev/loop0p6 -> /var/tmp/aib-kxl5o3kf/cache/tmp/buildroot-tmp-g4nf_o14/mounts/
mount/boot-efi (org.osbuild.fat): mounting /dev/loop0p1 -> /var/tmp/aib-kxl5o3kf/cache/tmp/buildroot-tmp-g4nf_o14/mounts/boot/efi
Installing image: docker://localhost/aib-c21c2b43c83771df5115a465
Initializing ostree layout
layers already present: 0; layers needed: 65 (372.0Â MB)
Deploying container image...done (7 seconds)
Bootloader: grub
Installing bootloader via bootupd
error: boot data installation failed: Failed to find BIOS update metadata
```

I don't see this on my machine which has a /sys/firmware/efi directory, but I am seeing it in the internal CI builds, which are probably running on a non-efi system.

alexlarsson:
Overall, its probably not good that bootupd makes decisions based on host hardware details, as such things will never be right when running in the context of generating images.

travier:
This was not meant to be triggered at install time but only at update time, so this is a bug.

alexlarsson:
Then it should be in run_update(), not query_update(), i think. query_update() is called from install().

travier:
The main concern that led to this PR was that we would remove the EFI binaries if the metadata was not there in the image and that would break an installed system on update: https://github.com/coreos/bootupd/pull/1018.

But this only matters for the EFI case as we do not "remove" the BIOS boot partition content on BIOS booted systems on update if there is no BIOS component. We just do not update it.

So I would remove the checks for BIOS and only keep the runtime check that if the system is booted via EFI, we should install an EFI component. This would break systems where we would only install BIOS but that does not exists?

alexlarsson:
@travier I don't think that is enough anyway. Consider building an image targeting android boot (no bios, no efi), this would have an empty efi component,  and we would fail to build such an image on an efi booted build machine. I think we need the full https://github.com/coreos/bootupd/pull/1096 fix

PR #1046, opened 2026-01-06
ci: cleanup; drop running COSA as UID 0

dustymabe:


cgwalters:
Thanks for looking at this! We have the same issue over in ostree.

The next failure looks like it's because we're trying to install bootupd inside the cosa container. But I think we should now always use bootupd from the target system, so we could probably drop that?

Or it would likely work to just add a `sudo` there?

HuijingHei (approved):

PR #1063, opened 2026-02-10
grubconfigs: ensure `bootuuid.cfg` and `grubenv` with mode 0600

### LINKED ISSUE #1058: Ensure bootupd-generated files have CIS-compliant permissions
This is a follow-up to issue #952.

The CIS compliance framework requires that all files in `/boot/grub2/` have a permission mode of `0600` or more restrictive. Currently, `/boot/grub2/bootuuid.cfg` and `/boot/grub2/grubenv` are created with mode `0644`. This causes [failures during compliance scans on RHEL Image Mode systems](https://github.com/ComplianceAsCode/content/issues/14254).

Would it be possible to set the permissions for these files to `0600` as well?

HuijingHei:
Fixes https://github.com/coreos/bootupd/issues/1058

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,10 @@ jobs:
           root_part=$(sudo sfdisk -l -J "${device}" | jq -r '.partitiontable.partitions[] | select(.name == "root").node')
           sudo mount "${root_part}" /mnt/
           sudo ls /mnt/boot/grub2/grub.cfg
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
Looks sane, but we probably want to also apply this fixup automatically on upgrades?

HuijingHei:
> Looks sane, but we probably want to also apply this fixup automatically on upgrades?

Good catch, will fix on upgrades later.

PR #1006, opened 2025-09-15
efi: Get the correct esp on multipath when install

HuijingHei:
efi: Check shim exists and return earlier if not found

---

efi: Get the correct esp on multipath

Get pointer from Colin:
```
what we need to do here instead is "peel" through the multipath 
to find the block device(s) backing it, and then pick one of those.
```
When I do testing using mpath or backing device, find the 
results are the same, so I use mpath instead.

As there is no partition attribute on multipath env, this leads
error when install on multipath.
Fixes: https://issues.redhat.com/browse/RHEL-81039

PR #1068, opened 2026-03-05
Add support for installing to devices with multiple parents

ckyrouac:
See individual commits for details. Generally, this migrates to using the new blockdev crate funtions and standardize on the Device struct operating on block devices. On top of this, handle the `--filesystem` and multiple `--device` flags to install the bootloader to multiple devices, e.g. a /dev/md0 RAID1 device backed by 3 physical disks, each with an ESP partition.

Keeping this as a draft because it will break older bootc versions prior to https://github.com/bootc-dev/bootc/pull/2048.

Leaving the TEMP code in to validate CI passes (previously validated on my fork).


src/efi.rs
```diff
@@ -83,6 +86,18 @@ impl Efi {
 
             let st = rustix::fs::statfs(&path)?;
             if st.f_type == libc::MSDOS_SUPER_MAGIC {
+                // Verify this is actually a mount point, not just a subdirectory
```
cgwalters, line 108: It'd be cleaner to use https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.is_mountpoint but yes that gets into the problem that we haven't ported this project to cap-std overall, but we can still use it here too

src/efi.rs
```diff
@@ -108,6 +123,20 @@ impl Efi {
             if !mnt.exists() {
                 continue;
             }
+
+            // Check if the target is already a mounted ESP (e.g. the host
```
cgwalters, line 144: Looks nearly duplicated with above?

cgwalters (commented):
I didn't do a deep review, only superficial. There's some nontrivial regression risk here - CI covers some basics, but far from everything.

.github/workflows/workflow-test.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Compute image tag
+        id: tag
+        run: |
+          BRANCH="${GITHUB_REF_NAME}"
+          SHA="$(git rev-parse --short HEAD)"
+          TAG="${BRANCH}-${SHA}"
+          echo "tag=${TAG}" >> "$GITHUB_OUTPUT"
+
+      - name: Get a newer podman for heredoc support (from debian testing)
```
cgwalters: Check out https://github.com/ostreedev/ostree/pull/3573 which I need to get back to but the basic idea here is we use the bootc-ubuntu-setup action for stuff like this.

ckyrouac: :+1: pushed that commit up just to see if it would actually run. Looks like it also needs `/ok-to-test` from a maintainer

cgwalters: I wrote up https://gist.github.com/cgwalters/715bb8ee634f724763f609dffe624e55

ckyrouac: I tried pushing to my main branch of this repo last week and it didn't trigger the actions, I thought the reason was something specific to this repo but maybe it was just part of the flakiness that was happening with github actions last week. I'll try again.

ckyrouac:
marking as ready for review but waiting on bootc 1.14.0 to release to remove the `temp` commit and for the arm/s390/ppc builds for the tests

src/bootupd.rs
```diff
@@ -108,15 +109,67 @@ pub(crate) fn install(
...
+        // devices that have an ESP partition.
+        let devices_to_install: Vec<Option<&Device>> = if devices.is_empty() {
+            // No devices specified: install once with auto-detection (None).
+            vec![None]
+        } else if component.name() == "EFI" {
+            // For EFI, only install to devices that have an ESP partition.
+            let with_esp: Vec<Option<&Device>> = devices
+                .iter()
+                .filter_map(|dev| match dev.find_partition_of_esp() {
+                    Ok(_) => Some(Some(dev)),
```
cgwalters: `Some(Some())` is an unusual construction

src/bootupd.rs
```diff
@@ -368,14 +421,31 @@ pub(crate) fn adopt_and_update(
     }
 }
 
+/// Get the block device backing the current root by trying `/boot` first,
```
cgwalters, line 427: I think it'd be cleaner to use `/sysroot` if it exists and is a mountpoint, otherwise use `/`.

In bootc/ostree we call this "physical root" - there's similar code there.

src/efi.rs
```diff
@@ -58,6 +60,23 @@ const STUB_INFO_VAR_STR: &str = "StubInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f";
...
+    let parent = path.parent().unwrap_or(path);
+    let parent = if parent.as_os_str().is_empty() {
+        Path::new(".")
+    } else {
+        parent
+    };
+    let parent_dir = Dir::open_ambient_dir(parent, cap_std::ambient_authority())?;
+    let basename = path
+        .file_name()
+        .ok_or_else(|| anyhow::anyhow!("path has no file name: {path:?}"))?;
```
cgwalters: Some similar code in https://github.com/coreos/cap-std-ext/blob/5493d689d12512ffcd1e644935a8bee5a26bd60b/src/dirext.rs#L463 that I think is a bit more elegant. I've been tempted to make it a public API

Cargo.toml
```diff
@@ -21,9 +21,9 @@ path = "src/main.rs"
 
 [dependencies]
 anyhow = "1.0"
-bootc-internal-blockdev = "0.1.0"
+bootc-internal-blockdev = { git = "https://github.com/bootc-dev/bootc", rev="99da62b6f09ed42841396ac66298030f25daf439"}
```
cgwalters: Unfortunately this will reintroduce the problem that bootupd doesn't use vendored deps in Fedora, so we either need to publish the updated crate (i can look) or fix bootupd to use vendored deps even in Fedora.

src/bootupd.rs
```diff
@@ -108,15 +109,67 @@ pub(crate) fn install(
...
+                        device_desc,
+                    )
+                })?;
+            log::info!(
+                "Installed {} {} to {}",
+                component.name(),
+                meta.meta.version,
+                device_desc,
+            );
+            // Only record state once per component (use first device's metadata)
```
cgwalters: Hmm, I thought we were handling multi-device ESP? But maybe not...

cgwalters (approved):
Offhand this is looking plausible. This may touch on e.g. https://github.com/bootc-dev/bootc/issues/1648 and similar issues, I hope we can try to see if we've fixed them with this.

src/bootupd.rs
ckyrouac: This pattern was copied from the [removed blockdev.rs code](https://github.com/coreos/bootupd/pull/1068/changes#diff-f3271c0e43c5c97d5dad1acba522420efc01e2ac56cf1f31f3fd07b0e5574d79L13). I'm not entirely sure why it was done this way originally, but I think your suggestion makes sense.

HuijingHei: Consider that if boot and root partitions are on separate disks, see https://github.com/bootc-dev/bootc/issues/1916, maybe `/boot` makes sense?

src/bootupd.rs
ckyrouac: updated with a clearer comment

cgwalters:
FYI I put up https://github.com/coreos/bootupd/pull/1072 which may textually (but not semantically) conflict with this

ckyrouac:
updated to use the bootc rpm from copr and the 0.2.0 release of the blockdev crate, should be ready for another test run and review

ckyrouac:
@cgwalters is there a way to trigger the tests now that /ok-to-test was applied?

cgwalters:
The prow job ran and passed, so that's good, although it's not testing too much more than Jenkins is I believe.

src/bootupd.rs
cgwalters: That's a very good point! Indeed...we should support that case. So I think things will work correctly if we list `/boot` first.

That said, there's a related thing here which is: we should honor XBOOTLDR from the [DPS spec](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/) if it exists.

At the current time we aren't setting that up in the Fedora-derivative ecosystem much that I know of, but we should support it.

src/efi.rs
```diff
@@ -389,19 +443,24 @@ impl Component for Efi {
...
-        let destpath = if let Some(destdir) = self.get_mounted_esp(Path::new(dest_root))? {
+        // Determine the destination path for the ESP.
+        // If a device is provided, find and mount its ESP partition (unmounting
+        // any previously mounted ESP first to target the correct device).
+        // If no device is provided, fall back to an already-mounted ESP.
+        let destpath = if let Some(dev) = device {
+            // Unmount any previously mounted ESP to ensure we install to the
+            // correct device. This is important for multi-device installs where
+            // each device has its own ESP.
+            self.unmount()?;
```
cgwalters, line 454: I think this is OK as is, but it'd probably be cleaner longer term to have the mount clearly scoped to the device.

Dockerfile
```diff
@@ -23,6 +23,18 @@ FROM $base
...
 RUN rpm -e bootupd
 COPY --from=build /out/ /
+# Install bootc from copr
+RUN <<EORUN
+set -xeuo pipefail
+dnf -y install dnf-plugins-core
+dnf -y copr enable rhcontainerbot/bootc centos-stream-9-x86_64
+dnf -y install bootc
+dnf clean all
+rm -rf /var/log
```
cgwalters, line 33: Yeah...at some point we'll need to do https://github.com/rpm-software-management/dnf5/issues/2011#issuecomment-2609698071

cgwalters (approved):
So...I think this looks good but we should talk about tests.

>  install the bootloader to multiple devices, e.g. a /dev/md0 RAID1 device backed by 3 physical disks, each with an ESP partition.

Does Anaconda support setting this up? I think we could automate that here. Other than that, another way would be a custom bit that spawns a `bcvk ephemeral` instance w/two virtio block devices attached, partitions both disks, then uses `bootc install to-filesystem`.

ckyrouac:
> So...I think this looks good but we should talk about tests.
> 
> > install the bootloader to multiple devices, e.g. a /dev/md0 RAID1 device backed by 3 physical disks, each with an ESP partition.
> 
> Does Anaconda support setting this up? I think we could automate that here. Other than that, another way would be a custom bit that spawns a `bcvk ephemeral` instance w/two virtio block devices attached, partitions both disks, then uses `bootc install to-filesystem`.

Currently the tests for multi device are in the bootc PR, see https://github.com/bootc-dev/bootc/pull/1911/changes#diff-c8f79904b9f0bdcea90dd5a8e8cfd0c84e7f6da4a36868cf0b86421c767f452b

ckyrouac:
> The prow job ran and passed, so that's good, although it's not testing too much more than Jenkins is I believe.

The cross arch builds didn't re-run after the /ok-to-test. I'm not sure if that's blocking this from merging. The builds should pass now with the copr rpm in the test container.

cgwalters:
The composefs fix got done in a better way in https://github.com/bootc-dev/bootc/pull/2108/changes

travier:
This apparently introduces a regression on multipath setups: https://redhat.atlassian.net/browse/COS-4364

The bot points at https://github.com/coreos/bootupd/pull/1006 as the fix that is reverted by this PR. I have not invested more yet.

PR #1069, opened 2026-03-12
bootupctl: only print the available component when running status in container

HuijingHei:
Skip the component if the update meatadata is missing


HuijingHei:
@cgwalters could you help to review this when you have a chance? Thank you!

PR #1073, opened 2026-03-22
efi: Support managing non-EFI components at the ESP root

martinezjavier:
Platforms like the Raspberry Pi place firmware blobs (device-tree files, GPU firmware, config.txt) directly at the ESP root. But currently, `bootupd` can only manage the files in the `EFI` sub-directory of the ESP.

This series extends `bootupd` to manage those files alongside traditional EFI bootloaders.

It also fixes two pre-existing issues: non-deterministic version mixing when multiple component versions coexist under `/usr/lib/efi/`, and truncation of hyphenated RPM names (e.g. `bcm2711-firmware` becomes `bcm2711`) during the `ostree-boot` transfer.

The commits were developed assisted by the Cursor IDE and Claude Opus 4 model. But all the code and commit messages were manually reviewed and edited by me, ensured that the changes are properly separated in different commits, the code is kept bisectable, etc.

AI assistance is noted via `Assisted-by:` tags per the bootc agent skills [review guidelines](https://github.com/bootc-dev/agent-skills/blob/main/REVIEW.md).

End-to-end was tested on a Raspberry Pi 4 using Fedora CoreOS 43, with the `bcm2711-firmware` installed on top of the initial OSTree deployment. The complete flow has been verified:  `generate-update-metadata` , `adopt-and-update`, `validate`, and the `update` path with a simulated `bcm2711-firmware` component version bump.

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -30,6 +30,8 @@ use crate::freezethaw::fsfreeze_thaw_cycle;
 use crate::model::*;
 use crate::ostreeutil;
 use crate::util;
+use uapi_version::Version;
+
 use crate::{component::*, packagesystem::*};
 use crate::{filetree, grubconfigs};
 
@@ -79,6 +81,11 @@ fn is_mount_point(path: &Path) -> Result<bool> {
         )
     }
 }
+
+/// Copy options that merge source contents into an existing destination
+/// directory instead of creating a subdirectory (used for non-EFI
+/// components like firmware that live at the ESP root).
+const OPTIONS_MERGE: &[&str] = &["-rp", "--reflink=auto", "-T"];
 
 /// Return `true` if the system is booted via EFI
 pub(crate) fn is_efi_booted() -> Result<bool> {
@@ -236,6 +243,65 @@ impl Efi {
         let device_path = device.path();
         create_efi_boot_entry(&device_path, esp_part_num.trim(), &loader, &product_name)
     }
+
+    fn ensure_efi_prefix(&self, mut ft: filetree::FileTree) -> filetree::FileTree {
+        let needs_prefix =
+            !ft.children.is_empty() && ft.children.keys().all(|k| !k.starts_with("EFI/"));
+        if needs_prefix {
+            ft.prepend_prefix("EFI");
+        }
+        ft
+    }
+
+    /// Build a `FileTree` from either the new `usr/lib/efi/` layout (when
+    /// pre-resolved components are provided) or the legacy
+    /// `usr/lib/bootupd/updates/EFI` directory.
+    ///
+    /// For the new layout, only the latest version of each component is
+    /// included so that duplicate destination keys cannot arise.
+    fn build_filetree(
+        &self,
+        root_dir: &openat::Dir,
+        components: Option<&[EFIComponent]>,
+    ) -> Result<(PathBuf, filetree::FileTree)> {
+        if let Some(components) = components {
+            let p = PathBuf::from(EFILIB);
+            let dir = root_dir
+                .sub_dir(&p)
+                .with_context(|| format!("opening {}", p.display()))?;
+            let latest = latest_versions(components);
+            let prefixes: Vec<String> = latest
+                .iter()
+                .map(|c| format!("{}/{}", c.name, c.version))
+                .collect();
+            let ft = filetree::FileTree::new_from_dir_strip_prefix_for(&dir, &prefixes)?;
+            Ok((p, ft))
+        } else {
+            let p = component_updatedirname(self);
+            let dir = root_dir
+                .sub_dir(&p)
+                .with_context(|| format!("opening {}", p.display()))?;
+            let mut ft = filetree::FileTree::new_from_dir(&dir).context("reading update dir")?;
+            ft = self.ensure_efi_prefix(ft);
+            Ok((p, ft))
+        }
+    }
+
+    /// Build the update `FileTree`, resolving EFI components from the
+    /// sysroot and delegating to `build_filetree`.
+    fn build_update_filetree(
+        &self,
+        sysroot: &openat::Dir,
+        sysroot_path: &Utf8Path,
+    ) -> Result<(PathBuf, filetree::FileTree)> {
+        let efilib_path = sysroot_path.join(EFILIB);
+        let components = if efilib_path.exists() {
+            get_efi_component_from_usr(sysroot_path, EFILIB)?
+        } else {
+            None
+        };
+        self.build_filetree(sysroot, components.as_deref())
+    }
 }
 
 #[context("Get product name")]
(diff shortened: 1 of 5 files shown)


src/efi.rs
```diff
@@ -806,100 +860,171 @@ pub struct EFIComponent {
...
+/// component name), return only the latest version for each name.
+///
+/// Versions are compared lexicographically; this is sufficient for RPM EVR
+/// strings where the epoch:version-release ordering is consistent.
+fn latest_versions(components: &[EFIComponent]) -> Vec<&EFIComponent> {
+    let mut by_name: std::collections::HashMap<&str, &EFIComponent> =
+        std::collections::HashMap::new();
+    for c in components {
+        let entry = by_name.entry(c.name.as_str()).or_insert(c);
+        if c.version > entry.version {
```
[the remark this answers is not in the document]
martinezjavier: Done.

src/efi.rs
```diff
@@ -236,6 +241,55 @@ impl Efi {
...
+        &self,
+        sysroot: &openat::Dir,
+        sysroot_path: &Utf8Path,
+    ) -> Result<(PathBuf, filetree::FileTree)> {
+        let efilib_path = sysroot_path.join(EFILIB);
+        if let Some(components) = if efilib_path.exists() {
+            get_efi_component_from_usr(sysroot_path, EFILIB)?
+        } else {
+            None
+        } {
```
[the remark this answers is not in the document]
martinezjavier: Done.

src/efi.rs
```diff
@@ -480,27 +526,39 @@ impl Component for Efi {
...
             let updates = component_updatedirname(self);
             let src = updates
                 .to_str()
                 .context("Include invalid UTF-8 characters in path")?;
             filetree::copy_dir_with_args(&src_dir, src, dest, OPTIONS)?;
-            &src.to_owned()
+            let mut ft = crate::filetree::FileTree::new_from_dir(&src_dir.sub_dir(src)?)?;
+            ft = self.ensure_efi_prefix(ft);
+            (ft, src_path.join(src))
         };
```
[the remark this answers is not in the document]
martinezjavier: Done.

alexlarsson:
I had a quick look at that (not an expert on bootupd, so not a thorough review), but it looks good, and exactly what we need for automotive (for rpi support).

martinezjavier:
> ## Code Review
> This pull request extends `bootupd` to manage non-EFI components at the ESP root, which is useful for platforms like Raspberry Pi. It also fixes issues with version selection and RPM name handling. The changes are logical and well-tested. I've identified a potential bug in version comparison and a couple of areas for code quality improvement to enhance maintainability.

I've addressed these since the code review bot was correct about both issues.

src/filetree.rs
```diff
@@ -103,6 +103,8 @@ pub(crate) struct FileTreeDiff {
     pub(crate) additions: HashSet<String>,
     pub(crate) removals: HashSet<String>,
     pub(crate) changes: HashSet<String>,
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub(crate) source_map: HashMap<String, String>,
```
HuijingHei, line 107: Does it make sense to keep only one version instead of multiple? WDYT? @travier @cgwalters

src/filetree.rs
```diff
@@ -531,9 +537,49 @@ pub(crate) fn apply_diff(
             .with_context(|| format!("copying {:?} to {:?}", copy_src, path_tmp))?;
     }
 
+    // Sort updates to enforce stable order and atomic sequence
+    let mut update_list: Vec<_> = updates.iter().collect();
+    update_list.sort_by(|(dst_a, tmp_a), (dst_b, tmp_b)| {
```
HuijingHei, line 542: Maybe we can make this as separated function?

src/filetree.rs
HuijingHei: Create https://bugzilla.redhat.com/show_bug.cgi?id=2451626 to track the issue: upgrade shim does not remove old version direcory `/usr/lib/efi/shim/<older version>`

src/efi.rs
```diff
@@ -236,6 +243,65 @@ impl Efi {
         let device_path = device.path();
         create_efi_boot_entry(&device_path, esp_part_num.trim(), &loader, &product_name)
     }
+
+    fn ensure_efi_prefix(&self, mut ft: filetree::FileTree) -> filetree::FileTree {
```
Rolv-Apneseth, line 247: I see that this is being used to normalise the old/existing trees, but (and this may be a naive question on my part) but what happens if there's ever a case where there's only non-EFI components? Would it be possible to instead add a marker field to `InstalledContent` for new versions and just check?

src/filetree.rs
Rolv-Apneseth: I would actually suggest also pulling all the data required for sorting out of the `sort_by` loop. The algorithm is (worst-case) `O(n * log(n))` and it's best we don't fetch all this data more than once for any entry - especially since there's I/O involved. This would also solve potential issues with keeping the sort deterministic when you have calls to `.metadata` (which can fail).

Maybe something like this?

```rust
struct UpdateEntry {
    dst: Utf8PathBuf,
    tmp: String,
    is_efi: bool,
    is_dir: bool,
}

let mut update_list: Vec<UpdateEntry> = updates
    .into_iter()
    .map(|(dst, tmp)| {
        let is_efi = dst.starts_with("EFI");
        let is_dir = destdir
            .metadata(tmp.as_str())
            .map(|m| m.is_dir())
            .unwrap_or(false);
        UpdateEntry {
            dst,
            tmp,
            is_efi,
            is_dir,
        }
    })
    .collect();

update_list.sort_by(|a, b| {
    b.is_efi
        .cmp(&a.is_efi)
        .then(b.is_dir.cmp(&a.is_dir))
        .then(a.dst.cmp(&b.dst))
});
```

Potentially another naive question but is it important that root-level directories come before individual files in this ordering?

src/efi.rs
```diff
@@ -806,100 +863,171 @@ pub struct EFIComponent {
...
+        .into_iter()
+        .filter_map(|e| e.ok())
+    {
+        if !entry.file_type().is_dir() {
+            continue;
+        }
 
-            let mut components = utf8_rel_path.components();
+        let abs_path = entry.path();
+        let rel_path = match abs_path.strip_prefix(sysroot) {
```
Rolv-Apneseth, line 895: Nit but could these also be `let else` blocks if the errors are just to be ignored anyway

src/efi.rs
```diff
@@ -806,100 +863,171 @@ pub struct EFIComponent {
...
     components.sort_by(|a, b| a.name.cmp(&b.name));
 
     Ok(Some(components))
 }
 
-/// Copy files from usr/lib/ostree-boot/efi/EFI to /usr/lib/efi/<component>/<evr>/
+/// Given a list of EFI components (potentially with multiple versions per
+/// component name), return only the latest version for each name.
+///
+/// Versions are compared lexicographically; this is sufficient for RPM EVR
```
Rolv-Apneseth, line 948: Is this still true?

Rolv-Apneseth (changes requested):
Take these with a grain of salt as I'm not very familiar with `bootupd` either, but adding a review to help move this PR forward

src/efi.rs
Rolv-Apneseth: *the lexicographical comparison that is

PR #1074, opened 2026-03-23
powerpc: Fix broken build

ckyrouac:
The powerpc build was attempting to use removed APIs from the blockdev crate. Update the powerpc target_device function to use the new Device APIs.

Assisted-by: Claude Code (Opus 4)


HuijingHei (approved):
LGTM

PR #1077, opened 2026-03-27
grubconfigs: Sync grub.cfg and bootuuid.cfg to all ESPs on multi-device setups

### LINKED ISSUE #1076: grub.cfg and bootuuid.cfg not synced to secondary ESPs on multi-device setups
## Summary

On multi-device setups (e.g. RAID 1 with an ESP per disk, or multi-disk installs via `--device`), `grubconfigs::install()` only writes `grub.cfg` and `bootuuid.cfg` to the single ESP currently mounted at `/boot/efi`. Secondary ESPs on other disks are left without these files. If the system boots from a disk whose ESP was not the one mounted during install, GRUB drops to a rescue shell because it cannot find its configuration.

## Steps to Reproduce

1. Set up a multi-device configuration with an ESP on each disk (e.g. RAID 1, or install with `bootupctl install --device /dev/sda --device /dev/sdb`)
2. Run `bootupctl install` (or let bootc perform the install)
3. Inspect the ESPs on each disk

## Expected Behavior

All ESPs should contain `EFI/<vendor>/grub.cfg` and `EFI/<vendor>/bootuuid.cfg` so the system can boot from any disk.

## Actual Behavior

Only the ESP mounted at `/boot/efi` receives these files. The other ESPs have the EFI binaries (shim, grub) but are missing the GRUB configuration files, making them unable to complete the boot process.

## Root Cause

`grubconfigs::install()` writes the static EFI `grub.cfg` and `bootuuid.cfg` into the `target_root`'s `/boot/efi/EFI/<vendor>/` directory. It has no awareness of additional ESP partitions on other devices. While `efi::install()` handles installing EFI binaries to all ESPs via `find_colocated_esps()`, the GRUB config writing step was not extended with the same multi-device logic.

## Impact

- Systems with redundant boot disks cannot boot from the secondary disk if the primary fails
- GRUB drops to a rescue shell with no `grub.cfg` present
- Defeats the purpose of redundant boot disk configurations

## Related

- #1059 — RAID 1 bootloader update failure (covers the ESP mount issue on RAID members, fixed separately in 2bf2dfb)

ckyrouac:
On multi-device setups (e.g. RAID 1 with an ESP per disk), grubconfigs::install() only writes grub.cfg and bootuuid.cfg to the single ESP currently mounted at /boot/efi. This means that if the system boots from a different disk, GRUB drops to a shell because there is no grub.cfg on that ESP.

Add sync_grub_configs_to_all_esps() which runs after the primary grubconfigs::install(). It discovers all ESP partitions across the provided devices, mounts each one to a temporary directory, and calls the new grubconfigs::install_to_esp() helper to write the static EFI grub.cfg and bootuuid.cfg.

Also refactor the ESP config writing in grubconfigs::install() into a shared write_esp_configs() helper to avoid code duplication.

Fixes #1076

Assisted-by: Claude Code (Opus 4.6)

### DIFF
--- a/src/bootupd.rs
+++ b/src/bootupd.rs
@@ -187,11 +187,30 @@ pub(crate) fn install(
         Some(uuid) => {
             let meta = get_static_config_meta()?;
             state.static_configs = Some(meta);
+
+            // On multi-device setups (e.g. RAID 1 with an ESP per disk) we
+            // need every ESP to carry grub.cfg and bootuuid.cfg so the system
+            // can boot from any disk.  Gather the additional ESP device paths
+            // now and pass them into grubconfigs::install so it writes to all
+            // of them.
+            let additional_esps: Vec<String> = if installed_efi_vendor.is_some() {
+                devices
+                    .first()
+                    .and_then(|dev| dev.find_colocated_esps().ok().flatten())
+                    .unwrap_or_default()
+                    .iter()
+                    .map(|dev| dev.path())
+                    .collect()
+            } else {
+                Vec::new()
+            };
+
             crate::grubconfigs::install(
                 sysroot,
                 Some(&source_root_dir),
                 installed_efi_vendor.as_deref(),
                 uuid,
+                &additional_esps,
             )?;
             // On other architectures, assume that there's nothing to do.
         }
--- a/src/bios.rs
+++ b/src/bios.rs
@@ -191,7 +191,7 @@ impl Component for Bios {
                 .context("Failed to backup GRUB config")?;
         }
 
-        crate::grubconfigs::install(&destdir, None, None, true)?;
+        crate::grubconfigs::install(&destdir, None, None, true, &[])?;
 
         // Remove the real config if it is symlink and will not
         // if /boot/grub2/grub.cfg is file
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -356,7 +356,7 @@ impl Component for Efi {
                 .context("Failed to backup GRUB config")?;
         }
 
-        grubconfigs::install(&sysroot, None, Some(&vendor), true)?;
+        grubconfigs::install(&sysroot, None, Some(&vendor), true, &[])?;
         // Synchronize the filesystem containing /boot/efi/EFI/{vendor} to disk.
         fsfreeze_thaw_cycle(efidir.open_file(".")?)?;
 
--- a/src/grubconfigs.rs
+++ b/src/grubconfigs.rs
@@ -27,12 +27,17 @@ const GRUB_FILES: [&str; 3] = ["bootuuid.cfg", GRUBCONFIG, GRUBENV];
 pub(crate) const GRUBCONFIG_FILE_MODE: u32 = 0o600;
 
 /// Install the static GRUB config files.
+///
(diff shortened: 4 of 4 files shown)


ckyrouac:
pushing up what claude generated in case anyone wants to review it while I test/validate locally and add a test.

src/bootupd.rs
```diff
@@ -210,6 +225,89 @@ pub(crate) fn install(
...
+    write_uuid: bool,
+) -> Result<()> {
+    if devices.is_empty() {
+        return Ok(());
+    }
+
+    // Collect all ESP partition device paths.
+    let esp_parts: Vec<String> = devices
+        .iter()
+        .filter_map(|dev| dev.find_partition_of_esp().ok().map(|p| p.path()))
```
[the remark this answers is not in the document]
cgwalters: Right, in the bootc PR I added a new API for this

src/bootupd.rs
```diff
@@ -210,6 +225,89 @@ pub(crate) fn install(
...
+        .context("Opening /boot/grub2")?;
+
+    let tmpdir = tempfile::tempdir().context("Creating temporary mount point")?;
+    let tmpmnt = tmpdir.path();
+
+    for esp_path in &esp_parts {
+        log::info!("Syncing GRUB configs to ESP {esp_path}");
+
+        // Mount the ESP
+        std::process::Command::new("mount")
```
cgwalters: Nonblocking but I'd like to use the mount-as-fd new mount API for stuff like this in the future

src/bootupd.rs
```diff
@@ -210,6 +225,89 @@ pub(crate) fn install(
...
+        })();
+
+        // Always unmount, even on error
+        std::process::Command::new("umount")
+            .arg(tmpmnt)
+            .run_inherited()
+            .with_context(|| format!("Unmounting ESP {esp_path}"))?;
+
+        result?;
+    }
```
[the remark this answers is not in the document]
cgwalters: We run in a private mountns anyways which avoids global leaks.

src/bootupd.rs
```diff
@@ -193,6 +194,20 @@ pub(crate) fn install(
                 installed_efi_vendor.as_deref(),
                 uuid,
             )?;
+
+            // grubconfigs::install writes grub.cfg and bootuuid.cfg to the
+            // single ESP currently mounted at /boot/efi.  On multi-device
+            // setups we need to sync these files to every other ESP so that
+            // any disk can boot independently.
+            if let Some(ref vendor) = installed_efi_vendor {
+                sync_grub_configs_to_all_esps(
```
cgwalters: I think it'd be a lot cleaner for us to get the list of explicit devices and then do an install on each one.

I think this would require changing the generic `crate::grubconfigs::install` invocation above instead.

Now I am not sure if we support multi-device BIOS. But basically WDYT about ensuring we gather all devices, and use that in the above?

cgwalters (commented):
One existing problem we had is that I messed up the static configs stuff and didn't wire them in as something updatable in place. If we'd had that, this could have just been part of a general "sync content across ESP".

src/bootupd.rs
ckyrouac: updated with this suggestion

PR #1072, opened 2026-03-17
update: Skip bootloader update when no block devices back the root

cgwalters:
Fix the problem that `bcvk ephemeral run quay.io/fedora/fedora-bootc:43` shows a systemd error by default.

PR #1088, opened 2026-04-21
Disable systemd service on CoreOS live filesystems

nikita-dubrovskii:
When running from live-iso `bootupctl` fails with:
```
    error: Querying adoptable state: creation time is not available for the filesystem
```
That happens because filesystem (erofs|squashfs) doesn't support statx.stx_btime field:
```
    $ stat /sysroot/.coreos-aleph-version.json
    Access: 2022-08-01 23:42:11.000000000 +0000
    Modify: 2022-08-01 23:42:11.000000000 +0000
    Change: 2022-08-01 23:42:11.000000000 +0000
     Birth: -
```
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/2136

### DIFF
--- a/systemd/bootloader-update.service
+++ b/systemd/bootloader-update.service
@@ -4,6 +4,9 @@ Documentation=https://github.com/coreos/bootupd
 
 [Service]
 Type=oneshot
+# It doesn't make sense to update bootloader in "Live" environments.
+# https://github.com/coreos/fedora-coreos-tracker/issues/2136
+ExecCondition=/bin/bash -c '[[ ! $(findmnt -n -o FSTYPE /sysroot) =~ ^(erofs|squashfs)$ ]]'
 ExecStart=/usr/bin/bootupctl update
 RemainAfterExit=yes
 # Keep this stuff in sync with SYSTEMD_ARGS_BOOTUPD in general


systemd/bootloader-update.service
```diff
@@ -4,6 +4,12 @@ Documentation=https://github.com/coreos/bootupd
 
 [Service]
 Type=oneshot
+# bootupd is enabled by default in CoreOS (F43+ / CentOS9+).
+# On live ISOs, it fails to read metadata from
+# "sysroot/.coreos-aleph-version.json" because the filesystem
+# does not support statx.stx_btime.
+# https://github.com/coreos/fedora-coreos-tracker/issues/2136
+ExecCondition=/bin/bash -c '[[ ! $(findmnt -n -o FSTYPE /sysroot) =~ ^(erofs|squashfs)$ ]]'
```
dustymabe, line 9: Optional: this might make it a little more readable:


```suggestion
ExecCondition=/bin/bash -c '[[ $(findmnt -n -o FSTYPE /sysroot) !~ ^(erofs|squashfs)$ ]]'
```

systemd/bootloader-update.service
```diff
@@ -4,6 +4,12 @@ Documentation=https://github.com/coreos/bootupd
 
 [Service]
 Type=oneshot
+# bootupd is enabled by default in CoreOS (F43+ / CentOS9+).
+# On live ISOs, it fails to read metadata from
+# "sysroot/.coreos-aleph-version.json" because the filesystem
```
dustymabe: I realize here that the problem we hit was specific (i.e. timestamp on `sysroot/.coreos-aleph-version.json`), but I'm thinking the reasoning for this condition shouldn't be that speicific.

i.e. I was thinking along the [same lines](https://github.com/coreos/fedora-coreos-tracker/issues/2136#issuecomment-4287776226) as @travier when he said: 

> Where are we running bootupd from the LiveISO? I don't think we should support that.


So I was thinking what we should do is detect when it wouldn't make sense for bootupd to run to try to update the bootloader. That's why my suggestion over in this comment](https://github.com/coreos/fedora-coreos-config/pull/4121#issuecomment-4290627199) was related to the filesystem on `/boot` (i.e. boot.mount) and not /sysroot.

All of this is to say.. should we change the check below to findmnt on `/boot` instead and change this comment to something more about "it doesn't make sense to update bootloader in in "Live" environments. Rather than specifically mentioning the aleph file timestamp thing.

dustymabe (approved):
LGTM - some suggestions.

systemd/bootloader-update.service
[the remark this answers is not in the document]
dustymabe: interesting.. if it works, could make this cleaner

dustymabe: see gemini suggestion too: https://github.com/coreos/bootupd/pull/1088#discussion_r3124005490

systemd/bootloader-update.service
nikita-dubrovskii: Problem with `/boot` is it's not mounted in fcos:
```
core@localhost:~$ findmnt /boot                                                                                                                                                                                                                                                  
core@localhost:~$ echo $?                                                                                                                                                                                                                                                                 
1
```

But checking `/sysroot` always works.

Update: i speak here about Live env.

systemd/bootloader-update.service
```diff
@@ -4,6 +4,10 @@ Documentation=https://github.com/coreos/bootupd
 
 [Service]
 Type=oneshot
+# Service is enabled by default in CoreOS (F43+ / CentOS9+), but
+# it doesn't make sense to update bootloader in "Live" environments.
+# https://github.com/coreos/fedora-coreos-tracker/issues/2136
+ExecCondition=/usr/bin/findmnt --mountpoint /sysroot --types noerofs,nosquashfs
```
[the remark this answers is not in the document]
nikita-dubrovskii: AI seems to be partially wrong:
```
core@localhost:~$ findmnt --mountpoint /sysroot --types nosquashfs,erofs
core@localhost:~$ findmnt --mountpoint /sysroot --types nosquashfs,noerofs
core@localhost:~$ findmnt --mountpoint /sysroot --types squashfs,noerofs
core@localhost:~$ findmnt --mountpoint /sysroot --types squashfs,erofs
TARGET   SOURCE     FSTYPE OPTIONS
/sysroot /dev/loop1 erofs  ro,relatime,seclabel,user_xattr,acl,cache_strategy=readaround

```

systemd/bootloader-update.service
nikita-dubrovskii: kept this with your suggestion, as `[[ ... ]]` imo looks better

systemd/bootloader-update.service
dustymabe: hmm. I guess maybe a behavior difference between FCOS and RHCOS. I had a RHCOS 10.2 ISO laying around and this is what I see:

```
Red Hat Enterprise Linux CoreOS 10.2.20260325-0 (Coughlan)
Kernel 6.12.0-211.5.1.el10_2.x86_64 on x86_64

SSH host key: SHA256:4u+zqJiPzdjNR5/nOGCzDDjAoc9lFZw6NYm3YIfsMRM (ED25519)
SSH host key: SHA256:z1K8u0Gjyz/XcwGvchYgILnlQ6HfgduV8K3TjAa7dqg (ECDSA)
SSH host key: SHA256:M9ythKHnv+LZajdYWb3MYs8m+11M6lOws4Jb67khk9w (RSA)
Ignition: ran on 2026/04/22 13:40:28 UTC (this boot)
Ignition: no config provided by user
localhost login: core (automatic login)

Last login: Wed Apr 22 13:40:41 on tty1

###########################################################################
Welcome to the CoreOS live environment. This system is running completely
from memory, making it a good candidate for hardware discovery and
installing persistently to disk. Here is an example of running an install
to disk via coreos-installer:

sudo coreos-installer install /dev/sda \
    --ignition-url https://example.com/example.ign

You may configure networking via 'sudo nmcli' or 'sudo nmtui' and have
that configuration persist into the installed system by passing the
'--copy-network' argument to 'coreos-installer install'. Please run
'coreos-installer install --help' for more information on the possible
install options.
###########################################################################

[core@localhost ~]
[core@localhost ~]$ findmnt /boot
TARGET SOURCE            FSTYPE OPTIONS
/boot  /dev/loop1[/boot] erofs  ro,relatime,seclabel,user_xattr,acl,cache_strategy=readaround
```

nikita-dubrovskii: Yep, on rhcos `/boot` is mounted.

systemd/bootloader-update.service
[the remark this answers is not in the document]
dustymabe: but that looks like the correct behavior? I think only the first and 4th invocations you showed are correct (IIUC from the man page `no` only goes at the front of the list and applies to all specified). 

so `nosquashfs,erofs` will only find filesystems that aren't squashfs or erofs and `squashfs,erofs` will only find ones that are either of those.

systemd/bootloader-update.service
nikita-dubrovskii: Upd: just re-checked and noticed `!~` doesn't exist, it's `awk`

systemd/bootloader-update.service
dustymabe: ok so /boot is mounted in the ISO in RHCOS, but not FCOS?

dustymabe (approved):
LGTM

systemd/bootloader-update.service
nikita-dubrovskii: yes.

cgwalters:
I believe https://github.com/coreos/bootupd/pull/1072 would fix this too, and I think that approach is cleaner. WDYT?

dustymabe:
> I would fix this too, and I think that approach is cleaner. WDYT?

@nikita-dubrovskii mind evaluating https://github.com/coreos/bootupd/pull/1072 ?

nikita-dubrovskii:
> I believe #1072 would fix this too

Nope, added some traces:
```
core@localhost:~$ journalctl -u bootloader-update.service 
Apr 23 09:18:51 localhost systemd[1]: Starting bootloader-update.service - Update bootloader on boot...
Apr 23 09:18:52 localhost bootupctl[1137]: TRACE: Calling get_devices from prep_before_update()
Apr 23 09:18:52 localhost bootupctl[1137]: TRACE: get_devices(target_root=/)
Apr 23 09:18:52 localhost bootupctl[1137]: TRACE: inspect_filesystem(path=.)
Apr 23 09:18:52 localhost bootupctl[1137]: TRACE: inspect_filesystem(path=.)
Apr 23 09:18:52 localhost bootupctl[1137]: TRACE: inspect_filesystem returned: source=/dev/loop1, fstype=erofs
Apr 23 09:18:52 localhost bootupctl[1137]: TRACE: get_devices returned: Some([])
Apr 23 09:18:52 localhost bootupctl[1137]: TRACE: Calling query_adopt_state from status() for component: BIOS
Apr 23 09:18:52 localhost bootupctl[1137]: TRACE: query_adopt_state()
Apr 23 09:18:52 localhost.localdomain bootupctl[1137]: error: Querying adoptable state: creation time is not available for the filesystem
Apr 23 09:18:52 localhost.localdomain systemd[1]: bootloader-update.service: Main process exited, code=exited, status=1/FAILURE
Apr 23 09:18:52 localhost.localdomain systemd[1]: bootloader-update.service: Failed with result 'exit-code'.
Apr 23 09:18:52 localhost.localdomain systemd[1]: Failed to start bootloader-update.service - Update bootloader on boot.
```

> and I think that approach is cleaner. WDYT?

Question is should we skip run in Live-env or run and exit gracefully. I'm for skipping run completely

nikita-dubrovskii:
@cgwalters , i've attached a patch to your PR, with it everything works:
```
core@localhost:~$ journalctl -u bootloader-update.service 
Apr 23 11:48:29 localhost systemd[1]: Starting bootloader-update.service - Update bootloader on boot...
Apr 23 11:48:30 localhost bootupctl[1069]: No block-backed boot filesystem found; bootloader update is not applicable, skipping.
Apr 23 11:48:30 localhost systemd[1]: Finished bootloader-update.service - Update bootloader on boot.
```

nikita-dubrovskii:
@cgwalters  saw your rebased PR. Still doesn't work:
```
Apr 24 07:43:05 localhost bootupctl[1107]: client_run_update: rootcxt=Device { name: "loop1", serial: None, model: None, partlabel: None, parttype: None, partuuid: None, partn: None, children: None, size: 904198144, maj_min: Some("7:1"), start: None, label: None, fstype: Some("ero"), uuid: Some("cb36fafd-58a7-4728-a485-6dba0af008c8"), path: Some("/dev/loop1"), pttype: None }
Apr 24 07:43:06 localhost.localdomain bootupctl[1107]: error: Querying adoptable state: creation time is not available for the filesystem
Apr 24 07:43:06 localhost.localdomain systemd[1]: bootloader-update.service: Main process exited, code=exited, status=1/FAILURE
```

Here is a [branch](https://github.com/coreos/bootupd/compare/main...nikita-dubrovskii:bootupd:ephemeral-noupdate) with fix, mind checking it?

systemd/bootloader-update.service
```diff
@@ -4,6 +4,10 @@ Documentation=https://github.com/coreos/bootupd
 
 [Service]
 Type=oneshot
+# Service is enabled by default in CoreOS (F43+ / CentOS9+), but
```
dustymabe: I would just drop this line and capitalize the `It` in the next line.

dustymabe:
I think it makes sense to do this AND #1072 

This still LGTM to merge if you like Nikita.

dustymabe (approved):

systemd/bootloader-update.service
nikita-dubrovskii: Done

Issue #1075, opened 2026-03-25
EFI timestamp should respect SOURCE_DATE_EPOCH for reproducible builds


p5:
bootupctl backend generate-update-metadata writes a timestamp field in usr/lib/bootupd/updates/EFI.json using `std::time::SystemTime::now()` ([src/efi.rs:604](https://github.com/coreos/bootupd/blob/main/src/efi.rs#L604)). This was introduced as a workaround for #933.

This is the sole source of non-reproducibility in our container image builds. When rebuilding the same image from the same inputs, the only difference between the two images is this timestamp:

```
-    "timestamp": "2026-03-25T10:37:38.173666102Z",
+    "timestamp": "2026-03-25T11:05:02.924829010Z",
```

When SOURCE_DATE_EPOCH is set, generate-update-metadata should use it for the EFI timestamp instead of wall-clock time. This is the standard (https://reproducible-builds.org/specs/source-date-epoch/) mechanism for reproducible builds and is already used by tools like podman build --source-date-epoch.

When SOURCE_DATE_EPOCH is not set, behaviour should remain unchanged (use `SystemTime::now()`).

The BIOS component ([src/bios.rs](https://github.com/coreos/bootupd/blob/main/src/bios.rs)) derives its timestamp from RPM build times via `packagesystem::query_files()`, so it is already deterministic and does not need this change.

cgwalters:
SGTM this should just be a one-liner patch

PR #1089, opened 2026-04-23
efi: Choose update layout depending on /usr/lib/efi

### LINKED ISSUE #1057: Make `bootupctl backend generate-metadata|install|update` use the new layout only on systems with `/usr/lib/efi`
See discussion in https://github.com/coreos/bootupd/pull/995#issuecomment-3775224575 and  https://github.com/coreos/bootupd/pull/995#issuecomment-3783532030

Johan-Liebert1:
In https://github.com/coreos/bootupd/pull/995 changes were made to move files from `usr/lib/ostree-boot/EFI` to `usr/lib/efi` which caused issues with older bootupd failing to understand the new layout created by newer version of bootupd. We will start putting grub, shim and other related stuff in /usr/lib/efi starting fedora44 while older releases still have them in /usr/lib/ostree-boot, we need bootupd to properly handle both cases.

To make the change backwards compatible, we now check for the existence of `usr/lib/efi`, and only use the new layout if we find it. Else, default back to the older layout at `usr/lib/bootupd`

Fixes: https://github.com/coreos/bootupd/issues/1057

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -568,50 +568,49 @@ impl Component for Efi {
     }
 
     fn generate_update_metadata(&self, sysroot: &str) -> Result<Option<ContentMetadata>> {
-        let sysroot_path = Path::new(sysroot);
+        let sysroot_path = Utf8Path::new(sysroot);
         let sysroot_dir = Dir::open_ambient_dir(sysroot_path, cap_std::ambient_authority())?;
 
-        if let Some(ostreeboot) = sysroot_dir
+        let ostreeboot = sysroot_dir
             .open_dir_optional(ostreeutil::BOOT_PREFIX)
-            .context("Opening usr/lib/ostree-boot")?
-        {
+            .context("Opening usr/lib/ostree-boot")?;
+
+        // Newer images >= F44 have boot entries in /usr/lib/efi, but older images
+        // have them in /usr/lib/ostree-boot, which we move to /usr/lib/bootupd/updates/EFI
+        let metadata = if sysroot_path.join(EFILIB).exists() {
+            println!("Generating metadata from {EFILIB}");
+            generate_meta_from_usr_efi(sysroot_path)?
+        } else {
+            match &ostreeboot {
+                Some(..) => {
+                    println!(
+                        "Moving {}/efi/EFI to {BOOTUPD_UPDATES_DIR}/EFI",
+                        ostreeutil::BOOT_PREFIX
+                    );
+                    // Transfer usr/lib/ostree-boot/EFI files to usr/lib/bootupd/updates/EFI
+                    transfer_ostree_boot_to_bootupd_updates(sysroot_path, self)?
+                }
+
+                // If /usr/lib/ostree-boot doesn't exist, assume /usr/lib/efi will
+                None => anyhow::bail!("/usr/lib/ostree-boot and /usr/lib/efi not found"),
+            }
+        };
+
+        if let Some(ostreeboot) = ostreeboot {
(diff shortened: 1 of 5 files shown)


src/efi.rs
```diff
@@ -801,6 +795,34 @@ fn find_file_recursive<P: AsRef<Path>>(dir: P, target_file: &str) -> Result<Vec<
     Ok(result)
 }
 
+#[context("Generating metadata from usr/lib/efi")]
+fn generate_meta_from_usr_efi(sysroot_path: &Utf8Path) -> Result<ContentMetadata> {
+    let Some(efi_components) = get_efi_component_from_usr(sysroot_path, EFILIB)? else {
+        anyhow::bail!("Failed to find EFI components");
+    };
```
jbtrystram, line 807: That's an unusual way of handling the unwrapping. 
Why do you need to `bail` here ? Can't the error simply be bubbled up ?

src/efi.rs
```diff
@@ -801,6 +795,34 @@ fn find_file_recursive<P: AsRef<Path>>(dir: P, target_file: &str) -> Result<Vec<
...
+    for efi in efi_components {
+        packages.push(format!("{}-{}", efi.name, efi.version));
+        modules_vec.push(Module {
+            name: efi.name,
+            rpm_evr: efi.version,
+        });
+    }
+    modules_vec.sort_unstable();
+
+    // change to now to workaround https://github.com/coreos/bootupd/issues/933
```
jbtrystram, line 820: Should we account for `SOURCE_DATE_EPOCH` here ?
See https://github.com/coreos/bootupd/issues/1075

jbtrystram (commented):
Superficial review since I am not really knowledgeable about this, but it feels a bit weird overall. 
If i understand this, this is addressing the case where a disk image was created using a more recent bootupd version that's what actually ships into the image. This is essentially a bug from the build system.

Also, why not simply always symlink `usr/lib/bootupd/updates` to `/lib/ostree-boot/EFI` when the second exist but not the first, instead of moving things around ?
I guess there is already migration logic the other way around that will happen later ?

src/efi.rs
Johan-Liebert1: `get_efi_component_from_usr` returns `Result<Option<T>>`, the `let .. else {}` is to handle the `Option` part after the result is bubbled up by the `?` operator

jbtrystram: Ah, `T` is an option already, I missed that. There is `flatten` that could work I guess.

src/efi.rs
Johan-Liebert1: This does sound like a one liner fix, afaics we are comparing package names and version for sorting and not the timestamps. This could be a separate PR though. wdyt?

travier:
Reading this PR, I've realized that the code in #955 is incorrectly copying things from /usr/lib/ostree-boot to /usr/lib/efi in the /usr/lib/efi case when it should only look at what files are present in /usr/lib/ostree-boot to generate the metadata but read the files from /usr/lib/efi (they should be an exact copy but well).

Johan-Liebert1:
> the /usr/lib/efi case when it should only look at what files are present in /usr/lib/ostree-boot

wouldn't /usr/lib/ostree-boot not have anything in the /usr/lib/efi case?

travier:
Damned, that's likely right and don't remember what we checked the last time :/ But the code does not read like that (I need to re-read that).

travier (approved):
We updated this one together with Pragyan. LGTM

Issue #1040, opened 2025-12-17
Rework safety guard to support empty EFI component


travier:
In the automotive use case, some images will have empty EFI components so we should support that.

In https://github.com/coreos/bootupd/issues/994, we added a check that makes sure the EFI component has at least 2 entries, which thus fails.

What we want instead if the following:

- If there are no EFI component found, then we don't include EFI.json
- If there are some EFI component found:
  - Check if shim is part of them:
    - If it is, then there must be another component (i.e. bootloader) included
    - If it's not, do nothing special
  - Check if grub is part of them:
    - If it is, then there must be shim as another component
    - If it's not, do nothing special

cgwalters:
> /usr/bin/bootupctl backend generate-update-metadata --expect grub2 --expect shim

SGTM

PR #1096, opened 2026-05-19
Only do "booted BIOS with no update" check during actual update

### LINKED ISSUE #1035: Fail updates if metadata is missing in the image for current boot type

alexlarsson:
This fixes https://github.com/coreos/bootupd/issues/1035 which introduced this check, but did so in every query_update(). However, that is called from non-update places too. In particular from install(), which causes bootc-image-builder to break if used on non-EFI systems.

Instead we break out the check to a separate component function query_requires_update() which is only called in the update-with-missing-component case.

### DIFF
--- a/src/component.rs
+++ b/src/component.rs
@@ -73,6 +73,11 @@ pub(crate) trait Component {
     /// Used on the client to query for an update cached in the current booted OS.
     fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>>;
 
+    /// This is called in the update code if query_update() returned no metadata.
+    /// It should return an error if the current booted system should expect some
+    /// metadata for this component.
+    fn query_requires_update(&self, sysroot: &openat::Dir) -> Result<()>;
+
     /// Used on the client to run an update.
     fn run_update(
         &self,
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -613,12 +613,15 @@ impl Component for Efi {
     }
 
     fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>> {
-        let content_metadata = get_component_update(sysroot, self)?;
+        get_component_update(sysroot, self)
+    }
+
+    fn query_requires_update(&self, _sysroot: &openat::Dir) -> Result<()> {
         // Failed as expected if booted with EFI and no update metadata
-        if content_metadata.is_none() && is_efi_booted()? {
+        if is_efi_booted()? {
             anyhow::bail!("Failed to find EFI update metadata");
         }
-        Ok(content_metadata)
+        Ok(())
     }
 
     fn validate(&self, current: &InstalledContent, device: &Device) -> Result<ValidationResult> {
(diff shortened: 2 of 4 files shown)


src/efi.rs
```diff
@@ -613,12 +613,15 @@ impl Component for Efi {
...
 
     fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>> {
-        let content_metadata = get_component_update(sysroot, self)?;
-        // Failed as expected if booted with EFI and no update metadata
-        if content_metadata.is_none() && is_efi_booted()? {
+        get_component_update(sysroot, self)
+    }
+
+    fn query_requires_update(&self, _sysroot: &openat::Dir) -> Result<()> {
+        // Failed as expected if booted with BIOS and no update metadata
```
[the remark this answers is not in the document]
alexlarsson: fixed

src/component.rs
```diff
@@ -73,6 +73,9 @@ pub(crate) trait Component {
     /// Used on the client to query for an update cached in the current booted OS.
     fn query_update(&self, sysroot: &openat::Dir) -> Result<Option<ContentMetadata>>;
 
+    /// Used on the client to query if the booted systems requires an update of this component
+    fn query_requires_update(&self, sysroot: &openat::Dir) -> Result<()>;
```
Johan-Liebert1, line 79: imo we should accept `Option<ContentMetadata>` here so that we only return an error if the metadata is `None`. Otherwise just looking at the function, it's hard to figure out why we just error our if `sys/firmware/efi` doesn't exist for BIOS, when it not existing would always be true for BIOS booted systems

Johan-Liebert1 (commented):
One nit, else lgtm

src/component.rs
alexlarsson: Then the caller is quite weird if it always calls query_update() and then query_requires_update(). Right now it only calls query_requires_update() if metadata is None. If you want to do it like that, then I think having query_requires_update() call query_update() and then do the check is better than having to call both and passing the metadata forward.

Johan-Liebert1: > Right now it only calls query_requires_update() if metadata is None

right, and this is a little weird to me as it's not documented. Maybe the best approach is to simply document this in the function comments? Something like `This function assumes that metadata for the component was None. It then throws an error if the system was booted with "component"`?

alexlarsson: I pushed a new version with the function comment having a bit more detail for when this is called and what it should do.

travier:
It feels to me that we are working around the fact that we added this as a quick workaround but what we really need is https://github.com/coreos/bootupd/issues/1040 instead.

cgwalters:
> It feels to me that we are working around the fact that we added this as a quick workaround but what we really need is https://github.com/coreos/bootupd/issues/1040 instead.

We need both though right? Or at least we need something like this which removes the dynamic check.

cgwalters (approved):

PR #1105, opened 2026-06-02
Release 0.2.35

ckyrouac:
See: https://github.com/coreos/bootupd/issues/1103


cgwalters:
I think Joel might have been starting this in https://github.com/coreos/bootupd/issues/1103 ?

ckyrouac:
nice, closing in favor of https://github.com/coreos/bootupd/issues/1103

Issue #1080, opened 2026-04-02
Add support for systemd-boot (and minimal GRUB2 builds)


travier:
We will need bootupd to support installing and updating minimal bootloaders like systemd-boot (and a potentially upcoming minimal build of GRUB2) alongside shim.

Initial support was being worked on in https://github.com/coreos/bootupd/pull/978 but we should not use `bootctl` directly as we also need to install and update `shim` alongside `systemd-boot`.

travier:
Or implement support in bootctl: https://github.com/systemd/systemd/issues/27234

travier:
The end goal is that we want to have a single image that will be used for all installation methods:
- Current state:
    - rpm-ostree/bootc with ostree+composefs backend without fs-verity
    - with GRUB2 + static GRUB configs
    - and split kernel/initramfs
- What we want to migrate the existing systems to:
    - bootc with pure composefs backend
    - with GRUB2 + static GRUB configs
    - and split kernel/initramfs
- Sealed images and potentially new default in the future:
    - bootc with pure composefs backend
    - with minimal GRUB2 build / systemd-boot, no GRUB config
    - and UKI or split kernel/initramfs

See https://hackmd.io/WDPXRiUsQe2ugBhjp147xQ for a longer discussion about those layouts.

Thus we need to include all of shim, current GRUB2 and the minimal GRUB2 / systemd-boot in a single image and add a way for bootupd to figure out which one to install and update.

At installation time, we will need some form of "runtime" configuration to select which bootloader to install as the configuration can not (always) be part of the image itself. This is similar to the filesystem option for bootc base images in that you can derive an image and set a fixed filesystem/bootloader but the default image will have to support all of them.

At update time, we can either rely on a key stored in `/boot/bootupd-state.json` or query the currently used bootloader from EFI vars to decide which one to use.

Shim currently hardcodes the name of the bootloader that it will execute once it's done setting things up: https://src.fedoraproject.org/rpms/shim-unsigned-x64/blob/rawhide/f/shim-unsigned-x64.spec#_127

As both the minimal GRUB2 and systemd-boot are not named as `grub<efiarch>.efi` in the packages themselves, we will have to add logic to bootupd to write the choosen bootloader under the expected name.

See:
- https://koji.fedoraproject.org/koji/rpminfo?rpmID=47120161
- https://koji.fedoraproject.org/koji/rpminfo?rpmID=47120163
[...]

PR #1113, opened 2026-06-12
Integration for GrubCC and SystemdBoot

Johan-Liebert1:
Add Bootloader enum

Now that we are going to support multiple bootloaders, we need a way to
keep track of which one we're operating on

---

For `bootupd generate-update-metadata` we now generate metadata for all
found bootloaders. The final json will look like

```json
{
  "timestamp": "2026-06-10T09:52:58.107743026Z",
  "version": "grub-cc-1:2.12-59.fc45,grub2-1:2.12-58.fc44,shim-16.1-5",
  "versions": [
    { "name": "grub-cc", "rpm_evr": "1:2.12-59.fc45" },
    { "name": "grub2", "rpm_evr": "1:2.12-58.fc44" },
    { "name": "shim", "rpm_evr": "16.1-5" }
  ]
}
```

This allows us to have multiple bootloaders in a single image and
install/update based on preference.

During install/update, we check to bootloader and filter out any and all
entries from the update metadata that do not match or are not required
for the bootloader

---

For grub we store the statefile in `/sysroot/boot/bootupd.json`. For
grub-cc, and in future systemd-boot, we won't have a `/boot`, so we now
store the state pre bootloader.

Grub2: Statefile is still in `/sysroot/boot/bootupd.json`
GrubCC: Statefile is stored in all the ESPs

Store the sysroot_path in StateLockGuard which we require for mounting
the ESP. Also, mount the ESP at `/boot` which is what the BLS spec suggest

---

See: https://github.com/coreos/bootupd/issues/1080

### DIFF
--- a/Dockerfile.bls
+++ b/Dockerfile.bls
@@ -68,7 +68,7 @@ rpm -ql systemd-boot-unsigned
 
 . /etc/os-release
 
-# Create the expected directory structure at /usr/lib/efi/grub-cc/<evr>/EFI/fedora/
+# Create the expected directory structure at /usr/lib/efi/systemd-boot/<evr>/EFI/fedora/
 mkdir -p "/usr/lib/efi/systemd-boot/${evr}/EFI/$ID"
 
 if [[ $(uname -m) == x86_64 ]]; then
--- a/Dockerfile.bls
+++ b/Dockerfile.bls
@@ -1,3 +1,4 @@
+# Dockerfile to test GrubCC and systemd-boot
 # Build from the current git into a fedora44 container image.
 
 ARG base=quay.io/fedora/fedora-bootc:44
@@ -42,7 +43,6 @@ mv ./*.rpm grub-cc.rpm
 EOF
 
 FROM $base
-ARG grubcc
 # Clean out the default to ensure we're using our updated content
 RUN rpm -e bootupd
 COPY --from=build /out/ /
@@ -50,7 +50,7 @@ COPY --from=grub-cc-download /grub-cc.rpm /var/grub-cc.rpm
 # Install bootc from copr
 RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp <<EORUN
 set -xeuo pipefail
-dnf -y install dnf-plugins-core bootc
+dnf -y install dnf-plugins-core bootc systemd-boot tree
 dnf clean all
 rm -rf /var/log
 rm -rf /var/lib
@@ -59,6 +59,29 @@ rm -rf /run/rhsm
 rm -rf /tmp/*
 EORUN
 
+# Install systemd-boot
+RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp <<EOF
+set -eux
+# epoch for systemd-boot-unsigned is "(none)"
+evr=$(rpm -q --qf '%{EPOCH}:%{VERSION}-%{RELEASE}\n' systemd-boot-unsigned)
+rpm -ql systemd-boot-unsigned
+
+. /etc/os-release
+
+# Create the expected directory structure at /usr/lib/efi/grub-cc/<evr>/EFI/fedora/
+mkdir -p "/usr/lib/efi/systemd-boot/${evr}/EFI/$ID"
+
+if [[ $(uname -m) == x86_64 ]]; then
+    grubName="grubx64.efi"
+else
+    grubName="grubaa64.efi"
+fi
+
+cp /usr/lib/systemd/boot/efi/systemd-boot*.efi "/usr/lib/efi/systemd-boot/${evr}/EFI/$ID/$grubName"
+
+EOF
+
+# Install grub-cc
 RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp <<EOF
 
 set -eux
@@ -67,11 +90,6 @@ cleanup() {
     # Clean up temporary files
     rm -rvf /var/grub-cc*
 }
-
-if [[ ${grubcc} != 1 ]]; then
-    cleanup
-    exit 0
-fi
 
 # Extract and install grub-cc at the correct path
 mkdir /var/grub-cc
@@ -83,9 +101,15 @@ evr=$(rpm -qp --qf '%{EPOCH}:%{VERSION}-%{RELEASE}' /var/grub-cc.rpm)
 
 . /etc/os-release
 
+if [[ $(uname -m) == x86_64 ]]; then
+    grubName="grubx64.efi"
+else
+    grubName="grubaa64.efi"
+fi
+
 # Create the expected directory structure at /usr/lib/efi/grub-cc/<evr>/EFI/fedora/
 mkdir -p "/usr/lib/efi/grub-cc/${evr}/EFI/$ID/"
-cp "$file" "/usr/lib/efi/grub-cc/${evr}/EFI/$ID/grubx64.efi"
+cp "$file" "/usr/lib/efi/grub-cc/${evr}/EFI/$ID/$grubName"
 
 cleanup
 
--- a/scripts/test-bootloader.sh
+++ b/scripts/test-bootloader.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+set -eux
+
+IMG_NAME=$1
+BOOTLOADER=$2
+
+case $BOOTLOADER in
+    systemd)
+        EFI_DIR_NAME=systemd-boot
+        ;;
+    grub-cc)
+        EFI_DIR_NAME=grub-cc
+        ;;
+    grub)
+        EFI_DIR_NAME=grub2
+        ;;
+esac
+
+cat <<-EOF > sfdisk-buf
+label: gpt
+label-id:  65be9332-59ba-11f1-9b26-6a8e2ab625e4
+size=1Gib, type=C12A7328-F81F-11D2-BA4B-00A0C93EC93B, name="EFI-SYSTEM"
+           type=4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709, name="root"
+EOF
+
+truncate -s4G "${BOOTLOADER}-test.img"
+
+cat sfdisk-buf | sfdisk --wipe=always "${BOOTLOADER}-test.img"
+
+mkdir -p /var/mnt
+
+# Also update kernel partition tables
+loopdev=$(losetup --find --show --partscan "${BOOTLOADER}-test.img")
+sleep 1
+
+mkfs.vfat "${loopdev}p1"
+mkfs.ext4 "${loopdev}p2"
+
+mount "${loopdev}p2" /var/mnt
+
+ESP="/var/mnt/efi"
+
+mkdir -p $ESP
+mount "${loopdev}p1" $ESP
+
+
+# Test installing the bootloader
+podman run --rm --net=host --privileged --pid=host \
+  --privileged \
+  --security-opt label=type:unconfined_t \
+  --env RUST_LOG=trace \
+  -v /dev:/dev \
+  -v /var/mnt:/var/mnt \
+  "$IMG_NAME" \
+  bootupctl backend install --bootloader "$BOOTLOADER" /var/mnt -vvvv
+
+# Make sure bootupd-state.json is in the esp
+test -f "$ESP/bootupd-state.json"
+
+cat "$ESP/bootupd-state.json" | jq
+
+version=$(cat "$ESP/bootupd-state.json" | jq -r ".installed.EFI.meta.version")
+
+if [[ $version != *shim* ]]; then echo "shim not found in version"; exit 1; fi
+if [[ $version != *"$EFI_DIR_NAME"* ]]; then echo "$BOOTLOADER not found in version"; exit 1; fi
+
+# Test if the correct binary has been installed
+actualShasum=$(podman run --rm "$IMG_NAME" find "/usr/lib/efi/$EFI_DIR_NAME" -type f -exec sha512sum {} + | awk '{print $1}')
+actualShasum="sha512:$actualShasum"
+
+if [[ $(uname -m) == "x86_64" ]]; then
+    grubName="grubx64.efi"
+else
+    grubName="grubaa64.efi"
+fi
+
+# TODO: Remove hardcoded "fedora" once we have support in centos
+storedShasum=$(cat "$ESP/bootupd-state.json" | jq -r --arg grub "$grubName" '.installed.EFI.filetree.children["fedora/\($grub)"].sha512')
+
+test "$actualShasum" == "$storedShasum"
+
+efiBinShasum=$(find "$ESP" -type f -name "$grubName" -exec sha512sum {} + | awk '{print $1}')
+efiBinShasum="sha512:$efiBinShasum"
+
+test "$efiBinShasum" == "$actualShasum"
+
+umount -Rl /var/mnt
--- a/scripts/test-bootloader.sh
+++ b/scripts/test-bootloader.sh
@@ -15,6 +15,10 @@ case $BOOTLOADER in
     grub)
         EFI_DIR_NAME=grub2
         ;;
+    *)
+        echo "Unknown bootloader $BOOTLOADER"
+        exit 1
+        ;;
 esac
 
 cat <<-EOF > sfdisk-buf
--- a/src/backend/statefile.rs
+++ b/src/backend/statefile.rs
@@ -1,11 +1,14 @@
 //! On-disk saved state.
 
 use crate::bootloader::Bootloader;
+use crate::bootupd::list_dev_current_root;
 use crate::efi::Efi;
 use crate::freezethaw::fsfreeze_thaw_cycle;
 use crate::model::SavedState;
 use crate::util::SignalTerminationGuard;
 use anyhow::{bail, Context, Result};
+use bootc_internal_blockdev::Device;
+use camino::Utf8PathBuf;
 use cap_std::ambient_authority;
 use cap_std::fs::{Dir, Permissions, PermissionsExt};
 use cap_std_ext::dirext::CapStdExtDirExt;
@@ -13,16 +16,59 @@ use fn_error_context::context;
 use fs2::FileExt;
 use std::fs::File;
 use std::io::prelude::*;
-use std::os::fd::{AsRawFd, FromRawFd};
 use std::path::Path;
-use tempfile::tempdir;
+
+fn parse_statefile(statusf: cap_std::fs::File) -> Result<Option<SavedState>> {
+    let mut bufr = std::io::BufReader::new(statusf);
+    let mut s = String::new();
+    bufr.read_to_string(&mut s)?;
+    let state: serde_json::Result<SavedState> = serde_json::from_str(s.as_str());
+
+    let r = match state {
+        Ok(s) => s,
+        Err(orig_err) => {
+            let state: serde_json::Result<crate::model_legacy::SavedState01> =
+                serde_json::from_str(s.as_str());
+            match state {
+                Ok(s) => s.upconvert(),
+                Err(_) => {
+                    return Err(orig_err.into());
+                }
+            }
+        }
+    };
+
+    Ok(Some(r))
+}
+
+/// lsblk: composefs:abc123..: not a block device
+/// is what lsblk throws on composefs booted systems if we try to
+/// get block devices using "/"
+///
+/// First, try to get the device from the `root` which is necessary
+/// during installs as we don't want (or can't) to open up /sysroot or /boot
+///
+/// If that fails, it means we're not on the install path so we get the
+/// device from checking mount point from /boot or /sysroot
+#[context("Getting parent device")]
+fn get_parent_device(root: &Dir) -> Result<Device> {
+    match bootc_internal_blockdev::list_dev_by_dir(root) {
+        Ok(d) => Ok(d),
+        Err(e) => {
+            // Not really an error just yet
+            log::debug!("{e:?}");
+            list_dev_current_root()
+        }
+    }
+}
 
 impl SavedState {
     /// System-wide bootupd write lock (relative to sysroot).
     const WRITE_LOCK_PATH: &'static str = "run/bootupd-lock";
     /// Top-level directory for statefile (relative to sysroot).
     pub(crate) const STATEFILE_DIR: &'static str = "boot";
-    /// On-disk bootloader statefile, akin to a tiny rpm/dpkg database, stored in `/boot`.
+    /// On-disk bootloader statefile, akin to a tiny rpm/dpkg database,
+    /// stored in `/boot` for Grub and in `ESP` for GrubCC
     pub(crate) const STATEFILE_NAME: &'static str = "bootupd-state.json";
 
     /// Try to acquire a system-wide lock to ensure non-conflicting state updates.
@@ -30,7 +76,10 @@ impl SavedState {
     /// While ordinarily the daemon runs as a systemd unit (which implicitly
     /// ensures a single instance) this is a double check against other
     /// execution paths.
-    pub(crate) fn acquire_write_lock(sysroot: Dir) -> Result<StateLockGuard> {
+    pub(crate) fn acquire_write_lock(
+        sysroot_path: Utf8PathBuf,
+        sysroot: Dir,
+    ) -> Result<StateLockGuard> {
         sysroot
             .atomic_write_with_perms(Self::WRITE_LOCK_PATH, "", Permissions::from_mode(0o644))
             .context("Creating lock file")?;
@@ -43,6 +92,7 @@ impl SavedState {
         lockfile.lock_exclusive().context("Acquiring lock")?;
 
         let guard = StateLockGuard {
+            sysroot_path,
             sysroot,
             termguard: Some(SignalTerminationGuard::new()?),
             lockfile: Some(lockfile),
@@ -52,8 +102,9 @@ impl SavedState {
 
     /// Use this for cases when the target root isn't booted, which is
     /// offline installs.
-    pub(crate) fn unlocked(sysroot: Dir) -> Result<StateLockGuard> {
+    pub(crate) fn unlocked(sysroot_path: Utf8PathBuf, sysroot: Dir) -> Result<StateLockGuard> {
         Ok(StateLockGuard {
+            sysroot_path,
             sysroot,
             termguard: None,
             lockfile: None,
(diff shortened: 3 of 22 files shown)


src/efi.rs
```diff
@@ -178,7 +184,7 @@ impl Efi {
         Ok(destdir)
     }
 
-    fn unmount(&self) -> Result<()> {
+    pub(crate) fn unmount(&self) -> Result<()> {
         if let Some(mount) = self.mountpoint.borrow_mut().take() {
```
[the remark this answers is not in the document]
Johan-Liebert1: Done

src/cli/bootupd.rs
```diff
@@ -90,6 +91,12 @@ pub struct GenerateOpts {
     /// Physical root mountpoint
     #[clap(value_parser)]
     sysroot: Option<String>,
+
+    /// The bootloader to generate metadata for
+    //
+    // We have a default to not break older systems
+    #[clap(long, default_value_t = Bootloader::Grub)]
+    bootloader: Bootloader,
```
cgwalters: Why wouldn't we detect this dynamically though? Like if grub-cc (and not grub) is installed then we just use it.

cgwalters: And if we make this change I think it simplifies a lot of the code here

Johan-Liebert1: On a call with @travier we had a discussion that, in the future, we'd probably have one single image with multiple bootloaders (grub, grub-cc, systemd-boot) and we'd want to switch between them. That being said, dynamic detection from the image does make sense if this cli option is `None`

travier: Yes, we want to be able to include all of the three options in a single image and select the one explicitly at install time and dynamically at update time.

cgwalters: I guess we should generally support:

- Static default in the code e.g. grub-cc > systemd-boot or whatever
- Choosing the default at build time `backend install` i.e. container build time; really this generalizes into a `/usr/lib/bootupd/updates.json` perhaps flag for `default: true` on one i.e. we have `RUN bootupctl backend set-default <entry>`
- Choosing the default at install time
- Switching across an update

Johan-Liebert1: > Choosing the default at build time backend install i.e. container build time; really this generalizes into a /usr/lib/bootupd/updates.json perhaps flag for default: true on one i.e. we have RUN bootupctl backend set-default <entry>

sgtm

src/backend/statefile.rs
```diff
@@ -12,20 +17,68 @@ use std::fs::File;
...
+
+    Ok(Some(r))
+}
+
+/// lsblk: composefs:abc123..: not a block device
+/// is what lsblk throws on composefs booted systems if we try to
+/// get block devices using "/"
+///
+/// First, try to get the device from the `root` which is necessary
+/// during installs as we don't want (or can't) to open up /sysroot or /boot
```
cgwalters, line 48: Hmm. This is is an interesting topic.

First of all it would probably make sense for us to have a more rigorous way to find the backing block devices behind the composefs - which is really "find backing filesystems behind overlayfs" in general.

src/backend/statefile.rs
```diff
@@ -12,20 +17,68 @@ use std::fs::File;
...
+/// If that fails, it means we're not on the install path so we get the
+/// device from checking mount point from /boot or /sysroot
+#[context("Getting parent device")]
+fn get_parent_device(root: &Dir) -> Result<Device> {
+    match bootc_internal_blockdev::list_dev_by_dir(root) {
+        Ok(d) => Ok(d),
+        Err(e) => {
+            // Not really an error just yet
+            log::debug!("{e:?}");
+            list_dev_current_root()
```
cgwalters: I think it'd be cleaner to more explicitly "peel" - if we detect that root is a composefs, then look for `sysroot` underneath that.

src/backend/statefile.rs
```diff
@@ -57,52 +112,104 @@ impl SavedState {
...
+
+                #[cfg(any(target_arch = "powerpc64", target_arch = "s390x"))]
+                Bootloader::GrubCC => {
+                    let arch = if cfg!(target_arch = "powerpc64") {
+                        "powerpc64"
+                    } else {
+                        "s390x"
+                    };
+
+                    anyhow::bail!("Only Grub is supported for {arch}");
```
cgwalters: I think it'd be cleaner to just deny using grub-cc on non-EFI architectures, we have a conditional for "is an EFI arch" I believe

src/backend/statefile.rs
```diff
@@ -57,52 +112,104 @@ impl SavedState {
...
+                }
+
+                #[cfg(any(
```
cgwalters: If we did ⬆️ then we could drop this arch list

src/backend/statefile.rs
```diff
@@ -123,6 +229,73 @@ impl StateLockGuard {
...
+        let all_esps = device
+            .find_colocated_esps()
+            .context("Searching for ESP")?
+            .ok_or_else(|| anyhow::anyhow!("ESP not found"))?;
+
+        let efi = Efi::default();
+
+        let serialized_state = serde_json::to_vec(state).context("Serializing state")?;
+
+        for esp in all_esps {
```
cgwalters, line 257: Can't we make just this code conditional and not the whole `update_state` function?

src/cli/bootupd.rs
```diff
@@ -113,6 +130,12 @@ impl DCommand {
 
     /// Runner for `install` verb.
     pub(crate) fn run_install(opts: InstallOpts) -> Result<()> {
+        if !matches!(opts.bootloader, Some(Bootloader::Grub) | None)
+            && cfg!(any(target_arch = "powerpc64", target_arch = "s390x"))
+        {
+            anyhow::bail!("Only Grub is supported for powerpc64 and s390x");
```
cgwalters: This is wrong btw (here and elsewhere), for s390x we only support zipl. So...we probably need to move that into the bootloader enum - which kind of overlaps with the components.

src/bootloader.rs
```diff
@@ -0,0 +1,91 @@
...
+use serde::{Deserialize, Serialize};
+use std::{fmt::Display, sync::OnceLock};
+
+#[derive(
+    Debug, Default, Copy, Clone, clap::ValueEnum, PartialEq, Eq, Hash, Serialize, Deserialize,
+)]
+pub enum Bootloader {
+    #[default]
+    Grub,
+    GrubCC,
```
cgwalters, line 10: Per previous comment, I think this overlaps with the "backend" concept and other PRs.

cgwalters: Sorry I know that's not the most helpful review comment. I know the codebase is already really complex, and basically some other PRs a bit like this have stalled out on trying to extend. Let's maybe have a chat about this?

src/cli/bootupd.rs
Johan-Liebert1: correct me if I'm wrong here but afaiu, we don't have proper support for ppcle and s390x in bootupd. We have some arch exclusions here and there, but we only properly support x86 and aarch

src/backend/statefile.rs
Johan-Liebert1: That's the approach I had at first, but the code ended up kinda ugly

src/backend/statefile.rs
Johan-Liebert1: `list_dev_current_root` looks for mountpoints for `/syrsoot` and `/boot`. But yeah, it's better to be explicit about this. I'll update this block

src/backend/statefile.rs
Johan-Liebert1: I ended up going with this approach inside of `build.rs`

```rust
fn main() {
    println!("cargo::rustc-check-cfg=cfg(efi_arch)");

    if cfg!(any(
        target_arch = "x86_64",
        target_arch = "aarch64",
        target_arch = "riscv64"
    )) {
        println!("cargo:rustc-cfg=efi_arch");
    }
}
```

I was thinking doing 

```rust
pub enum Bootloader {
    #[default]
    Grub,
    #[cfg(efi_arch)]
    GrubCC,
}

```

this does require adding `#[cfg(efi_arch)]` everywhere. 

I'm not sure which one's better. Compile time detection, or runtime errors

src/cli/bootupd.rs
cgwalters: In s390x zipl is always kind of UKI-like, so indeed there's really nothing to do in bootupd.

But we do use this project to install and update grub2 on ppc64le.

src/backend/statefile.rs
```diff
@@ -12,20 +17,70 @@ use std::fs::File;
 use std::io::prelude::*;
 use std::path::Path;
 
+fn parse_statefile(statusf: cap_std::fs::File) -> Result<Option<SavedState>> {
```
cgwalters, line 20: Could split this to a prep PR

Johan-Liebert1: Currently (in main) this logic is only used once. I believe splitting this into a separate PR will only introduce conflicts later on

src/cli/bootupd.rs
Johan-Liebert1: Ended up going with the `build.rs` approach and restrict GrubCC and SystemdBoot only for EFI systems

Johan-Liebert1:
I've tested `SystemdBoot` and `GrubCC` installation with bootc and it seems to work fine. Having this test in the CI would be helpful but we have a circular dependency between bootc and this PR

src/bootupd.rs
```diff
@@ -86,12 +90,74 @@ pub(crate) fn install(
...
+                if c.name() == "EFI" {
+                    efi_default = c.get_default_bootloader(&source_root_dir)?;
+
+                    if efi_default.is_none() {
+                        // We don't want to filter any bootloader
+                        efi_component_update = c.get_component_update(&source_root_dir, None)?;
+                    }
+                }
+
+                if c.name() == "BIOS" {
```
cgwalters: At this point we've *entirely* bypassed the component abstraction. Not fatal but probably argues for a rework...i.e. if we added a new component, this bit of code would break.

I guess I'd say at least for now let's make this an explicit `match` and error out on an unknown component?

src/bootloader.rs
```diff
@@ -0,0 +1,116 @@
...
+use fn_error_context::context;
+use serde::{Deserialize, Serialize};
+use std::fmt::Display;
+
+#[derive(
+    Debug, Default, Copy, Clone, clap::ValueEnum, PartialEq, Eq, Hash, Serialize, Deserialize,
+)]
+pub enum Bootloader {
+    #[default]
+    Grub,
```
cgwalters, line 8: I am still uncertain about calling this `Grub`. Isn't this really `Default`?

cgwalters (approved):

src/bootloader.rs
Johan-Liebert1: By "default" what comes to mind is - "install whatever's available" or whatever's the "default" in the metadata file. But, if there are multiple bootloaders present in an image, wouldn't choosing one by name be better?

src/bootupd.rs
Johan-Liebert1: Agreed on the match, but that won't completely solve this as we'll hit an error at runtime. How about an `enum Component` so we can have compile time check?

Johan-Liebert1: Ended up going with the enum approach

src/bootloader.rs
cgwalters: After live chat we think we can resolve this by just dropping the `#[default]`

Johan-Liebert1: Updated

Johan-Liebert1:
A lot of changes here. I'll test it out in fcos as an override

Dockerfile.bls
```diff
@@ -0,0 +1,125 @@
...
+# Create the expected directory structure at /usr/lib/efi/grub-cc/<evr>/EFI/fedora/
+mkdir -p "/usr/lib/efi/systemd-boot/${evr}/EFI/$ID"
+
+if [[ $(uname -m) == x86_64 ]]; then
+    grubName="grubx64.efi"
+else
+    grubName="grubaa64.efi"
+fi
+
+cp /usr/lib/systemd/boot/efi/systemd-boot*.efi "/usr/lib/efi/systemd-boot/${evr}/EFI/$ID/$grubName"
```
cgwalters, line 80: Hmmm, is this really required? I guess it must be - it's us taking over `bootctl install` and `bootctl update`.

That feels...unnecessary?

cgwalters: Now that I spend a bit of time on this, is this case actually "install systemd-boot pretending to be grub" so it gets chainloaded by shim? That looks like what this is.

That feels like a clearly distinct case of things here and really worth clearly documenting.

OK right, and we're doing the same with grub-cc?

Hmmm, I guess it's kind of an assumption woven into things here that we're using shim, which is probably the right one by default, but I think we do ultimately need to handle non-shim cases too.

I guess this isn't blocking per se, but we should have a note about this.

Johan-Liebert1: > Now that I spend a bit of time on this, is this case actually "install systemd-boot pretending to be grub" so it gets chainloaded by shim? That looks like what this is.

yes, that's what it is. We will probably hardlink the actual systemd-boot binary at `/usr/lib/efi/systemd-boot/**` so that all bootloader live in the same place

> OK right, and we're doing the same with grub-cc?

yes

> but I think we do ultimately need to handle non-shim cases too

We do. On a call with @travier we decided to handle that case later

travier: The name of the EFI binary loaded by shim is hardcoded in shim: https://src.fedoraproject.org/rpms/shim-unsigned-x64/blob/rawhide/f/shim-unsigned-x64.spec#_154

There are options to change it but they have a lot of downsides: https://github.com/coreos/bootupd/issues/1080#issuecomment-4543681753

Pretending to be `grub$arch.efi` is apparently the simplest option.

travier: We settled (at least for now) on leaving standalone systemd-boot (or potentially standalone grub-cc) support to systemd's bootctl as we don't need all the complexity from bootupd for those cases: https://github.com/coreos/bootupd/issues/1080#issuecomment-4741400502

scripts/test-bootloader.sh
```diff
@@ -0,0 +1,88 @@
...
+case $BOOTLOADER in
+    systemd)
+        EFI_DIR_NAME=systemd-boot
+        ;;
+    grub-cc)
+        EFI_DIR_NAME=grub-cc
+        ;;
+    grub)
+        EFI_DIR_NAME=grub2
+        ;;
```
Rolv-Apneseth, line 17: Should we have a failure for the default case?

src/backend/statefile.rs
Rolv-Apneseth, line 20: WDYT about doing something like:

```rust
fn parse_statefile(statusf: cap_std::fs::File) -> Result<SavedState> {
    let mut bufr = BufReader::new(statusf);
    let mut s = String::new();
    bufr.read_to_string(&mut s)?;
    let state: serde_json::Result<SavedState> = serde_json::from_str(s.as_str());

    match state {
        Ok(s) => Ok(s),
        Err(orig_err) => {
            let legacy_state: serde_json::Result<crate::model_legacy::SavedState01> =
                serde_json::from_str(s.as_str());
            legacy_state
                .map(crate::model_legacy::SavedState01::upconvert)
                .map_err(|_| orig_err.into())
        }
    }
}
```

And at the call sites:

```rust
root.open_optional(&path)
    .map_err(Into::into)
    .and_then(|o| o.map(parse_statefile).transpose())
```

src/backend/statefile.rs
```diff
@@ -57,52 +114,91 @@ impl SavedState {
...
+
+                    // Since we write the state file to all ESPs, it should be enough to get it
+                    // from the first one. Though, we could check the integrity by getting from
+                    // all the ESPs and making sure they're all the same...
+                    let esp = device.find_first_colocated_esp()?;
+
+                    // According to BLS, the ESP should be mounted at /boot or /boot/efi
+                    // which the following method already checks
+                    let mounted = efi
+                        .ensure_mounted_esp(Path::new("/"), Path::new(&esp.path()))
```
Rolv-Apneseth: Should this be using `root_path` in the case of an install to e.g. /var/mnt?

src/backend/statefile.rs
```diff
@@ -57,52 +114,91 @@ impl SavedState {
...
-        let statepath = Path::new(root_path.as_ref())
-            .join(Self::STATEFILE_DIR)
-            .join(Self::STATEFILE_NAME);
-        if statepath.exists() {
-            bail!("{} already exists", statepath.display());
+    pub(crate) fn ensure_not_present(
+        root_path: impl AsRef<Path>,
+        bootloader: Bootloader,
+    ) -> Result<()> {
+        let saved_state = SavedState::load_from_disk(&root_path, Some(bootloader))?;
```
Rolv-Apneseth, line 174: `load_from_disk` calls `device.find_first_colocated_esp()` for grub-cc and systemd-boot - do we need to handle the case where no ESP is found?

src/backend/statefile.rs
```diff
@@ -123,6 +218,64 @@ impl StateLockGuard {
...
+                SavedState::STATEFILE_NAME,
+                &serialized_state,
+                Permissions::from_mode(0o644),
+            )
+            .context("Writing state file")?;
+
+            // Do the sync before unmount
+            fsfreeze_thaw_cycle(dir.reopen_as_ownedfd()?)?;
+            drop(dir);
+            efi.unmount().context("unmount after update")?;
```
Rolv-Apneseth: Is it intended that this unmounts even when `was_mounted = true` due to how `Efi::unmount` sets `was_mounted = false`?

src/bootloader.rs
```diff
@@ -0,0 +1,113 @@
...
+    let bootloader = match get_loader_info() {
+        Some(info) => {
+            if info.to_lowercase().contains("grub cc") {
+                Bootloader::GrubCC
+            } else if info.to_lowercase().contains("systemd-boot") {
+                Bootloader::Systemd
+            } else {
+                Bootloader::Grub
+            }
+        }
```
Rolv-Apneseth:
```suggestion
        Some(info) => match info.to_lowercase() {
            i if i.contains("grub cc") => Bootloader::GrubCC,
            i if i.contains("systemd-boot") => Bootloader::Systemd,
            _ => Bootloader::Grub,
        },
```

src/bootupd.rs
```diff
@@ -86,24 +90,101 @@ pub(crate) fn install(
...
+
+            match (bios_default, efi_default) {
+                // EFI bootloader takes precedence
+                // Take the following example
+                //  - BIOS default = Grub (always)
+                //  - EFI default = GrubCC
+                //
+                //  We can't install GrubCC for BIOS as it's not supported
+                //  So we just default to installing GrubCC
+                (Some(_), Some(eb)) => eb,
```
Rolv-Apneseth:
```suggestion
                (_, Some(eb)) => eb,
```

src/model.rs
```diff
@@ -33,6 +38,51 @@ impl ContentMetadata {
             compare_package_versions(&self.version, &target.version)
         }
     }
+
+    /// Returns bootloaders are available for install
+    #[cfg(efi_arch)]
+    pub(crate) fn num_bootloader_available(&self) -> Vec<Bootloader> {
```
Rolv-Apneseth: I think this should be renamed since it doesn't actually return a number

src/model.rs
```diff
@@ -33,6 +38,51 @@ impl ContentMetadata {
...
+
+        return available;
+    }
+
+    #[cfg(efi_arch)]
+    pub(crate) fn bootloader_available(&mut self, bootloader: Bootloader) -> bool {
+        self.version
+            .split(",")
+            .any(|v| v.starts_with(bootloader.efi_component_name()))
+    }
```
Rolv-Apneseth, line 63:
```suggestion
    #[cfg(efi_arch)]
    pub(crate) fn is_bootloader_available(&self, bootloader: Bootloader) -> bool {
        self.version
            .split(',')
            .any(|v| v.starts_with(bootloader.efi_component_name()))
    }
```

Dockerfile.bls
```diff
@@ -0,0 +1,125 @@
...
+# Install systemd-boot
+RUN --mount=type=tmpfs,target=/run --mount=type=tmpfs,target=/tmp <<EOF
+set -eux
+# epoch for systemd-boot-unsigned is "(none)"
+evr=$(rpm -q --qf '%{EPOCH}:%{VERSION}-%{RELEASE}\n' systemd-boot-unsigned)
+rpm -ql systemd-boot-unsigned
+
+. /etc/os-release
+
+# Create the expected directory structure at /usr/lib/efi/grub-cc/<evr>/EFI/fedora/
```
Rolv-Apneseth: It looks like this was meant to be the systemd-boot path

Rolv-Apneseth (commented):
Couple of questions and suggestions but wow, this is cool

scripts/test-bootloader.sh
Johan-Liebert1: Added

src/backend/statefile.rs
Johan-Liebert1: Yes, that's a nice catch

Dockerfile.bls
Johan-Liebert1: Updated

src/bootupd.rs
Johan-Liebert1: Can't do that as that'll also match the (None, Some(eb)) below

src/backend/statefile.rs
Johan-Liebert1: That's handled by `find_first_colocated_esp`. It will throw an error if no ESP is found

src/backend/statefile.rs
Johan-Liebert1: Maybe it's just me but I don't like chaining functions in Rust. Gets confusing real fast. Not against it though, if you feel this is better, then we can have it

src/backend/statefile.rs
Johan-Liebert1: Ah, forgot this one. Updated

src/backend/statefile.rs
Rolv-Apneseth: Fair enough, just my preference then. You don't even want to drop the `Option` which is only ever `Some` in the function?

Rolv-Apneseth: Feel free to do it how you prefer though

src/backend/statefile.rs
Rolv-Apneseth: I guess the question is more _should_ it throw an error if this is `ensure_not_present` - though I may just be misunderstanding the purpose of the method.

src/bootupd.rs
Rolv-Apneseth: Ah, the selected range for the suggestion was inaccurate, sorry. I mean replace:

```rust
                (Some(_), Some(eb)) => eb,
                (None, Some(eb)) => eb,
```

With

```rust
                (_, Some(eb)) => eb,
```

Johan-Liebert1: oh, whoops, that definitely can be done

src/backend/statefile.rs
Johan-Liebert1: I see your approach is cleaner. 

The reason for `Ok(Some(state))` is because in other functions where we're getting the state from disk, it might or might not exist so those functions return `Result<Option<T>>`. This requires updating some functions here and there. 

This PR is already kinda huge, would you be fine with this as a followup?

Rolv-Apneseth: Of course yes, and it's not necessary anyway.

Is any fixup of the commits required or are you happy with them as is?

Johan-Liebert1:
> A lot of changes here. I'll test it out in fcos as an override

Tested locally overriding rootfs. There's one test failure. Looking into it 

```
systemctl status kola-runext.service:
× kola-runext.service
     Loaded: loaded (/etc/systemd/system/kola-runext.service; static)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: failed (Result: exit-code) since Mon 2026-07-13 05:42:18 UTC; 916ms ago
   Duration: 190ms
 Invocation: 760c18b9ef0b4f7681b94504d3ac32cb
    Process: 2350 ExecStart=/usr/local/bin/kola-runext-install-bootloader-multipath (code=exited, status=1/FAILURE)
   Main PID: 2350 (code=exited, status=1/FAILURE)
   Mem peak: 3.6M
        CPU: 101ms

Jul 13 05:42:18 qemu0 systemd[1]: Started kola-runext.service.
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + . /var/opt/kola/extdata/commonlib.sh
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: ++ IFS=' '
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: ++ read -r -a cmdline
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + case "${AUTOPKGTEST_REBOOT_MARK:-}" in
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + ok 'first boot'
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + echo ok 'first boot'
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: ok first boot
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2352]: + cat /proc/cmdline
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2353]: + grep -q rd.multipath=default
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2354]: ++ findmnt -nvr -o SOURCE /boot
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + boot_src=/dev/mapper/0x21017c2d5487beb8p3
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2355]: ++ multipath -l -v 1
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + parent_device=/dev/mapper/0x21017c2d5487beb8
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2356]: + udevadm info --query=property /dev/mapper/0x21017c2d5487beb8
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2357]: + grep -q MPATH_DEVICE_READY=1
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + mount -o remount,rw /boot
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + rm -f /boot/bootupd-state.json
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2350]: + bootupctl backend install --auto --write-uuid --device /dev/mapper/0x21017c2d5487beb8 --update-firmware /
Jul 13 05:42:18 qemu0 kola-runext-install-bootloader-multipath[2361]: error: boot data installation failed: Could not determine bootloader. Default bootloader not set
Jul 13 05:42:18 qemu0 systemd[1]: kola-runext.service: Main process exited, code=exited, status=1/FAILURE
Jul 13 05:42:18 qemu0 systemd[1]: kola-runext.service: Failed with result 'exit-code'.
--- FAIL: ext.config.boot.install-bootloader-multipath (27.58s)
        harness.go:1259: kolet failed: : kolet run-test-unit failed:  Error: Unit kola-runext.service exited with code 1
2026-07-13T05:42:19Z cli: Unit kola-runext.service exited with code 1: Process exited with status 1
```

src/backend/statefile.rs
Johan-Liebert1: I think the commits are fine. Let me know if you think they require any changes. We have a default to creating a merge commit for this repo anyway and not rebase

Rolv-Apneseth: Ah, forgot this one is different. Yeah looking good then

Johan-Liebert1:
Fixed. Issue was to do with BIOS having no default bootloader. Explicitly set default to Grub

Rolv-Apneseth (approved):
LGTM

PR #1116, opened 2026-06-21
Make package system independent

krolmiki2011:
I created a new packagesystem (I didnt write a manifest generator yet), i only want  to it was checked whether to accept,
and i copied old in packagesystem_legacy.rs


krolmiki2011:
Hello, can someone make /ok-to-test?

PR #1124, opened 2026-07-15
coreos: generalize to any aleph created by bootc

jbtrystram:
Rename the `coreos` module to `aleph` as `bootc install` will create an aleph file in all cases except the composefs path.

There is nothing specific to coreOS anymore in this path.

See https://github.com/bootc-dev/bootc/commit/c6112d2d2fa3b785c858741f1585b0578088eda2 And https://github.com/coreos/fedora-coreos-tracker/issues/2187


src/aleph.rs
```diff
@@ -0,0 +1,200 @@
...
+pub(crate) fn get_aleph_version(root: &Path) -> Result<Option<AlephWithTimestamp>> {
+    for aleph_path in ALEPH_PATHS {
+        let path = &root.join(aleph_path);
+        if !path.exists() {
+            continue;
+        }
+        let statusf = File::open(path).with_context(|| format!("Opening {path:?}"))?;
+        let meta = statusf.metadata()?;
+        let bufr = std::io::BufReader::new(statusf);
+        let aleph: Aleph = serde_json::from_reader(bufr)?;
```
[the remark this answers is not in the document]
Johan-Liebert1: This sounds like a good suggestion

src/aleph.rs
```diff
@@ -0,0 +1,200 @@
...
+        log::debug!("Found aleph version in {aleph_path}");
+        return Ok(Some(AlephWithTimestamp {
```
Johan-Liebert1, line 51: We're returning as soon as we find a file. Are we sure we won't have one in both the paths?

Johan-Liebert1 (commented):
One comment else LGTM

Johan-Liebert1:
Also needs `cargo fmt`

src/aleph.rs
jbtrystram: If both exists, the second will be a symlink to the bootc one anyway.

When this ships, we will drop the compat symlink

jbtrystram:
The jenkins test here will be fixed with https://github.com/coreos/fedora-coreos-config/pull/4261

Johan-Liebert1 (approved):

cgwalters:
I think this is fine, but...probably something got missed here in that the aleph stuff was only added to bootupd precisely because before bootupd existed, the aleph data had some (indirect) information about the bootloader state.

No system installed that has a *bootc* aleph will be missing bootupd data (except those using systemd-boot directly, but those don't have bootupd installed either!).

PR #1126, opened 2026-07-22
esp: TempMount ESP if not already mounted

Johan-Liebert1:
In the composefs Grub case we have `/sysroot/boot` and config related to Grub is stored in the directory. We were masking `/sysroot/boot` with an ESP mount which caused bootupd to write incorrect UUID (the ESP's) in `bootuuid.cfg` file rendering the system unbootable.

Fix this by mounting the ESP at a temp location if we don't find it already mounted at predetermined locations

---

component: Add `as_any` method for downcasting

There are a few places where the `Component` trait turns out to be too
generic and needs downcasting. Add a `as_any` method to allow the
downcast

---

Fix grub static config install

Accept the mounted ESP dir in `grubconfigs::install` so as to not assume
the ESP is mounted at `boot/efi`

### DIFF
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -350,6 +350,10 @@ fn skip_systemd_bootloaders() -> bool {
 }
 
 impl Component for Efi {
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
+
     fn name(&self) -> &'static str {
         self.component_type().into()
     }
--- a/src/efi.rs
+++ b/src/efi.rs
@@ -187,6 +187,11 @@ impl Efi {
         }
     }
 
+    /// Returns the ESP's mountpoint if present
+    pub(crate) fn get_esp_mountpoint(&self) -> std::cell::Ref<'_, Option<PathBuf>> {
+        self.mountpoint.borrow()
+    }
+
     /// Checks if we have already mounted the ESP
     /// If not, checks if the ESP is mounted at common mountpoints, i.e. /boot, /boot/efi
     /// Else, mounts the ESP at a temp path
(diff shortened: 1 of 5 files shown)


cgwalters (approved):
OK, makes sense to me.

However, one clear thing here is we should still probably push for adding `/efi` into our base images, because then this stuff becomes clearer.

Also that said: can't we just open the mount as a fd using the new mount api? I'm not aware of a strong reason for us to attach it to the namespace at all - and that would entirely avoid these bugs.

Johan-Liebert1:
> can't we just open the mount as a fd using the new mount api? I'm not aware of a strong reason for us to attach it to the namespace at all

This would be ideal, but would require a complete restructuring of the codebase as we expect `PathBuf` in a lot of places

src/efi.rs
```diff
@@ -318,6 +355,10 @@ fn skip_systemd_bootloaders() -> bool {
 }
 
 impl Component for Efi {
+    fn as_any(&self) -> &dyn std::any::Any {
```
cgwalters, line 338: I'm OK with this, but what I think would be cleaner in the future is if we moved the mountpoint holder outside of this struct - basically the grubconfigs stuff should directly use the efi module or so

src/efi.rs
```diff
@@ -185,14 +209,27 @@ impl Efi {
...
         *self.was_mounted.borrow_mut() = false;
+
         if let Some(mount) = self.mountpoint.borrow_mut().take() {
+            if let Some(tmpmount) = self.tempmount_guard.borrow_mut().take() {
+                drop(tmpmount);
+                log::trace!("Dropped tempmount");
+                return Ok(());
+            }
+
             Command::new("umount")
```
cgwalters: Shouldn't TempMount handle this now?

src/efi.rs
Johan-Liebert1: Yup, that's definitely better

src/efi.rs
Johan-Liebert1: Right, I'd missed that. Updated

PR #1130, opened 2026-07-27
Drop RPM dependency and move system packaging concerns to external scripts

krolmiki2011:
Thanks to these changes, it will be possible to eliminate the rpm dependency from bootupd and become more distribution-independent!

KEY CHANGES:
BIOS:
In the case of BIOS, ContentMetadata is created just as it is for UEFI; for BIOS, `grub2-install --version` and the `mtime` are used.
EFI (ostree-boot):
The situation is similar for ostree-boot in BIOS mode, except it iterates through files, and the version format is: legacy-ostree-boot-{mtime}

NOTE:
The old code, before the modification, is in the {bios, efi, packagesystem}_legacy.rs

Oh, and 'grub2-install --version' need package maintainers to patch grub2 upstream version to package version

This patch was inspired by issue: #468


krolmiki2011:
hello can someone help me, because tests not working

krolmiki2011:
Hello, its anyone here?

krolmiki2011:
Hello?

krolmiki2011:
Is there anyone here?

Rolv-Apneseth:
Hi @krolmiki2011. We have reduced capacity for going through `bootupd` contributions at the moment, but when I get the chance I'll try to look this over.

From a very brief glance, why keep the legacy code files? Investigating the failing CI will take a more thorough investigation, but it does seem related to these changes. Are we maybe missing something to filter out only bootloader components here?

Rolv-Apneseth:
One more point, though again it may just require a deeper look from my part, but https://github.com/coreos/bootupd/issues/468#issuecomment-4741389507 would lead me to believe this should be possible without making changes to files other than `packagesystem.rs`. Could you give a brief explanation of why that wasn't possible for this approach?

krolmiki2011:
> One more point, though again it may just require a deeper look from my part, but [#468 (comment)](https://github.com/coreos/bootupd/issues/468#issuecomment-4741389507) would lead me to believe this should be possible without making changes to files other than `packagesystem.rs`. Could you give a brief explanation of why that wasn't possible for this approach?

Okay, so I view this notebook as a place to keep things in case the code turns into spaghetti, but also as something to delete once the code is stable.

Regarding your second question—a good point—I noticed that BIOS and Legacy EFI (OSTree boot) both use RPM; while BIOS is handled by `query_bios_grub()` in `packagesystem.rs`, Legacy EFI isn't. I could certainly implement a similar function in `packagesystem` if you'd like.

Rolv-Apneseth:
So, trying to understand this a bit better, as TBH I'm still learning about `bootupd` and how it works:

We want to make `bootupd` more distro-agnostic, so we want to remove the requirement for `rpm`. This is currently used to get the version+build time of the files tracked by `bootupd` (to know if an update is needed). And, this is only the case for legacy EFI and BIOS implementations, since the newer EFI path parses version info from the directory structure (`/usr/lib/efi/<name>/<version>/EFI` - only implemented on Fedora 44+ though). The version(s) returned by `rpm` for each package (e.g. `grub2-tools-1:2.12-64.fc44.x86_64`) is persisted on existing systems for later comparison. The timestamp also gets persisted but I don't actually see that being used anywhere.

Quick overview of the data that gets stored:

<details>
<summary><b>ContentMetadata struct</b></summary>

```rust
pub(crate) struct ContentMetadata {
    /// The timestamp, which is used to determine update availability
    pub(crate) timestamp: DateTime<Utc>,
    /// Human readable version number, like ostree it is not ever parsed, just displayed
    pub(crate) version: String,
    /// Transfer version into Module struct list
    pub(crate) versions: Option<Vec<Module>>,
    /// The default bootloader to install if at install time no bootloader option is
    /// provided
    #[cfg(efi_arch)]
    pub(crate) default_bootloader: Option<Bootloader>,
}
```

Note that the descriptions for `timestamp` and `version` are out of date - `timestamp` appears unused, and `version` _is_ used as a legacy fallback. `versions` is used when available.

</details>

<details>
<summary><b>Example bootupd-state.json</b></summary>

```json
{
  "installed": {
    "BIOS": {
      "meta": {
        "timestamp": "2026-06-09T16:53:24Z",
        "version": "grub2-tools-1:2.12-60.fc44.x86_64",
        "versions": [
          {
            "name": "grub2",
            "rpm_evr": "1:2.12-60.fc44"
          }
        ]
      },
      "filetree": null,
      "adopted-from": null
    },
    "EFI": {
      "meta": {
        "timestamp": "2026-08-13T14:37:10.461466856Z",
        "version": "grub2-1:2.12-60.fc44,shim-16.1-5",
        "versions": [
          {
            "name": "grub2",
            "rpm_evr": "1:2.12-60.fc44"
          },
          {
            "name": "shim",
            "rpm_evr": "16.1-5"
          }
        ]
      },
      "filetree": {
        "children": {
          "BOOT/BOOTX64.EFI": {
            "source": "shim/16.1-5/EFI/BOOT/BOOTX64.EFI",
            "size": 1026520,
            "sha512": "sha512:0dc3725da36f3183b5cb5af0ba982caccc35019b35f8c80ee29545b7f9fa0672aa09aac4f5693f250ca603910aa4249538d45991595cae198f897ee2f406bb27"
          },
          "BOOT/fbx64.efi": {
            "source": "shim/16.1-5/EFI/BOOT/fbx64.efi",
            "size": 119280,
            "sha512": "sha512:46bf07b2212b2042f2c3eb44b0fc94527443ebd405204fd09014c55b4f3fd5d590007d78a5662f79570cfdb90e5c8ed5d794d7fa85ed8914d4c60b1d6ee9441d"
          },
          "fedora/BOOTX64.CSV": {
            "source": "shim/16.1-5/EFI/fedora/BOOTX64.CSV",
            "size": 110,
            "sha512": "sha512:0c29b8ae73171ef683ba690069c1bae711e130a084a81169af33a83dfbae4e07d909c2482dbe89a96ab26e171f17c53f1de8cb13d558bc1535412ff8accf253f"
          },
          "fedora/grubx64.efi": {
            "source": "grub2/1:2.12-60.fc44/EFI/fedora/grubx64.efi",
            "size": 4145576,
            "sha512": "sha512:1f86c5f4824cf292a9e36314186309977a8d92834c48b70ac128f7ec2289a544acddce16ea1b7d2457426cdd2fd5dfc65beab727f171c2fe443f8fc4ba19d684"
          },
          "fedora/mmx64.efi": {
            "source": "shim/16.1-5/EFI/fedora/mmx64.efi",
            "size": 874352,
            "sha512": "sha512:d63aafcab70aeedcf1e083fda32130d9ba997ad04f8466c0d058f5d364a48d554b7160c0a4f17441f19797d39dfd785516078d3bd841714df2c2aa3327f3fa9f"
          },
          "fedora/shim.efi": {
            "source": "shim/16.1-5/EFI/fedora/shim.efi",
            "size": 1026520,
            "sha512": "sha512:0dc3725da36f3183b5cb5af0ba982caccc35019b35f8c80ee29545b7f9fa0672aa09aac4f5693f250ca603910aa4249538d45991595cae198f897ee2f406bb27"
          },
          "fedora/shimx64.efi": {
            "source": "shim/16.1-5/EFI/fedora/shimx64.efi",
            "size": 1026520,
            "sha512": "sha512:0dc3725da36f3183b5cb5af0ba982caccc35019b35f8c80ee29545b7f9fa0672aa09aac4f5693f250ca603910aa4249538d45991595cae198f897ee2f406bb27"
          }
        }
      },
      "adopted-from": null
    }
  },
  "pending": null,
  "static-configs": {
    "timestamp": "1970-01-01T00:00:00Z",
    "version": "0.2.35",
    "versions": null
  }
}
```

Side note, but this also shows that the timestamps are inconsistent - BIOS is giving the build time of the RPM, whereas EFI (which I guess is using the filetree of `/usr/lib/efi`) is giving the time the state file was generated.

</details>

---

This patch is currently changing the approach for BIOS to directly query and parse `/usr/sbin/grub2-install --version`, which you admit would require extra work to actually return the output we're looking for (and not ignore patch-level version bumps):

> Oh, and 'grub2-install --version' need package maintainers to patch grub2 upstream version to package version

Since currently this returns something like `grub2-install (GRUB) 2.12` and we parse out `2.12` by splitting white space. However, I don't think that's a realistic expectation from package maintainers, and maybe I'm wrong but I feel like it doesn't make much sense to change a tool's output like that.

For the `ostree-boot` EFI path, which used to query all files under `/usr/lib/ostree-boot/efi/EFI` with `rpm` to find `grub` and `shim` versions, this patch instead creates a synthetic version with `"rpm_evr": "legacy-ostree-boot-{SystemTime::now()}"`, which 1. loses per-package breakdowns and 2. AFAICT would always then be considered update-able.

Worth also noting that the `ostree-boot` EFI path is what any system that doesn't have the `usr/lib/efi/<name>/<version>/EFI` layout would use, so that seems like the important one to do well. Currently, any system without `/usr/lib/ostree-boot` or `/usr/lib/efi` [will fail](https://github.com/coreos/bootupd/blob/main/src/efi.rs#L667), so some work would still be needed to support other distros that don't use `ostree` (if that's planned).

---

I think the first approach I would have thought of for this is to just parse version outputs from the main package managers (e.g. `rpm`, `apt`, `pacman`), finding whatever is installed on the system, and use those for versioning. But I believe the suggestion from @cgwalters (correct me if I'm wrong) was to not have this done in Rust, but rather have the base image provide a common script (e.g. `get-package-version`) which we could just call, shifting the burden of figuring out what package manager command is required out of `bootupd`.

Another approach that pops to mind is to use something like file hashes for versioning instead. So equal hashes of the file content means no update, not equal means update. That loses downgrade detection, but maybe that's fine for a bootloader updater. The EFI filetree already has hashes for each file, and for BIOS we could just hash the grub binary instead? I'm sure there's issues that I'm not foreseeing with this though.

The script approach probably means the least amount of work and changes for `bootupd`, and it could probably also be used when actually building the path layouts in `/usr/lib/efi` for the images that implement it. I don't have the context on whether we expect other distros to implement <https://fedoraproject.org/wiki/Changes/BootLoaderUpdatesPhase1>.

---

And thanks for working on this @krolmiki2011. I'd say let's decide on a solution first before we continue iterating. I'll try to follow up with others to see how to proceed, and maybe bring it up in a community meeting.

krolmiki2011:
@Rolv-Apneseth thanks for the reply and comment, and i kinda noticed, that grub2-install --version in fedora is 2.12, so less of topic i made pull request: https://src.fedoraproject.org/rpms/grub2/pull-request/246 to fix version

krolmiki2011:
Oh, and i open for another solutions

src/packagesystem.rs
```diff
@@ -37,73 +37,84 @@ impl PartialOrd for Module {
...
-        .collect::<Result<BTreeMap<&str, DateTime<Utc>>>>()?;
-    if pkgs.is_empty() {
-        bail!("Failed to find any RPM packages matching files in source efidir");
+/// Query bios using no rpm (grub2-install --version, and timestamp)
+pub(crate) fn query_bios_grub(grub_bin: &str) -> Result<ContentMetadata> {
+    use std::process::Command;
+    let grub_bin_path = Path::new(grub_bin);
+
+    let grub_version_output = Command::new(&grub_bin_path)
+        .arg("--version")
```
cgwalters: Per discussion in the issue I think this is a bit too naive by default. It doesn't handle shim either.

One thing that would clearly help is to make the rpm stuff a build-time feature so it's more pluggable.

We can fall back to `--version` style stuff if the rpm feature is off.

Rolv-Apneseth: >  It doesn't handle shim either.

Does BIOS use shim?

> One thing that would clearly help is to make the rpm stuff a build-time feature so it's more pluggable.

Could you explain why? I don't follow. Would we still be using the fallbacks if the feature is off?

Should we implement a script at the base image level which can handle calling the right package manager to get versions for `bootupd`? I believe that was your original suggestion, but correct me if I'm wrong.

Do we expect everyone to implement the `usr/lib/efi/<name>/<version>/EFI` convention?

krolmiki2011: > > It doesn't handle shim either.
> 
> Does BIOS use shim?
> 
> > One thing that would clearly help is to make the rpm stuff a build-time feature so it's more pluggable.
> 
> Could you explain why? I don't follow. Would we still be using the fallbacks if the feature is off?
> 
> Should we implement a script at the base image level which can handle calling the right package manager to get versions for `bootupd`? I believe that was your original suggestion, but correct me if I'm wrong.
> 
> Do we expect everyone to implement the `usr/lib/efi/<name>/<version>/EFI` convention?

Umm, i rather use script, i already made a concept: https://github.com/coreos/bootupd/issues/468#issuecomment-5352312115

Rolv-Apneseth: Apologies, missed the conversation on the issue

krolmiki2011: > Apologies, missed the conversation on the issue

@Rolv-Apneseth What do you think about my concept? https://github.com/coreos/bootupd/issues/468#issuecomment-5352312115

krolmiki2011:
@Rolv-Apneseth I made a script to begin with script is in packagesystem/query_file_owner

krolmiki2011:
@Rolv-Apneseth i think i replace rpm with script, what do you think?

src/packagesystem.rs
krolmiki2011: Yea, i updated the entire system with the script

krolmiki2011:
@Rolv-Apneseth i guess my code is ready to test

krolmiki2011:
Test needs to be fixed btw!

packagesystem/query-file-owner
```diff
@@ -0,0 +1,31 @@
...
+for FILE in "$@"; do
+    if [ ! -e "$FILE" ]; then
+        echo "Error: File '$FILE' does not exist."
+        exit 1
+    fi
+
+    # Use your package manager to find the package that owns the file, we use rpm for example
+    # Package Manager should return something like this: NAME-EPOCH(which is optional):VERSION-RELEASE.ARCH
+    # NOTE: Remember that every package manager has other method of output, so you need to personalize
+    PACKAGE=$(rpm -q --queryformat '%{name}-%|EPOCH?{%{epoch}:}:{}|%{version}-%{release}.%{arch}\n' -f "$FILE")
```
Rolv-Apneseth: I think let's just use whitespace as the delimiter so it's easier to split (since we're doing it manually):

```sh
rpm -q --qf '%{NAME} %{EVR}\n' -f "$FILE"
```

And we need reference implementations for `dpkg` and `pacman` too. Tested very briefly but maybe something like:

```sh
dpkg-query -W -f '${Package} ${Version}\n' "$(dpkg -S "$FILE" | cut -d: -f1)"
pacman -Q "$(pacman -Qoq "$FILE")"
```

Then we need to update the parsing logic accordingly. WDYT?

packagesystem/query-file-owner
```diff
@@ -0,0 +1,31 @@
...
+    echo $PACKAGE
```
Rolv-Apneseth: I'd say just let the command output to STDOUT instead of going through `echo`

src/packagesystem.rs
```diff
@@ -37,73 +37,78 @@ impl PartialOrd for Module {
...
-    if pkgs.is_empty() {
-        bail!("Failed to find any RPM packages matching files in source efidir");
-    }
-    let timestamps: BTreeSet<&DateTime<Utc>> = pkgs.values().collect();
-    // Unwrap safety: We validated pkgs has at least one value above
-    let largest_timestamp = timestamps.iter().last().unwrap();
-    let version = pkgs.keys().fold("".to_string(), |mut s, n| {
-        if !s.is_empty() {
-            s.push(',');
+/// Parse the output of `query-file-owner`.
```
Rolv-Apneseth: We should specify it's a script when mentioning it so anyone reading this knows where to look.

src/packagesystem.rs
```diff
@@ -37,73 +37,78 @@ impl PartialOrd for Module {
...
-    // Map the version into Module struct
-    let mut modules_vec: Vec<Module> = pkgs.keys().map(|pkg_str| parse_evr(pkg_str)).collect();
+        packages.insert(package);
+    }
+
+    if packages.is_empty() {
+        bail!("Failed to find any packages matching files");
+    }
+
+    let version = packages.iter().cloned().collect::<Vec<_>>().join(",");
```
Rolv-Apneseth:
```suggestion
    let version = packages.iter().copied().collect::<Vec<_>>().join(",");
```

Cargo.toml
```diff
@@ -40,7 +40,7 @@ cap-std = "4.0.2"
 openssl = "^0.10"
 os-release = "0.1.0"
 regex = "1.12.3"
-rpm-version = { version = "0.5.0", default-features = false, optional = true }
+rpm-rs = { package = "rpm", version = "0.16.1", default-features = false, optional = true }
```
Rolv-Apneseth: Leftover from testing or maybe a rebase? We may be able to drop the `rpm` feature entirely if it's not used anywhere. I think we only need `rpm-version` if it is used though.

Makefile
```diff
@@ -26,6 +26,7 @@ all:
 .PHONY: install
 install:
 	mkdir -p "${DESTDIR}$(PREFIX)/bin" "${DESTDIR}$(LIBEXECDIR)"
+	install -D -m 755 packagesystem/query-file-owner "${DESTDIR}/lib/bootupd/packagesystem/query-file-owner"
```
Rolv-Apneseth: If we're gonna have multiple reference implementations, we shouldn't install the script in the Makefile. The spec should be updated to install the `rpm` implementation though.

Also, side note but this would need to be `${DESDIR}${PREFIX}/lib/bootupd/packagesystem/query-file-owner`.

Rolv-Apneseth (commented):
Thanks for moving to the script approach. I have a couple of notes but WDYT @cgwalters

Makefile
Rolv-Apneseth: Ah - but CI would expect the script and that runs `make install`. I guess we still need it in the Makefile...

krolmiki2011:
I made changes, but i have a question @Rolv-Apneseth , should EFI timestamp be: `Utc::now() `?

Rolv-Apneseth:
> I made changes, but i have a question @Rolv-Apneseth , should EFI timestamp be: `Utc::now() `?

I think it should match whatever's being done for the `/usr/lib/efi` path, but it's not too important either way

krolmiki2011:
@Rolv-Apneseth yea, btw i made changes you proposed

Rolv-Apneseth:
> > I made changes, but i have a question @Rolv-Apneseth , should EFI timestamp be: `Utc::now() `?
> 
> I think it should match whatever's being done for the `/usr/lib/efi` path, but it's not too important either way

@krolmiki2011 Ok, I wasn't aware of https://github.com/coreos/bootupd/issues/1075 and https://reproducible-builds.org/docs/source-date-epoch. Please restore `get_metadata_timestamp` and use it in both locations.

krolmiki2011:
> > > I made changes, but i have a question @Rolv-Apneseth , should EFI timestamp be: `Utc::now() `?
> > 
> > I think it should match whatever's being done for the `/usr/lib/efi` path, but it's not too important either way
> 
> @krolmiki2011 Ok, I wasn't aware of https://github.com/coreos/bootupd/issues/1075 and https://reproducible-builds.org/docs/source-date-epoch. Please restore `get_metadata_timestamp` and use it in both locations.

Wait, and bios too? @Rolv-Apneseth

Rolv-Apneseth:
> Wait, and bios too? @Rolv-Apneseth

Just have `get_metadata_timestamp` handle all paths, yeah (so call it in `parse_package_metadata`). Internally it'll get the current timestamp if `SOURCE_DATE_EPOCH` isn't defined.

packagesystem/query-file-owner
```diff
@@ -0,0 +1,39 @@
...
+for FILE in "$@"; do
+    if [ ! -e "$FILE" ]; then
+        echo "Error: File '$FILE' does not exist."
+        exit 1
+    fi
+
+    # Use your package manager to find the package that owns the file
+    # Package Manager should return two space-separated values: NAME and VERSION
+    # Different distros use different package managers, so we check what's available
+    if command -v rpm &> /dev/null; then
```
Rolv-Apneseth: So, I think the ideal approach is probably to split it out into separate scripts so `command -v` isn't required, even though it does introduce a bit more work for packaging. However, if we do go with this approach, the check should at least be done outside of this loop. Also, if there's no matches there should be an error.

src/packagesystem.rs
```diff
@@ -37,113 +37,83 @@ impl PartialOrd for Module {
...
+    if packages.is_empty() {
+        bail!("Failed to find any packages matching files");
+    }
+
+    let version = packages.iter().copied().collect::<Vec<_>>().join(",");
+
+    let mut modules_vec: Vec<Module> = packages.iter().map(|pkg| parse_evr(pkg)).collect();
+
     modules_vec.sort_unstable();
+    modules_vec.dedup();
```
Rolv-Apneseth: Isn't this already being built from a set?

src/packagesystem.rs
```diff
@@ -37,113 +37,83 @@ impl PartialOrd for Module {
...
     paths: impl IntoIterator<Item = T>,
 ) -> Result<ContentMetadata>
 where
     T: AsRef<Path>,
 {
-    let mut c = ostreeutil::rpm_cmd(sysroot_path)?;
-    c.args(["-q", "--queryformat", "%{nevra},%{buildtime} ", "-f"]);
-    for arg in paths {
-        c.arg(arg.as_ref());
+    let query_files_script = "usr/lib/bootupd/packagesystem/query-file-owner";
```
Rolv-Apneseth: This should probably be a constant at the top of the file so it's more discoverable

.cci.jenkinsfile
```diff
@@ -17,7 +17,7 @@ parallel build: {
       }
       stage("Unit tests") {
         shwrap("""
-          cargo test --features rpm
+          cargo test
```
Rolv-Apneseth, line 20: No need for the double `cargo test` now

Cargo.toml
```diff
@@ -64,4 +63,4 @@ tag-message = "bootupd {{version}}"
 
 [features]
 default = []
-rpm = ["rpm-version"]
+rpm = []
```
Rolv-Apneseth, line 66: I think we can just remove the feature

Makefile
Rolv-Apneseth: You may have forgotten to adjust this to `${DESTDIR}${PREFIX}`

README-devel.md
```diff
@@ -21,13 +21,31 @@ There's a reference [Dockerfile](Dockerfile) that builds on [CentOS Stream bootc
...
+
+### Modifying query-file-owner
+
+This script provides package ownership detection for `packagesystem.rs`, the script is in `packagesystem/query-file-owner`,
+in the file filesystem the path is `usr/lib/bootupd/packagesystem/query-file-owner` in the sysroot.
+
+Now you need to modify this file to work with your package manager, ex. `deb/pacman/apk/rpm` (RPM is in there as an example)
+
+You can test it like this:
+`query-file-owner <file releted to package> <another file related to package>`
```
Rolv-Apneseth:
```suggestion
`query-file-owner FILE [FILE...]`
```

README-devel.md
```diff
@@ -21,13 +21,31 @@ There's a reference [Dockerfile](Dockerfile) that builds on [CentOS Stream bootc
...
+
+- Make sure it returns a list of packages info and it should look like this:
+  `grub2-efi-x64-1:2.06-95.fc38.x86_64
```
Rolv-Apneseth: Space-separated now, right? Though it may be worth discussing that version strings will be different now. I think we're gonna be moving away from using versions for comparisons anyway so I think it should be fine.

packagesystem/query-file-owner
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/bash
+set -euo pipefail
+
+# Script to detect package ownership of a file in a Linux system.
+
+# DONT TOUCH THE FOLLOWING LINES UNLESS YOU KNOW WHAT YOU ARE DOING
```
Rolv-Apneseth:
```suggestion
# Required environment setup for consistent output
```

packagesystem/query-file-owner
```diff
@@ -0,0 +1,39 @@
...
+# THERE ARE SOME ENVIRONMENT VARIABLES THAT ARE REQUIRED FOR THE SCRIPT TO WORK PROPERLY
+export LC_ALL=C
+export LANG=C
+
+# Usage: ./query-file-owner <file_path> <file_path> ... <file_path>
```
Rolv-Apneseth:
```suggestion
# Usage: ./query-file-owner FILE [FILE...]
```

packagesystem/query-file-owner
krolmiki2011: Yea, i think seperate is better

src/packagesystem.rs
krolmiki2011: We cannot rely on `BTreeSet`; we need to sort the list slightly.

src/packagesystem.rs
krolmiki2011: Done it in next Commit

.cci.jenkinsfile
krolmiki2011: Done it in next Commit

Cargo.toml
krolmiki2011: Done it in next Commit

Makefile
krolmiki2011: Done it in next Commit

README-devel.md
krolmiki2011: Done it in next Commit

README-devel.md
krolmiki2011: Done it in next Commit

packagesystem/query-file-owner
krolmiki2011: Done it in next Commit

packagesystem/query-file-owner
krolmiki2011: Done it in next Commit

src/packagesystem.rs
Rolv-Apneseth: Sure, we sort the list, but why `dedup()`?

packagesystem/query-file-owner
krolmiki2011: Done it in next Commit

src/packagesystem.rs
krolmiki2011: So i will leave what is it

Rolv-Apneseth: I mean, correct me if I'm wrong, but `dedup()` here should never have anything to do since the set already de-duplicates. May as well drop that line.

krolmiki2011: I guess you right

packagesystem/query-file-owner-apk
```diff
@@ -0,0 +1,29 @@
...
+for FILE in "$@"; do
+    if [ ! -e "$FILE" ]; then
+        echo "Error: File '$FILE' does not exist."
+        exit 1
+    fi
+
+    # Use your package manager to find the package that owns the file
+    # Package Manager should return two space-separated values: NAME and VERSION
+    apk info -W "$FILE" | sed 's/.*is owned by //' | \
+        sed -E 's/-r[0-9]+$//' | sed -E 's/^(.+)-([^-]+)$/\1 \2/'
```
Rolv-Apneseth: Tested this on Alpine and I think it's stripping the package release version e.g. `-r0`:

```sh
$ apk info -W "$FILE" | sed 's/.*is owned by //' | \
     sed -E 's/-r[0-9]+$//' | sed -E 's/^(.+)-([^-]+)$/\1 \2/'
busybox 1.37.0
```

I'd suggest this instead:

```suggestion
    apk info -W "$FILE" | sed 's/.*is owned by //' | \
        sed -E 's/-([0-9])/\ \1/'
```

Which keeps the release version:

```sh
$ apk info -W "$FILE" | sed 's/.*is owned by //' | \
     sed -E 's/-([0-9])/\ \1/'
busybox 1.37.0-r31
```

Let me know if you spot any issues with that though.

src/packagesystem.rs
```diff
@@ -37,113 +39,79 @@ impl PartialOrd for Module {
...
+    if !output.status.success() {
+        std::io::stderr().write_all(&output.stderr)?;
+        bail!("query-file-owner failed");
+    }
 
-    Some((name.to_string(), format!("{version}-{release}")))
+    parse_package_metadata(&output.stdout)
 }
 
 fn parse_evr(pkg: &str) -> Module {
```
Rolv-Apneseth: Might be worth renaming this function. Something like `parse_module`, or even `Module::parse`?

src/util.rs
```diff
@@ -21,6 +21,28 @@ pub(crate) fn getenv_utf8(n: &str) -> Result<Option<String>> {
     }
 }
 
+#[context("Getting timestamp for metadata")]
```
Rolv-Apneseth: Can we keep the function in the same place so it doesn't show up in the diff?

packagesystem/query-file-owner-apk
krolmiki2011: No its fine, actually this is event better!

src/packagesystem.rs
krolmiki2011: @Rolv-Apneseth `parse_package_metadata()` or `parse_evr()`

Rolv-Apneseth: Sorry, `parse_evr`. It's annoying I click on one line to highlight and the view does this.

krolmiki2011: I renamed it in the last commit

Rolv-Apneseth:
CI failure is just cause the jenkins job is run from the jenkinsfile in `main`, not this branch.

```
08:08:21  + cargo test --features rpm
08:08:21  error: the package 'bootupd' does not contain this feature: rpm
```

And before that, the failures in that job are because there's a couple places in `tests/e2e-update/e2e-update-in-vm.sh` we need to change to expect space-separated output.

---

The `bootc-e2e` CI failed with:

```
error: generating metadata failed: query-file-owner failed
```

1. I think we should include the stderr from the script in this error message
2. The issue (I believe) is with checking the existence of files (`[ ! -e "$FILE" ]`). Leave that check out since (at least) RPM doesn't require a file to actually exist and just checks its database to see who has a claim over that path. This is used in the legacy EFI handling, since files are moved from their original paths ( `transfer_ostree_boot_to_bootupd_updates`).

src/packagesystem.rs
```diff
@@ -118,7 +118,7 @@ fn parse_evr(pkg: &str) -> Module {
 fn parse_evr_vec(input: &str) -> Vec<Module> {
```
Rolv-Apneseth: Should rename this too, we're not really parsing any evrs

krolmiki2011: I will, just a moment

krolmiki2011: Done

Rolv-Apneseth:
Maybe just add the empty `rpm` feature back so we can see where we stand with CI @krolmiki2011

krolmiki2011:
> Maybe just add the empty `rpm` feature back so we can see where we stand with CI @krolmiki2011 

Thats why i added earlier

Rolv-Apneseth:
> Thats why i added earlier

Fair enough, didn't realise that was why

krolmiki2011:
Now lets see is this gonna work

krolmiki2011:
Query File Script works in test

krolmiki2011:
@Rolv-Apneseth i think i mess up in e2e test scripts

krolmiki2011:
and i mess up

krolmiki2011:
At lleast bootc worked

Rolv-Apneseth:
Not sure it's your fault, though I'm not sure why it doesn't already happen on main, but it seems like some kind of race condition:

```
[2026-08-26T12:23:12.251Z] # Starting /run/bootupd-source/tests/e2e-update//e2e-update-in-vm.sh
[2026-08-26T12:23:12.251Z] # Skipped: BIOS
[2026-08-26T12:23:12.251Z] # Validated: EFI
[2026-08-26T12:23:12.251Z] # ok validate
[2026-08-26T12:23:12.251Z] # Component BIOS
[2026-08-26T12:23:12.251Z] #   Installed: grub2-tools 1:2.12-56.fc44
[2026-08-26T12:23:12.251Z] #   Update: Available: grub2-tools 1:2.12-58.fc44
[2026-08-26T12:23:12.251Z] # Component EFI
[2026-08-26T12:23:12.251Z] #   Installed: grub2 1:2.12-56.fc44,shim 16.1-5
[2026-08-26T12:23:12.251Z] #   Update: Available: grub2 1:2.12-58.fc44,shim 16.1-5,test_bootupd_payload 1.0
[2026-08-26T12:23:12.251Z] # No components are adoptable.
[2026-08-26T12:23:12.251Z] # Aleph version: 44.20260826.20.dev1
[2026-08-26T12:23:12.251Z] # Boot method: EFI
[2026-08-26T12:23:12.251Z] # ok update avail
[2026-08-26T12:23:12.251Z] # Previous BIOS: grub2-tools 1:2.12-56.fc44
[2026-08-26T12:23:12.251Z] # Updated BIOS: grub2-tools 1:2.12-58.fc44
[2026-08-26T12:23:12.251Z] # [TRACE bootupd] executing cli
[2026-08-26T12:23:12.251Z] # [DEBUG bootupd::aleph] Found aleph version in sysroot/.coreos-aleph-version.json
[2026-08-26T12:23:12.251Z] # [TRACE bootupd::bootupd] Gathering status for installed component: BIOS
[2026-08-26T12:23:12.251Z] # [TRACE bootupd::bootupd] Gathering status for installed component: EFI
[2026-08-26T12:23:12.251Z] # [DEBUG bootupd::efi] Unmounting RefCell { value: None }
[2026-08-26T12:23:12.251Z] # [TRACE bootupd::bootupd] Remaining known components: 0
[2026-08-26T12:23:12.251Z] # No update available for any component.
[2026-08-26T12:23:12.251Z] # -rw-r--r--. 1 root root 39 Aug 26 12:23 out.txt
[2026-08-26T12:23:12.251Z] # # No update available for any component.
[2026-08-26T12:23:12.251Z] # error: File 'out.txt' doesn't match regexp 'Previous EFI: .*'
[2026-08-26T12:23:12.251Z] + fatal 'test failed'
[2026-08-26T12:23:12.251Z] + echo error: test failed
```

Could you try apply this patch and we'll see if that prevents it:

```diff
diff --git a/tests/e2e-update/e2e-update.sh b/tests/e2e-update/e2e-update.sh
index 07c6da00..9efa583c 100755
--- a/tests/e2e-update/e2e-update.sh
+++ b/tests/e2e-update/e2e-update.sh
@@ -93,6 +93,7 @@ systemd:
       enabled: true
       contents: |
         [Unit]
+        Before=bootloader-update.service
         RequiresMountsFor=/run/testtmp
         [Service]
         Type=oneshot
```

krolmiki2011:
Okay i applied

krolmiki2011:
@Rolv-Apneseth it failed again

krolmiki2011:
I find the issue

krolmiki2011:
E2E PASSED

krolmiki2011:
Kola left

krolmiki2011:
@Rolv-Apneseth tests works

packagesystem/query-file-owner-apk
```diff
@@ -0,0 +1,25 @@
...
+export LANG=C
+
+# Usage: ./query-file-owner [FILE] [FILE...]
+
+if [ "$#" -eq 0 ]; then
+    echo "Usage: $0 [FILE] [FILE...]"
+    exit 1
+fi
+
+# Check if the files exists
```
Rolv-Apneseth: Comment not needed anymore. And may as well clean up the extra whitespace in the loop.

Cargo.toml
```diff
@@ -37,7 +37,7 @@ log = "^0.4"
 cap-std = "4.0.2"
 openssl = "^0.10"
 os-release = "0.1.0"
-rpm-version = { version = "0.5.0", default-features = false, optional = true }
+regex = "1.12.3"
```
Rolv-Apneseth: This shouldn't be here, I assume something went wrong with the rebase. See https://github.com/coreos/bootupd/pull/1142

README-devel.md
```diff
@@ -21,13 +21,33 @@ There's a reference [Dockerfile](Dockerfile) that builds on [CentOS Stream bootc
...
 
 Many bootupd developers (and current CI flows) target Fedora CoreOS
 and derivatives, so it can be used as a "reference" for integration.
 
-There's two parts to integration:
+There's three parts to integration:
+
+### Modifying query-file-owner
+
+This script provides package ownership detection for `packagesystem.rs`, the script is in `packagesystem/query-file-owner`,
```
Rolv-Apneseth: Might be worth mentioning `query-file-owner-${package_system}` here and below. You don't necessarily modify the file but pick the correct one or adjust if your package manager isn't one of our reference implementations.

README-devel.md
```diff
@@ -21,13 +21,33 @@ There's a reference [Dockerfile](Dockerfile) that builds on [CentOS Stream bootc
...
+`query-file-owner [FILE] [FILE...]`
+
+- Make sure it returns a list of packages info and it should look like this:
+  `grub2-efi-x64 1:2.06-95.fc38
+  shim-x64 15.6-2` (one package per line)
+
+  So, the output of package looks like this:
+  `NAME VERSION`
+
+- Don't mess to much in the script, because you can break it, and output will be incomprehensible for `packagesystem`
```
Rolv-Apneseth: I don't think this warning is necessary

Rolv-Apneseth:
> @Rolv-Apneseth tests works

Fantastic. From my point of view the only (major) thing left would be cleaning up all commits into 1 or maybe 2 commits since we don't use squash commits (don't have the context on why).

I would really like at least 1 other review on here though (@Johan-Liebert1 maybe) since this is a big PR. There may be some concerns about making the name+version space-separated, but I think it's the way to go as otherwise we can't (AFAICT) reliably determine where package name ends and version starts when parsing something like `grub2-efi-x64-1:2.06-95.fc38.x86_64`. Other than that, we agreed a script is the way to go, and with https://github.com/coreos/bootupd/issues/1137 we'll be moving away from using versions for comparisons anyway.

Worth noting that this also addresses one of the fixes mentioned in https://github.com/coreos/bootupd/pull/1073

> truncation of hyphenated RPM names (e.g. bcm2711-firmware becomes bcm2711)

Cargo.toml
```diff
@@ -37,7 +37,6 @@ log = "^0.4"
 cap-std = "4.0.2"
 openssl = "^0.10"
 os-release = "0.1.0"
-regex = "1.12.3"
```
Rolv-Apneseth: Lockfile needs to be bumped too

krolmiki2011: I removed from the .lock

packagesystem/query-file-owner-apk
Johan-Liebert1, line 1: These scripts perform 80% of the same operations. It's probably cleaner to have one file with some switch cases. Can be a followup

src/packagesystem.rs
```diff
@@ -1,19 +1,20 @@
...
-use crate::ostreeutil;
+use crate::util::get_metadata_timestamp;
+
+pub(crate) const QUERY_FILE_OWNER_SCRIPT: &str = "usr/lib/bootupd/packagesystem/query-file-owner";
 
 #[derive(Serialize, Deserialize, Clone, Debug, Eq, Hash, PartialEq)]
 pub(crate) struct Module {
     pub(crate) name: String,
+
     pub(crate) rpm_evr: String,
```
Johan-Liebert1: Not "rpm" evr anymore with this change. Should probably just change this to "evr" and add serde aliases

src/packagesystem.rs
```diff
@@ -37,120 +38,86 @@ impl PartialOrd for Module {
...
-    let largest_timestamp = timestamps.iter().last().unwrap();
-    let version = pkgs.keys().fold("".to_string(), |mut s, n| {
-        if !s.is_empty() {
-            s.push(',');
+/// Parse the output of the `query-file-owner` script.
+///
+/// Each line contains one package with two space-separated values: NAME and VERSION.
+/// The format depends on the package manager (rpm, dpkg, pacman, etc.):
+fn parse_package_metadata(stdout: &[u8]) -> Result<ContentMetadata> {
+    let output = std::str::from_utf8(stdout)?;
```
Johan-Liebert1: Context on error here would be nice. Also debug log on what we got as the output

src/efi.rs
```diff
@@ -902,7 +903,7 @@ fn generate_meta_from_usr_efi(sysroot_path: &Utf8Path) -> Result<ContentMetadata
     let mut packages = Vec::new();
     let mut modules_vec: Vec<Module> = vec![];
     for efi in efi_components {
-        packages.push(format!("{}-{}", efi.name, efi.version));
+        packages.push(format!("{} {}", efi.name, efi.version));
```
Johan-Liebert1, line 906: Why remove the `-`? We could have some obscure test that maybe splits on `-` to test bootupd.

packagesystem/query-file-owner-rpm
```diff
@@ -0,0 +1,20 @@
...
+# Usage: ./query-file-owner [FILE] [FILE...]
+if [ "$#" -eq 0 ]; then
+    echo "Usage: $0 [FILE] [FILE...]"
+    exit 1
+fi
+
+for FILE in "$@"; do
+    # Use your package manager to find the package that owns the file
+    # Package Manager should return two space-separated values: NAME and VERSION
+    rpm -q --qf '%{NAME} %{EVR}\n' -f "$FILE"
```
Johan-Liebert1, line 18: Again let's keep the current semantics of `name-evr`? Is there a reason why you removed the `-`?

src/packagesystem.rs
```diff
@@ -37,120 +38,86 @@ impl PartialOrd for Module {
...
+            query_files_script_path
+        );
     }
 
-    let rpmout = c.output()?;
-    if !rpmout.status.success() {
-        std::io::stderr().write_all(&rpmout.stderr)?;
-        bail!("Failed to invoke rpm -qf");
+    let mut cmd = std::process::Command::new(query_files_script_path);
+    for path in paths {
```
Johan-Liebert1, line 106: Minor stuff, but paths implements iterator we should just be able to do `cmd.args(paths)`

src/packagesystem.rs
```diff
@@ -37,120 +38,86 @@ impl PartialOrd for Module {
...
-            (nevra.name().to_string(), nevra.evr().to_string())
-        }
-    };
+    parse_package_metadata(&output.stdout)
+}
 
-    let (name, _) = name_str.split_once('-').unwrap_or((&name_str, ""));
+fn parse_module(pkg: &str) -> Module {
+    // Format: "NAME VERSION" space-separated (e.g., "grub2 1:2.06-110.el9" from rpm)
+    let (name, evr) = pkg.split_once(' ').unwrap_or((pkg, ""));
```
Johan-Liebert1: this should be a hard error if we don't find the version

Johan-Liebert1: I guess the earlier version did the same thing which is incorrect imo

packagesystem/query-file-owner-apk
Johan-Liebert1: On second thoughts, can we have this in bootupd itself? We can always check `os-release` and switch commands. I don't have a strong opinion on this, @Rolv-Apneseth do you?

Johan-Liebert1 (changes requested):
Overall logic looks okay. Could really help with tests for other distros as well

packagesystem/query-file-owner-apk
Rolv-Apneseth: No not really. As I mentioned in https://github.com/coreos/bootupd/pull/1130#issuecomment-5326735450, that would have been my first approach. @cgwalters suggested a script in the discussion over in https://github.com/coreos/bootupd/issues/468, and confirmed it in https://github.com/coreos/bootupd/issues/468#issuecomment-5345928799, so that's what was done here.

The multiple scripts vs 1 was to avoid `command -v` on every command, where presumably the packagers know exactly what packaging system they're prepping `bootupd` for so they can just pick the correct one. Don't have a strong opinion on that either though.

src/packagesystem.rs
Rolv-Apneseth: Good catch

packagesystem/query-file-owner-rpm
Rolv-Apneseth: This was discussed above - space-separated so we can actually parse out the package name vs version in something like `grub2-efi-x64-1:2.06-95.fc38.x86_64`. I don't know if it's safe to just split on `-` and split on `:` to find the first component which has only numbers and `.`, but that seems fragile. Also, previous `-` splitting logic was truncating package names. https://github.com/coreos/bootupd/pull/1130#issuecomment-5427532820

src/packagesystem.rs
Rolv-Apneseth: Kinda tricky cause it's `paths: impl IntoIterator<Item = T> where T: AsRef<Path>`

```
the trait bound `T: std::convert::AsRef<std::ffi::OsStr>` is not satisfied
the trait `std::convert::AsRef<std::ffi::OsStr>` is not implemented for `T` [E0277]
process.rs:777:12: required by a bound in `std::process::Command::args`
```

src/packagesystem.rs
Rolv-Apneseth: I tend to agree, but maybe as we move away from versions for comparisons it doesn't matter as much?

Rolv-Apneseth:
> Could really help with tests for other distros as well

What kind of tests did you have in mind? Containers run in CI that check script output?

Johan-Liebert1:
> What kind of tests did you have in mind? Containers run in CI that check script output?

pretty much

packagesystem/query-file-owner-rpm
Johan-Liebert1: Okay, this makes sense. I'm not a 100% sure about this. If we're changing this anyway, maybe we could take a more structured approach where the script returns some sort of structured data, maybe json. Don't think this is a huge deal, but it feels like we're splitting the work 50-50 between the scripts and bootupd

src/packagesystem.rs
Johan-Liebert1: Ah, the joys of rust

Rolv-Apneseth:
> > What kind of tests did you have in mind? Containers run in CI that check script output?
> 
> pretty much

Ok. Yeah that's a good idea. Let's decide if we're keeping the script(s) first I guess.

Also @krolmiki2011  I realised that Alpine probably doesn't have `bash` by default so that script at least needs adjusting.

packagesystem/query-file-owner-apk
Rolv-Apneseth: In favour of the scripts, it's a lot easier for someone to adjust the script for a package system we don't have covered here than it would be if it was in Rust code. One line change if modifying one of these reference scripts.

packagesystem/query-file-owner-rpm
Rolv-Apneseth: I don't know how complex it would be to get JSON output from all of these package managers, but that doesn't sound easy. Splitting on whitespace just seemed like the way to go so it's easy to implement getting the expected output while keeping it simple for us to parse out name+version.

packagesystem/query-file-owner-apk
Rolv-Apneseth: Another point in favour of _separate_ scripts is the Alpine comment I made below - there may be subtle adjustments needed for each distro.

Johan-Liebert1 (changes requested):
The commit messages could really use some context. Maybe some context on why some changes were made so that others reading the commits can also chime in.

Also, probably worth splitting into multiple prep commits as currently each commit covers a lot of ground and not focused on one particular change

krolmiki2011:
Okay why its still processing updates?

Rolv-Apneseth:
Seems like misbehaving CI, don't worry about it. We need @Johan-Liebert1 to agree to the space-separated approach (or proceed with the structured output discussion above) before proceeding with this PR. Once we have that, we probably want, as previously mentioned, some CI to actually validate the scripts are returning output formatted like we're expecting.

Finally, we'll need to clean up the commits here. I'm thinking probably 1 commit, since I'm not sure we can break this up while still having something we can run. A suggestion for the message:

> *: drop RPM dependency and move system packaging concerns to external scripts
> 
> Replace the hard-coded RPM queries with a call to an external script at a known path (/usr/lib/bootupd/packagesystem/query-file-owner). This allows distros to provide their own implementation for whichever package manager they use
> 
> Reference implementations are included for rpm, dpkg, pacman and apk. The output format is space-separated 'NAME VERSION' with one package per line.
> 
> As part of this change, switch the version string format from NEVRA (e.g. grub2-efi-x64-1:2.06-95.fc44.x86_64) to 'NAME VERSION' (e.g. grub2-efi-x64 1:2.06-95.fc44). This avoids ambiguous name/version splitting which was previously truncating package names (e.g. bcm2711-firmware -> bcm2711).

krolmiki2011:
Yea, im gonna wait for yours decision

Johan-Liebert1:
> We need @Johan-Liebert1 to agree to the space-separated approach

I'm okay with this approach, just wanted to know the reasoning. I should've been clearer when resolving comments

The main issue remaining in this PR, imo, is/are the commits. We need them to be more descriptive on why certain design choices were made and if there are any implications wrt to said choices.

Ex https://github.com/coreos/bootupd/pull/1130/changes/56b4ec961323c95f35ed1f35c4e8eb0726374a38 commit message is extremely vauge and probably could be separated out into two - one for the scripts and other for the rust side. If they're interlinked (which I think they are), then the message lacks that description

Rolv-Apneseth:
I would do everything here in just 1 commit personally. Nothing here would run as it's own commit so I don't really think it makes sense to split them.

Implication for the space-separated choice (for the current version comparisons we do) is that any old metadata would be slightly different and would therefore be marked as needing an update (IIUC). I assume this is fine since we'll presumably be moving away from version comparisons anyway in https://github.com/coreos/bootupd/issues/1137 right?

krolmiki2011:
Yea, so i consolideded all changes into one commit

krolmiki2011:
@Johan-Liebert1 Can you do a review please?

Rolv-Apneseth:
@krolmiki2011, when you're trying to update the branch against `main`, use `rebase` instead of a merge commit. Also, the 1 commit still needs some renaming, with actual details included. Start with something like my suggestion from https://github.com/coreos/bootupd/pull/1130#issuecomment-5492033904, but @Johan-Liebert1 may want more details in the description.

krolmiki2011:
> @krolmiki2011, when you're trying to update the branch against `main`, use `rebase` instead of a merge commit. Also, the 1 commit still needs some renaming, with actual details included. Start with something like my suggestion from https://github.com/coreos/bootupd/pull/1130#issuecomment-5492033904, but @Johan-Liebert1 may want more details in the description.

What about now?

src/packagesystem.rs
```diff
@@ -1,33 +1,36 @@
...
+use log::debug;
+
+pub(crate) const QUERY_FILE_OWNER_SCRIPT: &str = "usr/lib/bootupd/packagesystem/query-file-owner";
 
 #[derive(Serialize, Deserialize, Clone, Debug, Eq, Hash, PartialEq)]
 pub(crate) struct Module {
     pub(crate) name: String,
-    pub(crate) rpm_evr: String,
+
+    #[serde(rename = "rpm_evr", alias = "evr")]
```
Johan-Liebert1: Why rename though? Wouldn't we want to go with `evr`?


```suggestion
    #[serde(alias = "rpm_evr")]
```

Cargo.toml
Johan-Liebert1, line 66: can just remove this now, no?

Johan-Liebert1 (commented):
I think this looks good now. This is a big change so should we make this incompatible with older versions by bumping the major version and releasing `v1.0.0` wdyt @Rolv-Apneseth @cgwalters ? 

I'm not entirely sure if this will work with metadata generated using older versions of bootupd. Probably worth adding backwards compatibility tests.

The code itself on the surface lgtm

Cargo.toml
Rolv-Apneseth: We can but CI fails, seems like the jenkinsfile from main is used instead of the updated one here

src/packagesystem.rs
krolmiki2011: Coderabbit told to change

Johan-Liebert1: The current serde alias will rename `evr` to `rpm_ever` in the final JSON which is not what we want right. We want the final JSON to have `evr` but `serde` also be able to parse `rpm_evr` for backwards compatibility

Rolv-Apneseth:
> I think this looks good now. This is a big change so should we make this incompatible with older versions by bumping the major version and releasing `v1.0.0` wdyt @Rolv-Apneseth @cgwalters ?

This should be released with a fix for https://github.com/coreos/bootupd/issues/1137, if we're going ahead with not comparing versions anymore. Not sure it's worth 1.0 though, but I don't have strong feelings on the versioning.

> I'm not entirely sure if this will work with metadata generated using older versions of bootupd. Probably worth adding backwards compatibility tests.

Probably worth it yes, but we can do that as a follow up (same with the CI for checking script output). I believe from when I was looking at it before that it'll be the name that's different if it's no longer getting truncated ("grub2" -> "grub2-efi-x64"), so `bootupd` will report there's an update even if nothing's actually changed.

src/packagesystem.rs
krolmiki2011: Okay, i will fix that

krolmiki2011: Now It should be fine

krolmiki2011:
That should be it right?, @Johan-Liebert1.

Johan-Liebert1 (approved):
This looks okay now from my side

Rolv-Apneseth:
Mine too. Should we get a release out before merging this though @Johan-Liebert, since we probably need https://github.com/coreos/bootupd/pull/1151 soon as possible

Johan-Liebert1:
Yeah, #1151 is a bit more pressing imo. I still need to run the fedora QA tests on that build. I'll report over there

Rolv-Apneseth:
Rest assured we will merge this PR soon as we can @krolmiki2011, thanks for all your work

krolmiki2011:
@coderabbitai review

Rolv-Apneseth:
I think we're ready to merge here. Could you try rebase to see if we can resolve that prow CI @krolmiki2011?

krolmiki2011:
> I think we're ready to merge here. Could you try rebase to see if we can resolve that prow CI @krolmiki2011?
> 

I did it @Rolv-Apneseth

Rolv-Apneseth:
Thank you @krolmiki2011!

PR #1131, opened 2026-07-27
status: align JSON and default output

jbtrystram:
Add `aleph-version` and `boot-method` to the `Status` struct so both `--json` and regular output are showing consistent content.

PR #1133, opened 2026-07-27
ci: fix cross job failures on s390x and ppc64le

Rolv-Apneseth:
Install build-essential in this job to ensure we get libc6-dev, which is what's needed to avoid the linking errors in the most recent CI runs.

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -35,7 +35,7 @@ jobs:
           run: |
             set -xeu
             apt update -y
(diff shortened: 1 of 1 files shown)


Rolv-Apneseth:
Fix seems to have worked, ready to merge once the remaining CI passes

cgwalters (approved):

Rolv-Apneseth:
Seems the failure for `continuous-integration/jenkins/pr-merge` is related to https://github.com/coreos/bootupd/pull/1131

PR #1134, opened 2026-07-30
Release v0.2.36

Johan-Liebert1:


Johan-Liebert1:
This has some pretty big changes, especially https://github.com/coreos/bootupd/pull/1113. Should we update the minor version to 3 instead? `0.3.0`?

cgwalters (approved):

cgwalters:
> This has some pretty big changes, especially https://github.com/coreos/bootupd/pull/1113. Should we update the minor version to 3 instead? 0.3.0?

At some point probably yeah...

PR #1138, opened 2026-08-21
varlink: add varlink interface for syncing fwupd capsule updates across ESPs

Rolv-Apneseth:
This is related to https://github.com/coreos/fedora-coreos-tracker/issues/1623, and adds a [varlink](https://varlink.org/) interface that `fwupd` can use to tell `bootupd` to sync firmware capsule updates to other co-located ESPs for our RAID setups (redundant ESPs).

There is already a [related `fwupd` PR](https://github.com/fwupd/fwupd/pull/10851), which @hughsie built against [a copr project I created](https://copr.fedorainfracloud.org/coprs/rapneset/bootupd-varlink/). The tests included here should confirm we're working as intended though.

The way it works is it accepts a `partuuid` of the device the update was written to, and a `capsule_dir` with the relative path to the update dir on the device, and syncs those updates across co-located ESPs by mounting each in turn. On my system, I can interact with the interface like this:

```sh
sudo dnf copr enable rapneset/bootupd-varlink
sudo dnf install bootupd-0.3d32bdb --repo 'copr:copr.fedorainfracloud.org:rapneset:bootupd-varlink'
sudo systemctl start bootupd-varlink.socket
sudo varlinkctl call /run/bootupd/org.coreos.bootupd1 \
  org.coreos.bootupd1.SyncFwupdUpdates \
  '{"partuuid": "2e126947-2730-49df-af3f-012de73bccfd", "capsule_dir": "EFI/fedora/fw"}'
```

Couple notes:

- Put the `varlink` subcommand under `bootupd` since it shouldn't be user-facing
- Implemented a simple service first, but now I'm leaning mostly towards socket-activated (both currently implemented)
- Socket will need to be enabled in `fedora-coreos-config`. We'll also need to have `udisks2` for `fwupd` to work as intended.
- I haven't made any updates to the readme just yet - waiting to see if we want to make any changes to the approach here
- RAID tests are divided by architecture just like the [raid1-boot](https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola/boot/raid1-boot) tests
- Since `zlink` required an async runtime, I chose `smol` as a lightweight option


Rolv-Apneseth:
CI failures related to https://github.com/coreos/fedora-coreos-config/pull/4261

Rolv-Apneseth:
I'll include a quick note on why we even need this:

Typical RAID setups only have a single ESP, but CoreOS duplicates the ESP as plain vfat partitions across all disks, which are all updated by `bootupd` (and hence kept in "sync"). If fwupd places capsule updates on one ESP, the firmware will only process them from the ESP on the boot disk. By syncing capsules to all co-located ESPs, we ensure the firmware update applies regardless of which disk the system boots from.

contrib/packaging/bootupd.spec
```diff
@@ -55,6 +55,8 @@ Conflicts:      bootc < 1.14.1
 %{_libexecdir}/bootupd
 %{_prefix}/lib/bootupd/grub2-static/
 %{_unitdir}/bootloader-update.service
+%{_unitdir}/bootupd-varlink.socket
+%{_unitdir}/bootupd-varlink.service
```
cgwalters, line 59: I think it'd be nice to have systemd units enforce locking.

So we have a single `bootupd.service` and change bootloader-update.service to actually call into the varlink API too, which would activate that service in the same way.

Alternatively, do we actually need a `.socket` unit? What we did in e.g. varlink for https://github.com/bootc-dev/bcvk/ is that it remains a CLI that has an interface one forks that enables varlink.

src/varlink.rs
```diff
@@ -0,0 +1,216 @@
...
+        unix::io::{FromRawFd, OwnedFd},
+    },
+    path::{Path, PathBuf},
+};
+
+use crate::{
+    bootupd::list_dev_current_root, efi::Efi, freezethaw::fsfreeze_thaw_cycle, model::SavedState,
+};
+
+const SOCKET_PATH: &str = "/run/bootupd/org.coreos.bootupd1";
```
cgwalters, line 18: See above re the socket

src/varlink.rs
```diff
@@ -0,0 +1,216 @@
...
+            message: format!("{err}"),
+        }
+    }
+}
+
+struct BootupdVarlinkService;
+
+#[zlink::service(interface = "org.coreos.bootupd1")]
+impl BootupdVarlinkService {
+    /// Sync capsule update files from a "primary" ESP to all colocated ESPs.
```
cgwalters, line 57: Should this really be specific to capsules? Isn't the general use case here "I am some software that touched one of the ESPs, please sync the others"?

systemd/bootupd-varlink.service
```diff
@@ -0,0 +1,17 @@
...
+Description=Bootupd Varlink Service
+Documentation=https://github.com/coreos/bootupd
+# Interface only implemented on EFI systems
+ConditionPathExists=/sys/firmware/efi
+
+[Service]
+Type=simple
+# It doesn't make sense to sync ESP updates in "Live" environments.
+# https://github.com/coreos/fedora-coreos-tracker/issues/2136
+ExecCondition=/bin/bash -c '[[ ! $(findmnt -n -o FSTYPE /sysroot) =~ ^(erofs|squashfs)$ ]]'
```
cgwalters, line 11: See above, if we changed bootloader-update.service then we could avoid copy-pasta this

src/varlink.rs
Rolv-Apneseth: That's what I assumed at first, but @travier wanted just the "fwupd folder" in https://github.com/coreos/fedora-coreos-tracker/issues/1623#issuecomment-5115350336

contrib/packaging/bootupd.spec
Rolv-Apneseth: Current state should [work without a socket unit](https://github.com/coreos/bootupd/pull/1138/changes#diff-0afb43571bce9e99ab69fe28c0f1b013c498d3c18eebc26b29951fccbab1e753R195) actually, so just `bootupd varlink` will run it's own daemon. I don't quite understand the suggestion for a single `bootupd.service` though, would you mind clarifying? If we're gonna have a service anyway why not let `systemd` handle the socket? I don't have a strong opinion on this and initially leaned towards a simple service at first but the PR over on `fwupd` assumed it was socket-activated so I shifted that direction.

Rolv-Apneseth: Thanks for taking a look by the way @cgwalters

src/cli/bootupd.rs
```diff
@@ -178,4 +185,14 @@ impl DCommand {
...
     }
+
+    fn run_varlink_service() -> Result<()> {
+        #[cfg(efi_arch)]
+        {
+            varlink::run_varlink_service()
+        }
+
+        #[cfg(not(efi_arch))]
+        Ok(())
```
Johan-Liebert1: I think making this `unreachable`/`anyhow::bail` is better since that could point to a potential issue instead of just silently exiting.

src/varlink.rs
```diff
@@ -0,0 +1,216 @@
...
+            return Err(BootupdVarlinkError::new(
+                "capsule_dir must not be empty".into(),
+            ));
+        }
+
+        let capsule_dir = Path::new(capsule_dir);
+        // Must be relative
+        if capsule_dir.is_absolute() {
+            return Err(BootupdVarlinkError::new(
+                "capsule_dir must be a relative path".into(),
```
Johan-Liebert1: specifying relative to *what* would be a useful error msg. I'm guessing it should be relative to the ESP?

src/varlink.rs
```diff
@@ -0,0 +1,216 @@
...
+            .components()
+            .any(|c| c == std::path::Component::ParentDir)
+        {
+            return Err(BootupdVarlinkError::new(
+                "capsule_dir must not contain '..' components - path traversal not allowed".into(),
+            ));
+        }
+
+        let root_device = list_dev_current_root()?;
+        let esp_devices = root_device.find_colocated_esps()?.unwrap_or_default();
```
Johan-Liebert1: Why isn't this a hard error if we find an empty vector of ESPs?

src/varlink.rs
```diff
@@ -0,0 +1,216 @@
...
+                "Capsule directory not found at: {src_capsule_path:?}"
+            )));
+        }
+
+        let src_dir = Dir::open_ambient_dir(&src_capsule_path, ambient_authority())
+            .context("opening source capsule dir")?;
+
+        // Sync to every other co-located ESP.
+        let mut synced_count = 0;
+        for esp in esp_devices.iter().filter(|dev| {
```
Johan-Liebert1: Nit: This filter could just just be a continue in the loop

src/varlink.rs
```diff
@@ -0,0 +1,216 @@
...
+            dev.partuuid
+                .as_ref()
+                .is_some_and(|u| !u.eq_ignore_ascii_case(partuuid))
+        }) {
+            let secondary_efi = Efi::default();
+            let dest_mount =
+                secondary_efi.ensure_mounted_esp(Path::new("/"), Path::new(&esp.path()))?;
```
Johan-Liebert1: I'm not quite sure this works. Have you tested this? Since in the `Efi` struct we might simply return the "booted from" ESP if we find it mounted at `/boot` or `/boot/efi`

Johan-Liebert1: Maybe to be absolutely safe, we should just use `TempMount`?

src/varlink.rs
```diff
@@ -0,0 +1,216 @@
...
+            create_dir_all(&dest_capsule_path)
+                .with_context(|| format!("creating {dest_capsule_path:?}"))?;
+
+            let dest_dir = Dir::open_ambient_dir(&dest_capsule_path, ambient_authority())
+                .context("opening destination capsule dir")?;
+
+            for entry in src_dir.entries().context("reading source capsule dir")? {
+                let entry = entry.context("reading dir entry")?;
+                let name = entry.file_name();
+                let contents = src_dir
```
Johan-Liebert1, line 160: Could be replaced with `.copy` method on `Dir`. Also, let's use the atomic write functions

src/varlink.rs
```diff
@@ -0,0 +1,216 @@
...
+                    .write(&name, &contents)
+                    .with_context(|| format!("writing {name:?}"))?;
+            }
+
+            fsfreeze_thaw_cycle(
+                dest_dir
+                    .reopen_as_ownedfd()
+                    .context("reopening dest dir as owned fd")?,
+            )?;
+            drop(dest_dir);
```
Johan-Liebert1, line 173: These are already dropped on scope exit. Any reason why we need to be explicit?

src/cli/bootupd.rs
Rolv-Apneseth: Yeah, looking at it again, the verb itself is behind the arch config so may as well put the whole function behind it too and drop that check entirely. Thanks

src/varlink.rs
Rolv-Apneseth: Ah, good point. Yes, relative to the ESP since `fwupd` will have it mounted at a different path

src/varlink.rs
Rolv-Apneseth: I can add an error there, sure. It should already fail right below when it tries to find the primary device, but can't hurt to fail sooner

src/varlink.rs
Rolv-Apneseth: True, I just tend to prefer `filter` in these scenarios, feels more explicit. I'll change it to `continue` to be more consistent with the rest of the code base though

src/varlink.rs
Rolv-Apneseth: The tests did pass, but I guess that's because we don't have any ESP mounted by default. I'll use `TempMount`, thanks

src/varlink.rs
Rolv-Apneseth: No reason in particular, just making it clear we want it unmounted. Would you prefer without the (now 2) `drop`'s?

src/varlink.rs
Rolv-Apneseth: Which would you prefer, as I think they're mutually exclusive (unless I've misunderstood). I've switched to the atomic write for now.

src/varlink.rs
```diff
@@ -0,0 +1,225 @@
...
+        for esp in esp_devices.iter() {
+            if esp
+                .partuuid
+                .as_ref()
+                .is_some_and(|u| u.eq_ignore_ascii_case(partuuid))
+            {
+                continue;
+            }
+
+            let secondary_efi = efi::mount_esp(&esp.path())
```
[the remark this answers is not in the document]
Rolv-Apneseth: Was meant to be included, forgot to append `efi.rs` changes to the commit

tests/kola/varlink/data/libtest_varlink.sh
```diff
@@ -0,0 +1,193 @@
...
+    assert_file_has_content out.txt "partuuid"
+    assert_file_has_content out.txt "capsule_dir"
+    ok "introspection shows interface"
+}
+
+check_service() {
+    unit="bootupd-varlink.socket"
+    if ! systemctl is-enabled "${unit}" 1> /dev/null; then
+        # TODO: remove when enabled by default
+        systemctl start "$unit"
```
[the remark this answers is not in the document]
Rolv-Apneseth: Temporary, we'll need to have this enabled by default in `fedora-coreos-config` before we can modify this test.

tests/kola/varlink/data/libtest_varlink.sh
```diff
@@ -0,0 +1,193 @@
...
+
+    efi_mount=$(mktemp -d)
+
+    # SINGLE FILE
+    mount "${esp1}" "${efi_mount}"
+    mkdir -p "${efi_mount}/${CAPSULE_DIR}"
+    echo "test" > "${efi_mount}/${CAPSULE_DIR}/${CAPSULE_FILE1}"
+    umount "${efi_mount}"
+
+    if ! varlink_sync "${esp1_partuuid}"; then
```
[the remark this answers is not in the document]
Rolv-Apneseth: That doesn't seem practical to try implement here. We'd need to kill the bootupd service mid-write, then restart + retry. The sync is a fast file copy, so the window for interruption is small, and there's no recovery path anyway. Running it again just does an idempotent copy, same as running it the first time.

Rolv-Apneseth:
Interesting concern brought up by the bot: 

> Use a stable inode for the write lock.
> 
> atomic_write_with_perms uses atomic_replace_with to replace run/bootupd-lock before flock opens it. Concurrent callers can therefore lock different inodes, which defeats cross-process exclusion. src/varlink.rs relies on this lock, so bootloader updates and capsule synchronization may overlap.
> 
> Create the lock file without replacing an existing pathname, then acquire flock on that stable file descriptor.

Seems unlikely we'd actually run into this, but I added a commit to address it anyway. Can remove it / open a separate PR if you prefer though @Johan-Liebert1

src/varlink.rs
Johan-Liebert1: Let's go with the atomic write functions. Sorry, should've specified

Johan-Liebert1:
> Seems unlikely we'd actually run into this, but I added a commit to address it anyway. 

that is a really nice catch. I tried it locally and yup, if the inodes are different then flock is kinda useless

> Can remove it / open a separate PR if you prefer though @Johan-Liebert1

A separate PR is better as it's pretty much unrelated to this one

Rolv-Apneseth:
Alright will remove the commit from here, thanks

Johan-Liebert1:
Also, maybe the race conditions we were seeing in the CI might've been because of this?

Rolv-Apneseth:
> Also, maybe the race conditions we were seeing in the CI might've been because of this?

Ah yes, seems likely doesn't it.

I moved that commit over to https://github.com/coreos/bootupd/pull/1138

cgwalters:
> Typical RAID setups only have a single ESP, but CoreOS duplicates the ESP as plain vfat partitions across all disks,

Conceptually I don't think this is CoreOS specific, it's really the only way to do it without having dedicated firmware support.

src/varlink.rs
```diff
@@ -0,0 +1,225 @@
...
+        // Avoid running at the same time as bootloader updates
+        let sysroot = Dir::open_ambient_dir("/", ambient_authority()).context("opening sysroot")?;
+        let _lock = SavedState::acquire_write_lock("/".into(), sysroot)?;
+
+        // Mount primary ESP and find capsule updates dir
+        let primary_efi = Efi::default();
+        let primary_mount =
+            primary_efi.ensure_mounted_esp(Path::new("/"), Path::new(&primary_device.path()))?;
+
+        let src_capsule_path = primary_mount.join(capsule_dir);
```
cgwalters, line 119: Don't we need to honor `partuuid` for finding this?

Rolv-Apneseth:
> Conceptually I don't think this is CoreOS specific, it's really the only way to do it without having dedicated firmware support.

True. My understanding is just that it's not common to do this.

src/varlink.rs
cgwalters: OK, but I want to discuss generalizing because I think the interface and implementation may actually end up *simpler*.

I'm basically proposing that we have a `sync` verb that looks at all ESPs, and if there is newer content/changes on one of them, then we sync it to the others. This should take a file path as an argument, but I feel like in the general case we can cheaply do "detect changes" by looking at modification times or so.

Rolv-Apneseth: I assume he had good reason for wanting to reduce the scope. The main argument I can think of is do we really want to support the use-case of arbitrarily syncing _anything_ across ESPs, when the main goal of this was just to enable reliable firmware updates for our RAID systems? Also, when I tried this when I was doing this as a full sync triggered by a dbus interface I needed more than just checking mtimes IIRC. Not that it was complex, basically just a file storing a timestamp for synced ESPs, but regardless, we'd need additional state tracking.

Any changes to the current interface would also require a change in fwupd to match, since they already merged their side of the changes.

But we can always add that functionality later if we decide we need it? The varlink endpoint could just be updated to internally use a new general syncing mechanism.

src/varlink.rs
Rolv-Apneseth: If you're talking about this using `ensure_mounted_esp` when it should just be a `TempMount` directly like for the secondary device(s), you're right, thanks

Rolv-Apneseth:
CI failure:

```
2026-09-02T10:41:50.961Z] + timeout 5m kola qemuexec --propagate-initramfs-failure --qemu-image /home/jenkins/agent/workspace/bootupd_PR-1138/builds/44.20260901.20.dev1/x86_64/fedora-coreos-44.20260901.20.dev1-qemu.x86_64.qcow2 --qemu-firmware uefi -i test.ign --bind-ro /home/jenkins/agent/workspace/bootupd_PR-1138,/run/cosadir --bind-ro /home/jenkins/agent/workspace/bootupd_PR-1138,/run/bootupd-source --bind-rw /var/tmp/bootupd-e2e.rjvnq95,/run/testtmp --console-to-file /home/jenkins/agent/workspace/bootupd_PR-1138/tmp/console.txt
[2026-09-02T10:46:57.369Z] qemu-system-x86_64: terminating on signal 15 from pid 4675 (timeout)
[2026-09-02T10:46:57.369Z] qemu-system-x86_64: Failed to read msg header. Read -1 instead of 12. Original request 0.
[2026-09-02T10:46:57.369Z] qemu-system-x86_64: Failed to write msg. Wrote -1 instead of 20.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost VQ 0 ring restore failed: -22: Invalid argument (22)
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost VQ 1 ring restore failed: -22: Invalid argument (22)
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost_set_vring_call failed 22
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost_set_vring_call failed 22
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to write msg. Wrote -1 instead of 20.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost VQ 0 ring restore failed: -22: Invalid argument (22)
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost VQ 1 ring restore failed: -22: Invalid argument (22)
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost_set_vring_call failed 22
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost_set_vring_call failed 22
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to read msg header. Read -1 instead of 12. Original request 0.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to write msg. Wrote -1 instead of 20.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost VQ 0 ring restore failed: -22: Invalid argument (22)
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost VQ 1 ring restore failed: -22: Invalid argument (22)
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost_set_vring_call failed 22
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: Failed to set msg fds.
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: vhost_set_vring_call failed 22
[2026-09-02T10:46:57.371Z] qemu-system-x86_64: tpm-emulator: Could not cleanly shutdown the TPM: Invalid argument
```

Re-running

Johan-Liebert1 (approved):
Overall LGTM. @cgwalters maybe the across ESP sync can be a followup to this?

contrib/packaging/bootupd.spec
cgwalters: > I don't quite understand the suggestion for a single bootupd.service though, would you mind clarifying? If we're gonna have a service anyway why not let systemd handle the socket? 

Well if fwupd already merged support for calling into the socket, then clearly we should probably do that.

However! There's new support for "varlink registry" stuff in systemd, see https://github.com/systemd/systemd/pull/40590 among others. We should probably participate in that if we do have a bound socket.

But what I think is actually a good bit more valuable is having *a single systemd .service unit* - if we only have one, then systemd enforces locking for it (while I wouldn't argue to drop our lockfile, having systemd do it is just better).

It means that one can more consistently check logs in one place. It means isolation and configuration only appear in one place, not two.

src/varlink.rs
cgwalters: Sure, I'm fine generalizing later

src/varlink.rs
```diff
@@ -0,0 +1,231 @@
...
+        }
+        // Must not contain path traversal components
+        if capsule_dir
+            .components()
+            .any(|c| c == std::path::Component::ParentDir)
+        {
+            return Err(BootupdVarlinkError::new(
+                "capsule_dir must not contain '..' components - path traversal not allowed".into(),
+            ));
+        }
```
cgwalters: nit stuff like this is a bit better as a little `validate_relative_path(&str) -> &Path` or so

src/varlink.rs
```diff
@@ -0,0 +1,231 @@
...
+        })?;
+        let primary_mount = primary_efi.dir.path().to_path_buf();
+
+        let src_capsule_path = primary_mount.join(capsule_dir);
+        if !src_capsule_path.is_dir() {
+            drop(primary_efi);
+            return Err(BootupdVarlinkError::new(format!(
+                "Capsule directory not found at: {src_capsule_path:?}"
+            )));
+        }
```
cgwalters: Why is this better than just letting the `open_ambient_dir` below throw an error instead?

src/varlink.rs
```diff
@@ -0,0 +1,231 @@
...
+                .with_context(|| format!("creating {dest_capsule_path:?}"))?;
+
+            let dest_dir = Dir::open_ambient_dir(&dest_capsule_path, ambient_authority())
+                .context("opening destination capsule dir")?;
+
+            for entry in src_dir.entries().context("reading source capsule dir")? {
+                let entry = entry.context("reading dir entry")?;
+                let name = entry.file_name();
+                let contents = src_dir
+                    .read(&name)
```
cgwalters: Reading the entire file into memory is unnecessary, https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.atomic_replace_with is the generalized API and we can open the src as a file and use `std::io::copy` between the two.

Hmm that said actually what I would say an important thing to do here is to `fdatasync` each file before renaming

Cargo.toml
```diff
@@ -45,6 +45,8 @@ tempfile = "^3.26"
 uapi-version = "0.4.0"
 walkdir = "2.3.2"
 signal-hook-registry = "1.4.8"
+zlink = { version = "0.7.0", default-features = false, features = ["smol", "service", "tracing"] }
+smol = "2.0.2"
```
cgwalters: I strongly prefer tokio.

cgwalters (changes requested):

src/varlink.rs
Rolv-Apneseth: Sure yeah, makes sense to separate. The list of things to check kept growing as I worked on it and I forgot to extract into a function.

Cargo.toml
Rolv-Apneseth: Fair enough. Any reason? I chose `smol` for the smaller footprint since we are making very limited usage of the async runtime here.

Rolv-Apneseth: Fair enough. Any reason why? I chose `smol` for the smaller footprint since we are making very limited usage of the async runtime here.

src/varlink.rs
Rolv-Apneseth: I suppose it's not much better, I just wanted the custom error message.

src/varlink.rs
Rolv-Apneseth: Thanks I'll look into both points

Cargo.toml
cgwalters: Because mixing async runtimes is hard, and tokio is way more general, and my impressions are that the reasons to use smol have shrunk as tokio has implemented more optimizations for both small and large scale.

cgwalters:
It's *really* important to emphasize that with this we are basically doing a full 180° pivot from https://github.com/coreos/bootupd/pull/663

Cargo.toml
Rolv-Apneseth: I was gonna argue for `smol` since our usage here is so limited, but it looks like `bootc-internal-utils` is already pulling in `tokio` anyway... I'll swap it out.

src/varlink.rs
Rolv-Apneseth: Done now, but I'm not entirely sure it's correct. Mind verifying?

contrib/packaging/bootupd.spec
Rolv-Apneseth: > Well if fwupd already merged support for calling into the socket, then clearly we should probably do that.

It was just documentation, the only assumption in the code is that there's a varlink interface that's callable. I was just explaining why I went that direction.

> However! There's new support for "varlink registry" stuff in systemd, see https://github.com/systemd/systemd/pull/40590 among others. We should probably participate in that if we do have a bound socket.

Sure, don't see why not.

> But what I think is actually a good bit more valuable is having a single systemd .service unit - if we only have one, then systemd enforces locking for it (while I wouldn't argue to drop our lockfile, having systemd do it is just better).
> 
> It means that one can more consistently check logs in one place. It means isolation and configuration only appear in one place, not two.

Just trying to understand, so I have some questions:

We wouldn't be renaming the current unit, right?

Do you imagine it still running the update on boot, and then starting the varlink listener? Would we keep the transient update service for `bootupctl`, or would every command have to go through a `varlink` interface?

Are we fine with requiring people who mask the service currently to manually start `bootupd varlink` when they need the varlink functionality?

Rolv-Apneseth:
> It's _really_ important to emphasize that with this we are basically doing a full 180° pivot from #663

@cgwalters understood. I'll try do the single service, just need to understand what we're looking for a bit better (https://github.com/coreos/bootupd/pull/1138#discussion_r4045877702).

> Preserve the configurable LIBEXECDIR target.

Addressing this unrelated @coderabbitai suggestion in a separate PR: https://github.com/coreos/bootupd/pull/1161

PR #1146, opened 2026-08-27
secureboot: add check to validate signature db certificates before updating

### LINKED ISSUE #1099: Support for the Microsoft Secure Boot certificate transition (integrate sbchooser)
The current certificate created by Microsoft, enrolled in devices and used for signing shim for Secure Boot is expiring at the end of June. Existing devices will keep working past the expiry date but no new signatures will be done by Microsoft past this date.

A new certificate (2023) has been created and new versions of shim are now signed with both keys. As PE signatures can be detached/attached "freely", the latest versions of shim in Fedora/CentOS Stream/RHEL have or will have both signatures soon.

Thus new devices that will only trust the new certificate will also work with the same build of shim.

The issue here is that at one point, we will have to update shim (security issue, new Fedora/CS/RHEL certificate, etc.) and after June 2026, this new build will only be signed with the new key.

Whether or not the trust database can be updated on existing systems depends on the platform or hardware manufacturer. In most cases, fwupd will be responsible of doing the firmware / trust database update.

But before that update happens, bootupd should refuse to update systems to a newer shim signed only with the 2023 Microsoft key if the device does not trust it yet.

A tool has been created to figure that out: sbchooser: https://github.com/rhboot/efivar/pull/294

References:
- https://access.redhat.com/articles/7128933
- https://access.redhat.com/solutions/5436071
- https://blog.einval.com/2026/05/22#secure_boot_ca_rollover
- https://discourse.ubuntu.com/t/microsoft-uefi-ca-rotation-what-it-means-for-ubuntu-users-and-vendors/82652
- https://github.com/microsoft/secureboot_objects/wiki/OEM-Certificate-Key-Rolling

Rolv-Apneseth:
Closes https://github.com/coreos/bootupd/issues/1099

The main concern this resolves is that, with the Microsoft UEFI CA 2011 Certificate having expired, any further updates to the shim will not be dual-signed, and will only be signed by the 2023 key. This means that any system with Secure Boot enabled will be broken if `bootupd` updates the shim and the system does not have the certificate available in their firmware's signature database. So we should make sure `bootupd` is aware of this requirement so we don't break user's Secure Boot systems.

As stated in the issue, we wanted a way to incorporate some of the `sbchooser` logic to achieve this goal. However, I found [this crate](https://crates.io/crates/virtfw-libefi) which suits our use-case perfectly. I discussed briefly with the creator and he agreed it was a good fit.

I've implemented this in such a way that any update will be blocked if Secure Boot is enabled and the 2023 cert is not in the system's signature DB. So updating is allowed as long as any one of the following is true:

1. System is not EFI-booted
2. System does not have Secure Boot enabled
3. System has the 2023 cert in the signature database

Something that might be nice but would add some complexity is only blocking updates if the shim is actually being updated, as that's the only one that matters here. Any thoughts on this? I think it's gonna be a brief enough window where this is actually helpful that keeping it as simple as possible will benefit us more in the long run. Plus Secure Boot systems with outdated certificates should indeed have their firmware updated as soon as possible. I'd be happy to try implement the module comparison logic if we think it's important to only block on shim updates though.

Also for  testing, I have a positive case test, and tested the negative case manually. I can't think of a good way to test the negative case with kola. Maybe including a feature in kola to enable using the db with an older certificate (binary blob?), but I'm not sure that's worth it.

### DIFF
--- a/src/secureboot.rs
+++ b/src/secureboot.rs
@@ -1,0 +1,75 @@
+//! Module with checks related to Secure Boot.
+//!
+//! This is primarily used for checking whether it is safe to update a system which has Secure
+//! Boot enabled. There can be complications if updating to a new version of shim which is only
+//! signed by Microsoft keys which the user's system does not have matching certificates for,
+//! so in that scenario we just block updates until the firmware is updated to match (most likely by
+//! fwupd).
+//!
+//! See https://github.com/coreos/bootupd/issues/1099
+
+use anyhow::{bail, Result};
+use fn_error_context::context;
+use log::info;
+use virtfw_libefi::efivar::{ids, sigdb::EfiSigDB};
+use virtfw_libefi::sb::certs;
+use virtfw_libefi::varstore::sysfs;
+
+/// Check whether Secure Boot is enabled.
+fn is_secureboot_enabled() -> bool {
+    let Some(var) = sysfs::varstore_read(ids::SECURE_BOOT.name, ids::SECURE_BOOT.guid) else {
+        info!("Could not read the Secure Boot EFI variable. Assuming Secure Boot is not enabled.");
+        return false;
+    };
+    // First byte represents a boolean for "is_enabled"
+    match var.data().first() {
+        Some(&b) => b != 0,
+        None => false,
+    }
+}
+
+/// Check whether the firmware's signature database contains the Microsoft UEFI CA 2023 certificate.
+fn db_contains_ms_2023_cert() -> Result<bool> {
+    let Some(var) = sysfs::varstore_read(ids::DB.name, ids::DB.guid) else {
+        // At this stage, we'll have confirmed the user does have Secure Boot enabled.
+        // Best to be safe and abort any updates if we can't even read the EFI variable.
+        bail!("Could not read the signature database variable. Assuming it is not safe to update.");
+    };
+    let Some(sigdb) = EfiSigDB::new_from_bytes(var.data()) else {
+        bail!("Failed to parse Secure Boot signature database (unknown layout). Assuming it is not safe to update.");
+    };
+
+    Ok(sigdb
+        .get_x509_list()
+        .contains(&certs::MICROSOFT_DB_UEFI_2023))
+}
+
+/// Attempt to validate that the system can safely accept an EFI bootloader update (if EFI-booted).
+///
+/// If Secure Boot is enabled, we can't allow updates if the signature database doesn't contain
+/// the Microsoft UEFI CA 2023 certificate.
+#[context("Validating Secure Boot certificate compatibility")]
+pub(crate) fn validate_secureboot_for_update() -> Result<()> {
+    if !crate::efi::is_efi_booted()? {
+        info!("Not EFI-booted, skipping Secure Boot certificate check");
+        return Ok(());
+    }
+
+    if !is_secureboot_enabled() {
+        info!("Secure Boot not enabled, skipping Secure Boot certificate check");
+        return Ok(());
+    };
+
+    match db_contains_ms_2023_cert() {
+        Ok(true) => {
+            info!("Secure Boot DB contains Microsoft UEFI CA 2023 certificate. Safe to update.");
+            Ok(())
+        }
+        Ok(false) => bail!(
+            "Secure Boot is enabled but the Microsoft UEFI CA 2023 certificate was not \
+             found in the firmware's signature database. Updating the shim could render this system \
+             unbootable. Please update your system firmware by using, for example, fwupd."
+        ),
+        Err(e) => Err(e),
+    }
+}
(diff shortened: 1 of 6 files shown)


src/secureboot.rs
```diff
@@ -0,0 +1,75 @@
...
+fn is_secureboot_enabled() -> bool {
+    let Some(var) = sysfs::varstore_read(ids::SECURE_BOOT.name, ids::SECURE_BOOT.guid) else {
+        info!("Could not read the Secure Boot EFI variable. Assuming Secure Boot is not enabled.");
+        return false;
+    };
+    // First bit represents a boolean for "is_enabled"
+    match var.data().first() {
+        Some(&b) => b != 0,
+        None => false,
+    }
```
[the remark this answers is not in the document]
Rolv-Apneseth: Hmm, maybe. I guess if we've confirmed that the system is EFI-booted, this should realistically be readable / have data and there might be something weird happening if it isn't.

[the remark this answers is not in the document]
Rolv-Apneseth: Alright LLM - implemented the errors

Rolv-Apneseth:
Bot suggestion to bail when unable to read Secure Boot variables was incorrect as proven by the CI failure. UEFI can be enabled while the Secure Boot variables don't even exist.

src/secureboot.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    match db_contains_ms_2023_cert() {
+        Ok(true) => {
+            info!("Secure Boot DB contains Microsoft UEFI CA 2023 certificate. Safe to update.");
+            Ok(())
+        }
+        Ok(false) => bail!(
+            "Secure Boot is enabled but the Microsoft UEFI CA 2023 certificate was not \
+             found in the firmware's signature database. Updating the shim could render this system \
+             unbootable. Please update your system firmware by using, for example, fwupd."
```
yasminvalim, line 71: I think the error message is good, but could be improved. Instead of focus on fwupd it could be a little more generic and give the user other options to be more clear, something like:

""Please update your system firmware using your firmware setup menu, fwupd, or similar tools."

WDYT?

yasminvalim:
Thanks for working on this, overall looks pretty good. I just made a suggestion about improving the message error, but is not a blocker. Approved. :)

yasminvalim (approved):

src/secureboot.rs
Rolv-Apneseth: The fwupd maintainer did tell me that 99% of updates are done through some kind of UI not `fwupdmgr`... I was tailoring this more towards FCOS itself but maybe that's reasonable if we want `bootupd` to be distro-agnostic.

src/secureboot.rs
```diff
@@ -0,0 +1,75 @@
...
+        info!("Not EFI-booted, skipping Secure Boot certificate check");
+        return Ok(());
+    }
+
+    if !is_secureboot_enabled() {
+        info!("Secure Boot not enabled, skipping Secure Boot certificate check");
+        return Ok(());
+    };
+
+    match db_contains_ms_2023_cert() {
```
cgwalters, line 63: And we're just assuming that by the time someone updates bootupd with this patch, they will *also* have a need to update the corresponding shim?

That could break some legitimate use cases where someone e.g. pulls a newer bootupd for an unrelated bug fix but is on an older OS stream and hasn't yet planned to update shim (though arguably they should).

One option is we could try to extract the signing state of the shim binary? Another is we could offer a configuration knob (embedded in the image) that says whether this feature should be on or not.

Something like `bootupctl backend install --require-ms-2023-cert` that drops a bit of state in `EFI.json` which is honored here.

Then we change how we generate the base image to enable that flag by default, but anyone who wanted to opt out could do so with a custom image build.

And perhaps simpler it'd be nice to have e.g. `bootupctl backend --require-ms-2023-cert=false` which would mutate the JSON to drop that requirement

Rolv-Apneseth: The flag to disable it is a good idea. I believe your last point is to have it be enabled by default and we just ask users who do have the more niche use cases to manually disable it after they've updated `bootupd`, right? That would be my preferred approach.

cgwalters: We've gotten burned in the past by enabling things like this by default. In this case, it's probably fine.

Rolv-Apneseth: I understand, it's just hat the intent here is specifically to help prevent people from breaking their own setups unknowingly. Seems like disabling updates is the safer default in this scenario.

I'll add the flag and necessary state updates. If I add a test, the only thing I can think of doing is checking the flag updates state as expected. Any better ideas?

cgwalters: Making it a CLI option is also viable and perhaps the simplest...anyone who hits this and wants to bypass it could just pass that.

Kind of in the general case we need to have config options embedded in the image and overridable via dynamic invocations.

Rolv-Apneseth: So you're happy with no state being written to disk for this, just the flag to disable it for that run?

cgwalters: Yeah that seems fine to me, should be ~trivial to implement right?

Rolv-Apneseth: Yeah, should be

Rolv-Apneseth: Let me know if that's looking about how you imagined

cgwalters (approved):

PR #1149, opened 2026-09-01
statefile: use stable inode for write lock

Rolv-Apneseth:
The previous implementation used atomic_write_with_perms to create the lockfile, which replaces the file via rename on every call. This meant that concurrent callers could flock different inodes, defeating the purpose of the lock.

Let's replace that with OpenOptions create-or-open, which only creates the file if it doesn't already exist and opens the existing one otherwise, ensuring all callers lock the same inode. The lack of atomic writing here should not be an issue as the file has no content anyway.

Originally pointed out by the bot on https://github.com/coreos/bootupd/pull/1138, see https://github.com/coreos/bootupd/pull/1138#issuecomment-5479650954

### DIFF
--- a/src/backend/statefile.rs
+++ b/src/backend/statefile.rs
@@ -9,7 +9,7 @@ use anyhow::{bail, Context, Result};
 use bootc_internal_blockdev::Device;
 use camino::Utf8PathBuf;
 use cap_std::ambient_authority;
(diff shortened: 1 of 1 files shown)


Johan-Liebert1 (approved):

cgwalters:
Ouch.

Issue #1139, opened 2026-08-22
Fedora ostree installer image installs broken on UEFI with 0.2.36 due to #1126


AdamWill:
When 0.2.36 landed in Fedora Rawhide recently, ostree install tests on UEFI started failing (BIOS installs are still OK). The install apparently succeeds, but the installed system [boots to the grub emergency shell](https://openqa.fedoraproject.org/tests/5173872#step/_graphical_wait_login/1).

In the install logs, we see this:
```
Could not find /boot/efi/EFI when installing "EFI/fedora/grub.cfg"
```
which looks pretty significant. Indeed if I run an install and do not reboot at the end, but go to a console and check `/mnt/sysroot/boot/efi/EFI/fedora` (`/mnt/sysroot` is where the Fedora installer mounts the installed system root) it contains no `grub.cfg`, which is very likely the issue. (It also seems odd that the failure to install this file is just logged as an info, but the bootupctl command still exits 0?)

On a hunch, I tested a [scratch build](https://koji.fedoraproject.org/koji/taskinfo?taskID=149341885) with #1126 reverted, and it [passed tests](https://openqa.stg.fedoraproject.org/tests/overview?version=46&distri=fedora&groupid=2&build=Kojitask-149341885-NOREPORT). So it looks like that PR is the cause, @Johan-Liebert1 .

I'll send an official Fedora build with the PR reverted for now, but it'd be great to get a proper fix for this. CC @cgwalters . To reproduce, just try installing from e.g. https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20260821.n.0/compose/Silverblue/x86_64/iso/Fedora-Silverblue-Installer-Rawhide-20260821.n.0.x86_64.iso on a UEFI system or VM, and booting the installed system.

Johan-Liebert1:
Hi, could you guide me on how the installation is done here? I tried with `bootc install to-disk` and though I do not get the error

```
Could not find /boot/efi/EFI when installing "EFI/fedora/grub.cfg"
```

I do get grub getting stuck in the emergency shell. In my case that was due to `bootuuid.cfg` not being included in the ESP due to the following

```toml
# /usr/lib/bootc/install/10-boot.toml

# We don't want bootupd to stamp the boot and UEFI partitions
# with the boot uuid because these will change at first boot.
# See https://github.com/coreos/fedora-coreos-config/pull/3898
[install.bootupd]
skip-boot-uuid = true
```

[...]

PR #1151, opened 2026-09-04
BIOS Fix

Johan-Liebert1:
state: Skip state check for GrubCC and Systemd on non-efi systems

We were relying on `cfg(efi_arch)` to check for the existence of
bootupd's statefile, but that's not correct.

Using just `cfg(efi_arch)` is insufficient, as if the binary's
built on x86, but the target system is only BIOS, then we WILL enter
the branch which will want to check the statefile in the ESP and will
encounter an error.

Also, skip statefile check if we're only installing for BIOS and we do
not have EFI component

---

tests: Add bootc install tests for BIOS and UEFI

Also check whether the final VM boots or not

Related to: #1139

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,35 @@ jobs:
     runs-on: [ "${{ matrix.runner }}" ]
 
     steps:
+      - name: Install QEMU and OVMF for VM boot tests
+        if: ${{ matrix.runner == 'ubuntu-24.04' }}
+        shell: bash
+        run: |
+          set -xeuo pipefail
+          sudo apt install -y qemu-utils qemu-kvm ovmf
+
+      - name: Enable unprivileged /dev/kvm access
+        if: ${{ matrix.runner == 'ubuntu-24.04' }}
+        shell: bash
+        run: |
+          set -xeuo pipefail
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+          ls -l /dev/kvm
+
+      - name: Configure Podman
+        shell: bash
+        run: |
+          podman --version
+          mkdir -p ~/.config/containers
+          echo '{"defaultAction":"SCMP_ACT_ALLOW"}' > ~/.config/containers/seccomp.json
+          printf '[containers]\napparmor_profile = "unconfined"\nseccomp_profile = "%s/.config/containers/seccomp.json"\n' "$HOME" > ~/.config/containers/containers.conf
+          USER_NAME=$(id -un)
+          grep -q "^${USER_NAME}:" /etc/subuid || echo "${USER_NAME}:100000:65536" | sudo tee -a /etc/subuid
+          grep -q "^${USER_NAME}:" /etc/subgid || echo "${USER_NAME}:100000:65536" | sudo tee -a /etc/subgid
+          podman info --format '{{.Store.GraphRoot}}'
+
       - name: Setup env
         run: |
           IMG_NAME=localhost/bootupd-grubcc-${{ matrix.grubcc }}
@@ -153,3 +182,16 @@ jobs:
           set -xeuo pipefail
           sudo ./scripts/test-bootloader.sh "$IMG_NAME" "grub-cc"
           sudo ./scripts/test-bootloader.sh "$IMG_NAME" "systemd"
+
+      - name: Test BIOS installs and VM boot
+        if: ${{ matrix.runner == 'ubuntu-24.04' }}
+        run: |
+          set -xeuo pipefail
+          sudo ./scripts/test-bios-vm-boot.sh "$IMG_NAME"
+
+      - name: Test UEFI installs and VM boot
+        if: ${{ matrix.runner == 'ubuntu-24.04' }}
+        run: |
+          set -xeuo pipefail
+          sudo ./scripts/test-uefi-vm-boot.sh "$IMG_NAME" "ostree"
+          sudo ./scripts/test-uefi-vm-boot.sh "$IMG_NAME" "composefs"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,18 +48,6 @@ jobs:
           sudo udevadm trigger --name-match=kvm
           ls -l /dev/kvm
 
-      - name: Configure Podman
-        shell: bash
-        run: |
-          podman --version
-          mkdir -p ~/.config/containers
-          echo '{"defaultAction":"SCMP_ACT_ALLOW"}' > ~/.config/containers/seccomp.json
-          printf '[containers]\napparmor_profile = "unconfined"\nseccomp_profile = "%s/.config/containers/seccomp.json"\n' "$HOME" > ~/.config/containers/containers.conf
-          USER_NAME=$(id -un)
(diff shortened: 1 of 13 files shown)


Johan-Liebert1:
Finally the BIOS tests pass. Some weird issue with bcvk though. Looking into it

src/backend/statefile.rs
```diff
@@ -136,8 +136,20 @@ impl SavedState {
...
+
+                    // Using just `cfg(efi_arch)` is insufficient here, as if the binary's
+                    // built on x86, but the target system is only BIOS, then we WILL enter
+                    // this branch which is less than ideal
+
+                    if !is_efi_booted()? {
+                        log::debug!(
+                            "Testing saved state for {b}, but system is not EFI booted. Skipping"
+                        );
+                        return Ok(None);
```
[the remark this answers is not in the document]
Johan-Liebert1: > If a BIOS-booted host installs to a UEFI target with existing GrubCC or Systemd state, this returns Ok(None). ensure_not_present() then permits the re-install, and the later state update can overwrite the target state.

this'll never happen as a BIOS booted host won't have `/sys/firmware/efi`

.github/workflows/ci.yml
```diff
@@ -31,6 +31,35 @@ jobs:
...
+          set -xeuo pipefail
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+          ls -l /dev/kvm
+
+      - name: Configure Podman
+        shell: bash
+        run: |
+          podman --version
```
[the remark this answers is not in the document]
Johan-Liebert1: Yup, fixed

scripts/test-bios-vm-boot.sh
```diff
@@ -0,0 +1,75 @@
...
+IMAGE=$1
+
+. ./helpers.sh
+
+./test-bios-bootc-install.sh "$IMAGE"
+
+set +e
+
+umount -R /var/mnt 2>/dev/null || true
+losetup -j "$DISK_IMAGE" | cut -d: -f1 | xargs -r losetup -d
```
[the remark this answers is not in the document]
Johan-Liebert1: Fixed

Johan-Liebert1:
Need to run the fedora QA tests on that build. I'll report here once I get the status

.github/workflows/ci.yml
```diff
@@ -161,3 +170,21 @@ jobs:
...
           sudo ./scripts/test-bootloader.sh "$IMG_NAME" "grub-cc"
           sudo ./scripts/test-bootloader.sh "$IMG_NAME" "systemd"
+
+      - name: Test BIOS installs and VM boot
+        if: ${{ matrix.runner == 'ubuntu-24.04' }}
+        run: |
+          set -xeuo pipefail
+          sudo ./scripts/test-bios-vm-boot.sh "$IMG_NAME"
+
+      - name: Copy image to rootless podman storage
```
Rolv-Apneseth: Is this required if the tests run as root?

scripts/helpers.sh
```diff
@@ -0,0 +1,63 @@
...
+
+export DISK_IMAGE=/var/test-img.img
+export BOOT_TIMEOUT=600
+
+create_mount_device_bios() {
+    set -e
+
+    mkdir -p /var/mnt
+
+    if mount | grep -qeE /var/mnt; then
```
Rolv-Apneseth: I believe this is an issue, since `-e` will take whatever comes after as the pattern.

```
$ mount | grep -qeE /boot
grep: /boot: Is a directory
# It can work if we change the order
mount | grep -qEe /boot
# But this is probably fine here if we don't need regex?
mount | grep -qF /boot
```

scripts/helpers.sh
```diff
@@ -0,0 +1,63 @@
...
+
+    IMG_NAME=$1
+
+    DEVICE=$(losetup -j "$DISK_IMAGE" | cut -d: -f1)
+
+    # Skip IMG_NAME
+    local bootloader=("${@:2}")
+
+    podman run --rm --net=host --privileged --pid=host \
+      --privileged \
```
Rolv-Apneseth, line 56: Nit but this is a duplicate `--privileged`

scripts/test-bios-vm-boot.sh
```diff
@@ -0,0 +1,77 @@
...
+
+IMAGE=$1
+
+. ./helpers.sh
+
+./test-bios-bootc-install.sh "$IMAGE"
+
+set +e
+
+if mount | grep -qeE /var/mnt; then
```
Rolv-Apneseth: Same as above, maybe just `grep -qF /var/mnt`

scripts/test-uefi-vm-boot.sh
```diff
@@ -0,0 +1,91 @@
...
+. ./helpers.sh
+
+IMAGE=$1
+BACKEND=$2
+
+./test-uefi-bootc-install.sh "$IMAGE" "$BACKEND"
+
+set +e
+
+if mount | grep -qeE /var/mnt; then
```
Rolv-Apneseth: Here too

.github/workflows/ci.yml
Johan-Liebert1: yes, we need root perms to set up loop devices

scripts/helpers.sh
Johan-Liebert1: You're right. I'll update it

scripts/helpers.sh
Johan-Liebert1: ah, copy pasta-ing has some issues

.github/workflows/ci.yml
Rolv-Apneseth: But why save it into rootless podman storage then? Maybe I'm not understanding fully or just missing where that's being used

Johan-Liebert1: Yeah, that was a byproduct of me trying to use bcvk for uefi but something wasn't working out. Removed it

Johan-Liebert1: Wait, I had removed it, but it's still here somehow

scripts/test-uefi-vm-boot.sh
```diff
@@ -0,0 +1,91 @@
...
+boot_ok=0
+elapsed=0
+
+while [ "$elapsed" -lt "$BOOT_TIMEOUT" ]; do
+    if ! kill -0 "$QEMU_PID" 2>/dev/null; then
+        echo "QEMU exited prematurely after ${elapsed}s."
+        break
+    fi
+
+    if grep -qiE 'login:|welcome to|reached target.*multi-user' "$SERIAL_LOG" 2>/dev/null; then
```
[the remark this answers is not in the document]
Johan-Liebert1: We only want to check if VM boots or not. If the initrd fails we don't really care

scripts/test-bios-vm-boot.sh
```diff
@@ -0,0 +1,78 @@
...
+
+. ./helpers.sh
+
+./test-bios-bootc-install.sh "$IMAGE"
+
+if mount | grep -qF /var/mnt; then
+    umount -R /var/mnt
+fi
+
+set +e
```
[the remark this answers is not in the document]
Rolv-Apneseth: Doesn't the unmount come before `set +e` here?

Rolv-Apneseth (approved):
LGTM. I'll leave it up to you if you think the bot's suggestion is worth changing

scripts/test-bios-vm-boot.sh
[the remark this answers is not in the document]
Johan-Liebert1: yes, we want it that way as the very first time it runs, we don't have a loop device set up and we don't really want to fail on that

PR #1153, opened 2026-09-04
Add TMT gating tests and Packit CI

c4rt0:
Superseeds the TMT approach in [bootupd](https://github.com/coreos/bootupd/pull/967).

This follows the approach used for [console-login-helper-messages](https://github.com/coreos/console-login-helper-messages/pull/136), while using bootupd’s Rust packaging tooling.

* Add FMF/TMT smoke-test metadata.
* Add a minimal bootupd --version smoke test. 
* Support two TMT modes:
  * GitHub Actions builds the PR source and runs the source-built binary.
  * Packit/Testing Farm installs the RPM built from the PR source.
* Add Packit copr_build and tests jobs for Fedora Rawhide and CentOS Stream 10.
* Add cargo xtask spec to generate a spec referencing the PR-versioned .crate source and filtered vendor archive.
* Do not add gating.yaml; OSCI auto-detects TMT tests.

The original bootupd #967 workflow installed a stock distribution RPM, so it did not test PR content.  This change addresses that by having Packit build the PR RPM in Copr before Testing Farm runs the TMT smoke plan.bootupctl status and real bootloader-update behavior need an EFI-capable bootc/CoreOS-style environment. This is initial smoke testing - image-mode functional coverage can follow separately.


c4rt0:
Initial hosted validation seems to be working. Packit ran on this draft PR and built the RPMs in Copr.
Fedora Rawhide  test however failed as DNF reported bootupd RPM as `not a rpm` ?
I will investigate the above after the weekend...

.github/workflows/tmt-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Set additional paths
+        run: echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+      - name: Install dependencies
+        run: |
+          set -xeuo pipefail
+          sudo apt-get update
+          sudo apt-get install -y podman
+          pip install --user tmt
```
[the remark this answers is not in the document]
c4rt0: implemented

.github/workflows/tmt-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Set additional paths
+        run: echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+      - name: Install dependencies
+        run: |
+          set -xeuo pipefail
+          sudo apt-get update
+          sudo apt-get install -y podman
+          pip install --user tmt[provision-container]
```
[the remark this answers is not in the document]
c4rt0: ouch... result of manual changes. Fixed in the next push.

.github/workflows/tmt-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+
+jobs:
+  tmt-tests:
+    runs-on: ubuntu-latest
+    env:
+      PLAN_FILTER: ${{ github.event.inputs.plan_filter }}
+      USE_BUILT_FROM_SRC: ${{ github.event.inputs.use_built_from_src || 'true' }}
+
+    steps:
+      - uses: actions/checkout@v7
```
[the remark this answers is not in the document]
c4rt0: For a test-only workflow that just runs TMT and doesn't need git auth after checkout, disabling it costs nothing and removes the risk. I will add the proposed change in the next fix.

.github/workflows/tmt-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+    inputs:
+      plan_filter:
+        description: Test plan filter, for example tag:smoke.
+        required: false
+        default: ''
+      use_built_from_src:
+        description: Build bootupd from source instead of installing a distribution package.
+        required: false
+        default: 'true'
+
```
Rolv-Apneseth, line 18: Could we add a `concurrency` section like we do in: https://github.com/coreos/bootupd/blob/02e598afe1cbf68480fde1012e84c8a31712e65c/.github/workflows/rust.yml#L13-L16

xtask/src/main.rs
```diff
@@ -27,11 +27,13 @@ fn main() {
 fn try_main() -> Result<()> {
     let task = std::env::args().nth(1);
     let sh = xshell::Shell::new()?;
+    std::env::set_current_dir(sh.current_dir()).context("Changing to project root")?;
```
Rolv-Apneseth, line 30: How come this is required?

Rolv-Apneseth (commented):
2 very minor comments. Also, is any of this functionality already included in `coreos/repo-templates`, and if not should it be?

.github/workflows/tmt-tests.yml
c4rt0: Good catch, adding it. I didn't know about this before. it cancels in-progress runs on new pushes which makes sense here.

xtask/src/main.rs
c4rt0: AFAIK cargo xtask can be run from any directory in the workspace, and xshell figures out the project root internally, but file operations use the OS process working dir, which (again AFAIK) stays wherever cargo xtask was called from. I added this to make sure both agree on the same root, but happy to remove it if it's not really needed.

Rolv-Apneseth: I guess I'm not familiar with the specifics, just seemed like there was a specific needed if it was added for this path. Feel free to keep or remove, not a blocker

c4rt0:
> Also, is any of this functionality already included in `coreos/repo-templates`, and if not should it be?

Thanks for the review. To answer your question:  no, as it isn't the scope for this PR. The task is just to add gating tests for bootupd - I've done it previously for [CLHM](https://github.com/coreos/console-login-helper-messages/commit/ac209a88fda652162bff71984783212caf06c6ce).

Not sure if we want to add this globally...

Rolv-Apneseth (approved):
Fair enough. Feel free to merge when CI passes

